PR #14, opened 2021-11-16
Use brewfile

stephengroat:
unpinned the protobuf version, need to make sure that's OK


appflowy:
Wow, big thanks!! I didn't know Brewfile - just learned something new!

Issue #62, opened 2021-11-21
[Bug] Unable to build backend 


vnepogodin:
**Describe the bug**
The backend Cargo.toml doesn't build.

**To Reproduce**
1. ``git clone https://github.com/AppFlowy-IO/appflowy.git``
2. ``cd backend``
3. ``cargo build --release``

**Expected behavior**
The backend should be built.

**Additional context**
 - Environment [e.g. flutter doctor -v or rustup show]
 ``` bash
❯ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/vl/.rustup

installed toolchains
--------------------

nightly-2021-04-24-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)

active toolchain
----------------

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.58.0-nightly (2885c4748 2021-11-20)
```

**Comment by hobdob** [comment]
Can you post the error output?

**Comment by vnepogodin** [comment]
Yeah sure

**Comment by vnepogodin** [comment]
```
warning: field is never read: `secure_delete`
   --> /home/vl/appflowy/frontend/rust-lib/lib-sqlite/src/pool.rs:104:5
    |
104 |     pub(crate) secure_delete: bool,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

   Compiling lib-infra v0.1.0 (/home/vl/appflowy/frontend/rust-lib/lib-infra)
warning: `lib-sqlite` (lib) generated 1 warning
   Compiling actix-web v4.0.0-beta.11
   Compiling sqlx-macros v0.5.9
   Compiling actix-web-actors v4.0.0-beta.7
   Compiling actix-identity v0.4.0-beta.3
   Compiling actix-cors v0.6.0-beta.3
   Compiling hyper-tls v0.5.0
   Compiling reqwest v0.11.6
   Compiling backend-service v0.1.0 (/home/vl/appflowy/shared-lib/backend-service)
   Compiling lib-ws v0.1.0 (/home/vl/appflowy/shared-lib/lib-ws)
   Compiling sqlx v0.5.9
   Compiling ormx v0.7.0
   Compiling backend v0.1.0 (/home/vl/appflowy/backend)
error: error communicating with the server: Connection refused (os error 111)
   --> src/service/user/auth.rs:209:13
    |
[...]

Issue #13, opened 2021-11-16
[Question] Any plans for a Docker image?


sfkpmr:
I see .dockerignore, but no Dockerfile. Would love to see one!

stemount:
@sfkpmr what do you propose the use of Docker for? (building? hosting flutter as a pwa or something - not a flutter expert)

nicolasdanelon:
@sfkpmr maybe I don't want to install Rust on my machine, maybe I just want to download the docker images and run the app on my local environment without installing any dependency... 
also docker containers can be used to deploy the app faster on tons of boxes.
if some developer is using M1 and doesn't have a dependency or is just running Linux docker will have all dependencies already built in, configured and ready to run.
Docker is a must

appflowy:
@sfkpmr  The docker file is used to deploy the AppFlowy backend on the local host.

franciscoafonsoo:
Would you guys be open to a Pull Request for a docker-compose and some Dockerfiles ? 

I was thinking that it would help local development in the wild to have a docker-compose that launches the db, the backend and the frontend (never dockerized a flutter app but should be possible).

PabloCastellano:
Yes, please! The Rust+Flutter stack isn't that common nowadays and having to spend more than 1 hour (my case so far) for installing all the Rust and Flutter dependencies, plus compiling... is discouraging, unless you have been waiting for a Notion open source clone for years!

With Docker it would be super easy and fast to just fetch AppFlowy and run it

iamaravindim:
Please DO create the Docker Image we are interested in hosting in on our servers.

huangxinping:
Please DO create the Docker Image we are also VERY interested.

It has taken more than 3 hours to install all dependencies so far without success.

annieappflowy:
> Would you guys be open to a Pull Request for a docker-compose and some Dockerfiles ?
> 
> I was thinking that it would help local development in the wild to have a docker-compose that launches the db, the backend and the frontend (never dockerized a flutter app but should be possible).
> 
> Edit: i'm checking the database scripts. The idea would be to simplify this logic so that random developers can launch the project in a very simple and straight forward way

We're definitely open to PRs. To understand this better, do you want a docker backend or both the backend and the frontend?

annieappflowy:
[...]

PR #66, opened 2021-11-21
fix backend build

vnepogodin:
solving ticket #13, https://github.com/AppFlowy-IO/appflowy/issues/62

docker backend should be built with that arguments:
```bash
docker build -f backend/Dockerfile
```

### DIFF
--- a/backend/src/entities/token.rs
+++ b/backend/src/entities/token.rs
@@ -80,7 +80,7 @@ use backend_service::config::HEADER_TOKEN;
 use futures::future::{ready, Ready};
 
 impl FromRequest for Token {
-    type Config = ();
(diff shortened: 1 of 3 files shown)


backend/src/entities/token.rs
```diff
@@ -80,7 +80,7 @@ use backend_service::config::HEADER_TOKEN;
 use futures::future::{ready, Ready};
 
 impl FromRequest for Token {
-    type Config = ();
+    //type Config = ();
```
appflowy, line 83: The latest beta version of actix-web has removed this. I'll fix this by updating cargo.toml with actix-web = "4.0.0-beta.11"

Issue #50, opened 2021-11-20
[Bug] Colored background of highlighted text covers cursor


myl7:
**Describe the bug**
When text is highlighted with colored background, the background will cover the
'I' cursor put on it

**To Reproduce**
Run the desktop app and nagivate to "Read me" page, choose any highlighted text and put cursor on it.

**Expected behavior**
The cursor should be over the colored background, directly seen, like most rich-text editors.

**Screenshots**
![Cursor put on the highlighted text](https://user-images.githubusercontent.com/49339031/142719654-3942f6ae-f16f-4367-8305-69b8cdd579e0.png)
The blue point in the middle is the "I" cursor, covered.

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser: Chrome
 - Version: 96.0.4664.45

**Additional context**
`flutter doctor -v`:

<details>
<summary>Click to expand</summary>

```
[✓] Flutter (Channel dev, 2.6.0-11.0.pre, on Arch Linux 5.15.2-arch1-1, locale
    en_US.UTF-8)
    • Flutter version 2.6.0-11.0.pre at /opt/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 4b330ddbed (9 weeks ago), 2021-09-16 17:29:58 -0700
    • Engine revision 5b81c6d615
    • Dart version 2.15.0 (build 2.15.0-116.0.dev)
    • Pub download mirror https://pub.flutter-io.cn
    • Flutter download mirror https://storage.flutter-io.cn

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /home/myl/.local/lib/Android/Sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at:
      /home/myl/.local/share/JetBrains/apps/AndroidStudio/ch-0/203.7784292/jre/b
      in/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • CHROME_EXECUTABLE = /usr/bin/google-chrome-stable

[✓] Linux toolchain - develop for Linux desktop
    • clang version 13.0.0
    • cmake version 3.21.4
    • ninja version 1.10.2
    • pkg-config version 1.8.0

[✓] Android Studio (version 2020.3)
    • Android Studio at
[...]

PR #81, opened 2021-11-23
Added fix for cursor appearing below highlights

Hari-07:
Fix for #50


appflowy:
Thank you for contributing, I also check the editor code, and find out it seems like setting this value based on the current platform would be better?

image

Hari-07:
I'm not entirely sure why it's being configured in 2 different places, also isn't it better to set it independent of platform, because without it the bug would show up on the platforms it's not enabled

PR #85, opened 2021-11-23
ci: flutter git clone depth 1

pratikbalar:
This will save flutter clone time & bandwidth.


appflowy:
Thank you for optimizing it.

Issue #68, opened 2021-11-21
[Bug] Can't build application on Mac OS


Maksimka101:
**Describe the bug**
I can't build application on mac os. I followed the instruction in readme but I face an error in `build_sdk.sh`

```
Start building rust sdk
Default host: x86_64-apple-darwin
rustup home:  /Users/maksimzemlyanikin/.rustup

installed toolchains
--------------------

stable-x86_64-apple-darwin (default)
nightly-2021-04-24-x86_64-apple-darwin
nightly-x86_64-apple-darwin

active toolchain
----------------

nightly-x86_64-apple-darwin (directory override for '/Users/maksimzemlyanikin/Developer/appflowy/frontend')
rustc 1.58.0-nightly (2885c4748 2021-11-20)

[cargo-make] INFO - cargo make 0.35.6
[cargo-make] INFO - Project: dart-ffi
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: flowy-sdk-dev
[cargo-make] INFO - Profile: development-mac
[cargo-make] INFO - Running Task: legacy-migration
[cargo-make] INFO - Running Task: echo_env
CRATE_TYPE: staticlib 
BUILD_FLAG: debug 
TARGET_OS: macos 
RUST_COMPILE_TARGET: x86_64-apple-darwin 
FEATURES: flutter 
PRODUCT_EXT: app 
 
[cargo-make] INFO - Running Task: setup-crate-type
[cargo-make] INFO - Running Task: sdk-build
cargo build --package=dart-ffi --target x86_64-apple-darwin --features=flutter
   Compiling flowy-document v0.1.0 (/Users/maksimzemlyanikin/Developer/appflowy/frontend/rust-lib/flowy-document)
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> flowy-document/src/lib.rs:1:1
  |
1 | #![feature(vecdeque_binary_search)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
error: could not compile `flowy-document` due to previous error
[cargo-make] ERROR - Error while executing command, exit code: 101
[cargo-make] WARN - Build Failed.
```

**To Reproduce**
1. Follow steps 2, 3 and 4 in instruction
2. Install rust using `brew install rust`
3. Optionally run `rustup override set nightly ; rustup install nightly`
[...]

PR #88, opened 2021-11-23
fix: build_sdk script to use nightly version of cargo

RanKKI:
https://github.com/AppFlowy-IO/appflowy/issues/68#issuecomment-976448988

ps:
I don't know why `cargo +nightly` doesn't work


appflowy:
It seems like is your local toolchain issue?

RanKKI:
Aha, yeah my bad

Issue #59, opened 2021-11-21
[FR] Hover on bottom toolbar icons doesn't display a tooltip


Pranav4399:
Hovering on the icons at the bottom doesn't display a tooltip which can be useful to understand what that particular icon does.

![image](https://user-images.githubusercontent.com/28348429/142756668-382cf5f8-8d5e-49c6-a38c-faae9d82df37.png)

Hari-07:
I can take this up if its not being worked on by anyone as of now

PR #90, opened 2021-11-23
Implemented tooltip on toolbar buttons

Hari-07:
Implements #59 

I implemented it in the way I felt makes the most sense, please do let me know if it would be better done in another way, and I'd be happy to make the changes needed


appflowy:
LGTM!

Issue #91, opened 2021-11-23
[Bug] Docker build failed


e-tornike:
**Describe the bug**
Docker backend build fails.

**To Reproduce**
1. `git clone https://github.com/AppFlowy-IO/appflowy.git`
2. `docker build -f backend/Dockerfile .`
3. See error

**Expected behavior**
Docker should successfully create an image.

**Desktop (please complete the following information):**
 - Docker version 20.10.5, build 55c4c88

**Additional context**
 - Version: commit 627aac9
```
error[E0599]: no variant or associated item named `Empty` found for enum `AnyBody<_>` in the current scope
  --> /app/shared-lib/backend-service/src/response/response_http.rs:21:32
   |
21 |             Err(_) => AnyBody::Empty,
   |                                ^^^^^
   |                                |
   |                                variant or associated item not found in `AnyBody<_>`
   |                                help: there is an associated function with a similar name: `empty`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `backend-service`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
The command '/bin/sh -c RUSTFLAGS="-C opt-level=2" cargo build --release' returned a non-zero code: 101
```

vnepogodin:
It is not particularly docker(like at all)
And you can confirm it by just running same command at `backend/`
``` bash
SQLX_OFFLINE=true APP_ENVIRONMENT="production" cargo build --release
```

which is giving the same error: 
```
error[E0599]: no variant or associated item named `Empty` found for enum `AnyBody<_>` in the current scope
  --> /home/vl/appflowy/shared-lib/backend-service/src/response/response_http.rs:21:32
   |
21 |             Err(_) => AnyBody::Empty,
   |                                ^^^^^
   |                                |
   |                                variant or associated item not found in `AnyBody<_>`
   |                                help: there is an associated function with a similar name: `empty`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `backend-service`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

appflowy:
I have added the cargo.lock file to git. Would you mind pulling the latest main branch and try again

e-tornike:
Build is successful! Thank you :)

PR #95, opened 2021-11-24
Fix backend compilation + adding fallback for docker container

vnepogodin:
Fixes https://github.com/AppFlowy-IO/appflowy/issues/91

### DIFF
--- a/backend/src/middleware/auth_middleware.rs
+++ b/backend/src/middleware/auth_middleware.rs
@@ -87,7 +87,7 @@ where
             let fut = self.service.call(req);
             return Box::pin(async move {
                 let res = fut.await?;
-                Ok(res.map_body(|_, body| AnyBody::from_message(body)))
+                Ok(res.map_body(|_, body| AnyBody::new_boxed(body)))
             });
         } else {
             Box::pin(async move { Ok(req.into_response(unauthorized_response())) })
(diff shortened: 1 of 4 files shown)


backend/src/middleware/auth_middleware.rs
```diff
@@ -87,7 +87,7 @@ where
             let fut = self.service.call(req);
             return Box::pin(async move {
                 let res = fut.await?;
-                Ok(res.map_body(|_, body| AnyBody::from_message(body)))
+                Ok(res.map_body(|_, body| AnyBody::new_boxed(body)))
```
appflowy, line 90: It's weird, I can't find the definition of new_boxed.

vnepogodin: https://docs.rs/actix-http/3.0.0-beta.13/src/actix_http/body/body.rs.html#51
https://docs.rs/actix-http/3.0.0-beta.13/src/actix_http/body/body.rs.html#236

appflowy: Got it. I think the version of actix-web should be updated in the cargo.toml. 

Current:
actix-web = "4.0.0-beta.11"
actix-http = "3.0.0-beta.12"

It seems like the beta version will be updated automatically if the Cargo.lock does not exist.

appflowy:
I think I should add the Cargo.lock to git, as described in https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

> If you’re building a non-end product, such as a rust library that other rust packages will depend on, put Cargo.lock in your .gitignore. If you’re building an end product, which are executable like command-line tool or an application, or a system library with crate-type of staticlib or cdylib, check Cargo.lock into git.

vnepogodin:
Well build is fixed. So I'll revert commit

Issue #118, opened 2021-12-04
[Bug] GitHub misspelled


factotvm:
**Describe the bug**

On the ApPFlOwY web site, you do not spell GitHub correctly. They have a capital "G" and a capital "H" and while we can argue about whether that makes sense or not, it's their name and they're very consistent about spelling it the way they'd like it to be written.

**To Reproduce**
1. Visit https://www.appflowy.io/
2. Search using your browser's mechanism for "git"
3. Notice all instances are spelled "Github" (consistent, at least)

**Expected behavior**
Spell the name of the service you're using correctly

**Screenshots**

<img width="342" alt="image" src="https://user-images.githubusercontent.com/38792/144694758-818fbdc8-659f-42f9-9c7f-0612384d2e3c.png">

<img width="128" alt="image" src="https://user-images.githubusercontent.com/38792/144694763-8be3319f-6991-4c58-8615-9a5a0bb7a0da.png">

<img width="195" alt="image" src="https://user-images.githubusercontent.com/38792/144694775-736c6231-b351-477c-82de-50751ff07cb5.png">

**Additional context**

This is a bit tongue-in-cheek, but I do believe attention to detail is important and I hope you feel the same.

annieappflowy:
Hi factotvm,

Thank you for your feedback. I see your point, and I will update the UX text on the website.

Annie

annieappflowy:
Updated it. Closing this issue. Thanks

PR #120, opened 2021-12-05
Correct the capitalization of GitHub

factotvm:
To accomplish this task, I ran the following command (on macOS):

    find . -type f -print0 | xargs -0 sed -i '' 's/Github/GitHub/g'

GitHub, for better or worse, decided that the way their company is
spelled is with a capital "G" and a capital "H" and it's their name so
they get to do that if they want.

When I went to check out AppFlowy's web site, I noticed that GitHub
wasn't spelled correctly, and for me, this made the project feel less
than polished and myself reluctant to try it out. However, there is only
one way to find out about the project: I [opened a ticket]. And BAM! It
was resolved in mere hours by @annieappflowy. Nice.

With fresh motivation to check out the project, I downloaded the app
and...found myself staring at the same issue on the first screen of the
app. Well, as I was a bit flippant with the issue I opened up for the
web site misspellings, I decided to fix this one myself.

[opened a ticket]: https://github.com/AppFlowy-IO/appflowy/issues/118


annieappflowy (approved):
Thank you. I share the opinion of aiming for excellence!

Issue #117, opened 2021-12-03
[FR] i18n support


0xN0x:
Hi there !

As i was reading the code, i saw that the app is not at all made for i18n support. I can understand it's not the priority but by not working at least with a unique english file you will lose some time in the future to migrate every strings used by the app.

The app is still new but already watched by a lot of people. Planning a multi-language support is mandatory if you want to be a good alternative to Notion.
And with the open-source community, the translations can be very quick.

Every users can benefits from this, but mainly people don't having english as their main language. It can also ultimately facilitate (when the project has grown) the management of strings and therefore finding typos.

Also, there's tools like Weblate that can help you and the community to manage the translations files.

It's going to be something quite a long time to catch up (and the more something is delayed, the longer it will take to catch up). But at least, now there is an issue for that and it can be discussed.

irfanbacker:
@annieappflowy I'll take this issue up if no ones working on it

annieappflowy:
> @annieappflowy I'll take this issue up if no ones working on it

Awesome. I’ll tag you on this issue posted on Discord #pr-wanted

Please feel free to reach us if you need help.

irfanbacker:
# i18n approach
- Using package [easy_localization](https://pub.dev/packages/easy_localization)
- Translation files will under `assets/translations/` directory
- Translations files are `json` files named in the format `<lang_code>-<country_code>.json` or just `<lang_code>.json`.
eg: `en.json`, `en-UK.json`
- A wrapper around `MaterialApp` will be added to provide localization
    ```dart
    runApp(
      EasyLocalization(
          supportedLocales: const [Locale('en'), Locale('es', 'BR')], // Newly added languages should be added here
          path: 'assets/translations', // Path to translation .json files
          fallbackLocale: const Locale('en'), // Default locale if system locale not in supportedLocales
          child: app), // App startpoint i.e, ApplicationWidget in this case
    );
    ```
- Example `en-US.json` file
    ```json
    {
        "appName": "Appflowy",
        "defaultUsername": "Me",
        "welcomeText": "Welcome to @:appName",
        "githubStarText": "Star on GitHub",
        "subscribeNewsletterText": "Subscribe to Newsletter",
        "letsGoButtonText": "Let's Go",
    }
    ```
    Other features such as named arguments, gender and plurals etc. are supported and can be implemented refering to the documentation of `easy_localization`.
- Locale usage for `appName` from example above:
[...]

PR #129, opened 2021-12-07
Add i18n support

irfanbacker:
PR for #117 

I have added `en.json` translation file along with this, structured and named the keys as i see fit. Do let me know if changes are needed


appflowy:
LGTM! It saves my day.

PR #133, opened 2021-12-10
Updated Linux build document

MikeWallaceDev:
Marked Arch Linux as being verified (however protobuf isn't working yet)
Restructured the steps to better separate concerns
Added instructions for Arch Linux
Highlighted some steps that may be overlooked by someone who already has Flutter installed (or that I found confusing during installation
Moved environment settings to Step 5

### DIFF
--- a/doc/BUILD_ON_LINUX.md
+++ b/doc/BUILD_ON_LINUX.md
@@ -120,14 +120,6 @@ cd [frontend/]app_flowy/product/0.0.2/linux/Debug/AppFlowy/app_flowy
 ## [Optional] Step 5: Build Server side application (if you need to host web service locally)
 ------------------------------
 
-1. Environment settings
-
-Modify the following files to suite your environment.
-``` shell
-cp backend/configuration/base.example.yaml backend/configuration/base.yaml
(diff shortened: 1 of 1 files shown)


doc/BUILD_ON_LINUX.md
```diff
@@ -84,25 +98,36 @@ cargo make --profile development-linux-x86 flowy-sdk-dev
...
 # export DISPLAY=localhost:10
-# cd app_flowy/product/0.0.2/linux/Release/AppFlowy
-# ./app_flowy
-```
 
-## [Optional] Step 3: Build Server side application (optional if you don't need to host web service locally)
+## [Optional] Step 5: Build Server side application (if you need to host web service locally)
 ------------------------------
 
+1. Environment settings
```
appflowy, line 123: I think this step is unnecessary because these YAML files only work for the backend. The AppFlowy frontend and backend are not connected by now.

MikeWallaceDev: I totally agree, I just didn't want to remove anything from the doc :)

PR #135, opened 2021-12-12
Fix build sdk sh

MikeWallaceDev:
Fixed a typo (aarch64 -> x86)
Since I was here : Capitalized the OS names


irfanbacker:
Isn't step 5 already doing whats there in step 3? Feels redundant.

appflowy:
LGTM!

PR #139, opened 2021-12-13
Add i18n/zh-CN support

naughtz:
add language support(chinese)

### DIFF
--- a/frontend/app_flowy/assets/translations/zh_CN.json
+++ b/frontend/app_flowy/assets/translations/zh_CN.json
@@ -1,0 +1,122 @@
+{
+  "appName": "Appflowy",
+  "defaultUsername": "我",
+  "welcomeText": "欢迎使用@:appName",
+  "githubStarText": "Star on GitHub",
+  "subscribeNewsletterText": "消息订阅",
+  "letsGoButtonText": "开始",
+  "title": "标题",
+  "signUp": {
+    "buttonText": "注册",
+    "title": "注册@:appName账户",
+    "getStartedText": "开始",
+    "emptyPasswordError": "密码不能为空",
+    "repeatPasswordEmptyError": "确认密码不能为空",
+    "unmatchedPasswordError": "两次密码输入不一致",
+    "alreadyHaveAnAccount": "已有账户?",
+    "emailHint": "邮箱",
+    "passwordHint": "密码",
+    "repeatPasswordHint": "确认密码"
+  },
+  "signIn": {
+    "loginTitle": "登录@:appName",
+    "loginButtonText": "登录",
+    "buttonText": "登录",
+    "forgotPassword": "忘记密码?",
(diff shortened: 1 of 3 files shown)


frontend/app_flowy/assets/translations/zh_CN.json
```diff
@@ -0,0 +1,122 @@
...
+    "lastName": "姓",
+    "stepX": "第{X}步"
+  },
+  "oAuth": {
+    "err": {
+      "failedTitle": "无法连接到您的账户。",
+      "failedMsg": "请确认您已在浏览器中完成登录。"
+    },
+    "google": {
+      "title": "GOOGLE账号登录",
```
annieappflowy, line 115: For better readability, there should be a space between an English word and 中文字
eg： Google 账号登录

frontend/app_flowy/assets/translations/zh_CN.json
```diff
@@ -0,0 +1,122 @@
...
+      "instruction1": "为了导入您的Google联系人，您需要在浏览器中给予本程序授权。",
```
annieappflowy, line 116: same here, 为了导入您的 Google 联系人

annieappflowy (commented):
Thanks a lot for doing the translation. Please check out my comments. Other than those, it looks good to me.

frontend/app_flowy/assets/translations/zh_CN.json
C1oud555: This is some optimization I want to mention about Chinese. Can we add the space while rendering not inside inputs. It will be much more convenient for the mix input of Chinese and English. Also the full-width and Half-width Symbols.

I don't know if it's hard to implement since I know little about coding. But notion took a long time to add this feature(till now, it's still not completed". But it's Chinese alike "wolai" add this optimization at the beginning.

annieappflowy (approved):
LGTM. Thanks!

frontend/app_flowy/assets/translations/zh_CN.json
naughtz: ACtually, it is controversial whether spaces should be added between Chinese and English. But it seems that the two languages should be separated in some way. From the use experience, the gap caused by spaces is too large, probably because Chinese itself does not need space separation. Therefore, some software will insert their designed margin between Chinese and English, and the effect is good. Therefore, using spaces is an imperfect solution, but it works. If AppFlowy has a better plan in the future, it would be better than using spaces.

C1oud555: Display Margin! That's what I meant.

C1oud555: It seems a render enhancement about flutter.

https://github.com/flutter/flutter/issues/94531

So, there's nothing we can do except waiting.

PR #156, opened 2021-12-14
docs: Add Arch Linux dependence

yolo2h:
Arch Linux does not install xdg-user-dir by default. When I run app_flowy I got this error:

```shell
[ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: ProcessException: Failed to find "xdg-user-dir" in the search path.
  Command: xdg-user-dir 
#0      LocalProcessManager.runSync (package:process/src/interface/local_process_manager.dart:118)
#1      getUserDirectory (package:xdg_directories/xdg_directories.dart:154)
#2      PathProviderLinux.getApplicationDocumentsPath (package:path_provider_linux/path_provider_linux.dart:55)
#3      getApplicationDocumentsDirectory (package:path_provider/path_provider.dart:138)
#4      InitRustSDKTask.initialize (package:app_flowy/startup/tasks/init_sdk.dart:16)
#5      AppLauncher.launch (package:app_flowy/startup/launcher.dart:36)
#6      System.run (package:app_flowy/startup/startup.dart:54)
#7      main (package:app_flowy/main.dart:17)
<asynchronous suspension>
```

### DIFF
--- a/doc/BUILD_ON_LINUX.md
+++ b/doc/BUILD_ON_LINUX.md
@@ -110,6 +110,12 @@ cargo make -p production-linux-x86 appflowy-linux
 ## Step 4: Run the application
 ------------------------------
 
+Arch Linux need install `xdg-user-dirs`:
+
+```shell
+sudo pacman -S xdg-user-dirs
+```
+
 ```
 cd [frontend/]app_flowy/product/0.0.2/linux/Debug/AppFlowy/app_flowy
 ./app_flowy


MikeWallaceDev:
Why flavor of Arch are you using?  I'm on Manjaro and xdg-user-dirs was already installed.
If we determine that it is needed, please edit your change to put it in the correct place, above, with the other package installations.

yolo2h:
Oh I see, I build appflowy in docker. should we add run dependency? Then user can install run dependency when Linux binary released. Thank you for your reply!

MikeWallaceDev:
@appflowy I looked at the code and it is possible that the xdg_directories package (used by the path_provider package) throws an Exception if xdg-usr-dirs is not found. In fact, path_provider throws a different exception if a path isn't found. Neither of these exceptions are dealt with in AF. 

I'm not sure how to proceed here. 
* I don't think that we should add a dependency to xdg-usr-dirs in the BUILD_ON_LINUX.md document because if xdg-usr-dirs isn't there it's probably because there is something weird going on (maybe the person is trying to install on a server version of Linux!). If xdg-usr-dirs isn't there, the user's Desktop Environment doesn't use it, I don't think we should install it.
* I don't know what to do if we do trap that Exception, can we create a directory somewhere?? (It's a cross platform question, so difficult to answer)
* I do think that our initialization process should catch errors so that the app quits gracefully. But I don't know where you want to place that code.
* I do think that we should encapsulate path_provider so that we can easily switch it out if/WHEN something better comes along. But I don't know if you agree with that :)

MikeWallaceDev:
I'm not sure why this was cloase. @appflowy can you look into this?

MikeWallaceDev:
Ok, now I see why :D  It was a PR, I thought it was issue. Sorry, my bad.

I've opened an issue that will track this problem ( #394 )

PR #189, opened 2021-12-22
Add debug info button

### LINKED ISSUE #145: [FR] Create a menu item in the "?" that generates the required system info for a Bug report
**1~3 main use cases of the proposed feature**
As a developer it would be benificial to me to receive accurate information about the user's system
As a user it would help me to be able to just click a button a retrieve all the information that the dev requires.
As a user I sometimes don't know exactly what information the dev needs, or how to retrieve it.

**what types of users can benefit from using your proposed feature**
All AF devs and user's who need to create an issue report.

**Additional context**
I did not validate it, but there is a package for this : [Device Info Plus](https://pub.dev/packages/device_info_plus)

Poly-Pixel:
This pull request adds a menu item button to the question mark button in the lower right corner of the application, which will copy system info to the clipboard when pressed.

This is my first PR to AppFlowy, so please let me know if I forgot to do something or did something wrong.

fixes #145

### DIFF
--- a/frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
+++ b/frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
@@ -80,7 +80,6 @@ class QuestionBubble extends StatelessWidget {
                       ),
                     );
 
-
                     fToast.showToast(
                         child: toast,
                         gravity: ToastGravity.BOTTOM,
--- a/frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
+++ b/frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
@@ -5,6 +5,7 @@ import 'package:flowy_infra_ui/flowy_infra_ui.dart';
 import 'package:flowy_infra_ui/style_widget/button.dart';
 import 'package:flowy_infra_ui/style_widget/text.dart';
 import 'package:flowy_infra_ui/widget/spacing.dart';
+import 'package:flowy_log/flowy_log.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:provider/provider.dart';
@@ -47,20 +48,15 @@ class QuestionBubble extends StatelessWidget {
                   break;
                 case BubbleAction.debug:
                   final deviceInfoPlugin = DeviceInfoPlugin();
-
-                  final dynamic deviceInfo;
-                  if(Platform.isLinux) {
-                    deviceInfo = deviceInfoPlugin.linuxInfo;
-                  } else if(Platform.isMacOS) {
-                    deviceInfo = deviceInfoPlugin.macOsInfo;
-                  } else if(Platform.isWindows) {
-                    deviceInfo = deviceInfoPlugin.windowsInfo;
-                  } else {
-                    throw Exception('Unsupported platform when getting debug info');
-                  }
+                  final deviceInfo = deviceInfoPlugin.deviceInfo;
                   
                   deviceInfo.then((info) {
-                    Clipboard.setData(ClipboardData( text: info.toMap().toString() ));
+                    var debugText = "";
+                    info.toMap().forEach((key, value) {
+                      debugText = debugText + "$key: $value\n";
+                    });
+
+                    Clipboard.setData(ClipboardData( text: debugText ));
 
                     Widget toast = Container(
                       padding: const EdgeInsets.symmetric(horizontal: 24.0, vertical: 12.0),
@@ -85,8 +81,33 @@ class QuestionBubble extends StatelessWidget {
                         gravity: ToastGravity.BOTTOM,
                         toastDuration: const Duration(seconds: 3),
                     );
-                  });
+                  }).catchError((error) {
+                    Log.info("Debug info has not yet been implemented on this platform");
 
+                    Widget toast = Container(
+                      padding: const EdgeInsets.symmetric(horizontal: 24.0, vertical: 12.0),
+                      decoration: BoxDecoration(
+                        borderRadius: BorderRadius.circular(25.0),
+                        color: Colors.red,
+                      ),
+                      child: Row(
+                      mainAxisSize: MainAxisSize.min,
+                      children: const [
+                          Icon(Icons.close),
+                          SizedBox(
+                            width: 12.0,
+                          ),
+                          Text("Unable to copy debug info to clipboard"),
+                        ],
+                      ),
+                    );
+
+                    fToast.showToast(
+                        child: toast,
+                        gravity: ToastGravity.BOTTOM,
+                        toastDuration: const Duration(seconds: 3),
+                    );
+                  }, test: (e) => e is UnimplementedError);
                   break;
               }
             });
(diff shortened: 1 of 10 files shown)


MikeWallaceDev:
First impression : Very nice! Does exactly what we asked for!
Comments : 
* I noticed that the text copied to the clipboard is in json format. Interesting choice!
* I tried to view the json in two viewers, neither could view it. Which viewer did you use?
* If the codegen_loader.g.dart file is generated by the build process, we should probably put it in .gitignore
* same thing for locale_keys.g.dart
* I'm looking at the pubspec and you used device_info_plus 2.0.0, when the current version is 3.2.0, what is the reason for that? 
Second impression : Good job! :+1: The code looks very nice and clean.

frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
```diff
@@ -39,6 +44,50 @@ class QuestionBubble extends StatelessWidget {
                   break;
                 case BubbleAction.help:
                   _launchURL("https://discord.gg/9Q2xaN37tV");
+                  break;
+                case BubbleAction.debug:
+                  final deviceInfoPlugin = DeviceInfoPlugin();
+
+                  final dynamic deviceInfo;
+                  if(Platform.isLinux) {
```
MikeWallaceDev, line 52: I'm not a fan of this switch, this won't work once we add mobile devices...

From what I saw in the documentation for 3.2.0 you can just do :
```dart
final deviceInfoPlugin = DeviceInfoPlugin();
final deviceInfo = await deviceInfoPlugin.deviceInfo;
final map = deviceInfo.toMap();
// Push [map] to your service.
```
and this would return all the information that it has on the device, no matter the platform

frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
```diff
@@ -39,6 +44,50 @@ class QuestionBubble extends StatelessWidget {
...
+                          Icon(Icons.check),
+                          SizedBox(
+                          width: 12.0,
+                          ),
+                          Text("Copied debug info to clipboard!"),
+                        ],
+                      ),
+                    );
+
+
```
MikeWallaceDev, line 83: extra line to delete please

MikeWallaceDev (changes requested):
I think I have to mark this as "Request Changes".

Poly-Pixel:
- The info is not actually in JSON format. It's a map of the data that device_info_plus provides, cast to a string. Would you like it formatted differently?
- I used device_info_plus 2.0.0 as there appears to be a bug with 3.2.0. When I tried using `deviceInfoPlugin.deviceInfo` with 3.2.0, I got an unhandled exception `UnimplementedError: linuxInfo() has not been implemented`. I did look at the issue tracker for device_info_plus, and today I found that there is a pull request in progress to fix it (https://github.com/fluttercommunity/plus_plugins/pull/663). Should this PR wait until that fix is available and upgrade device_info_plus before merging?

MikeWallaceDev:
My initial thought was that people would just paste that information into their bug reports. I think that a human readable format would be better for that use case.
I think that the json/map format would be good if we decide to automatically send information to a server somewhere. But that's not for a while...
If you don't mind, a human readable format would be better, do you agree?

As for 3.0.2, that's a toughy. It really depends on them. If they're fast, then we're better off waiting.

Here's what I suggest : use 3.0.2, but catch that exception. That way, we'll be ready for all platforms. It won't work on Linux right away, but when they make their change it will work like magic. And, if we run on other platforms, we're safe from that exception.

Poly-Pixel:
I agree, human-readable is best - I'll do that now.

So, whenever they make the change, all we have to do is update the dependency, right?

Poly-Pixel:
The codegen_loader.g.dart and locale_keys.g.dart have to be generated by the commands in doc/TRANSLATION.md, and the app won't build without them.

MikeWallaceDev:
Yeah, we just have to update our dependency. It will also happen automatically (up to a certain point) with the ^ in the version number.

frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
```diff
@@ -5,13 +5,19 @@ import 'package:flowy_infra_ui/flowy_infra_ui.dart';
...
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:provider/provider.dart';
 import 'package:dartz/dartz.dart' as dartz;
 import 'package:styled_widget/styled_widget.dart';
 import 'package:package_info_plus/package_info_plus.dart';
 import 'package:url_launcher/url_launcher.dart';
 import 'package:app_flowy/generated/locale_keys.g.dart';
+import 'package:device_info_plus/device_info_plus.dart';
+import 'dart:io';
```
MikeWallaceDev, line 18: I get a warning that 'dart.io' isn't used.

frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
```diff
@@ -40,6 +46,69 @@ class QuestionBubble extends StatelessWidget {
...
+                        color: theme.main1,
+                      ),
+                      child: Row(
+                      mainAxisSize: MainAxisSize.min,
+                      children: const [
+                          Icon(Icons.check),
+                          SizedBox(
+                          width: 12.0,
+                          ),
+                          Text("Copied debug info to clipboard!"),
```
MikeWallaceDev, line 74: Rats, I should have seen this the first time. My bad.  The text should use the translation mechanism. Note that it is not up to you to add translations (you can, nobody is stopping you...) As a developer you just have to make sure that the code uses translations.  The translators will do the translating in time...

frontend/app_flowy/lib/workspace/presentation/widgets/float_bubble/question_bubble.dart
```diff
@@ -40,6 +46,69 @@ class QuestionBubble extends StatelessWidget {
...
+                        color: Colors.red,
+                      ),
+                      child: Row(
+                      mainAxisSize: MainAxisSize.min,
+                      children: const [
+                          Icon(Icons.close),
+                          SizedBox(
+                            width: 12.0,
+                          ),
+                          Text("Unable to copy debug info to clipboard"),
```
MikeWallaceDev, line 100: this text should use translations too :)

MikeWallaceDev (changes requested):
Thank you for this! Just some slight changes and we should be good!

MikeWallaceDev (approved):

MikeWallaceDev:
@appflowy I want you to check this because
* There are changes to the pubspeck lock file. 
* There are generated files added to the repo

What should the procedure be for those kind of things? I think generated files should not be checked-in, and we generate them in the build process. What do you think?
I think that changes to the lock file should be approved by you.

appflowy:
@MikeWallaceDev 

- I think updating the lock file is ok.
- Auto-generated files should be ignored by default. Build-process is the best place to do the auto-generated things. One script that used for all the auto-generated files.

MikeWallaceDev:
I did some more checking, yes it's recommended to keep the lock file in open projects.

I made a PR where I remove the generated files from the repository and added a step in the build file

PR #199, opened 2021-12-25
Remove generated language files

MikeWallaceDev:


appflowy:
It seems that `flutter pub get` must run before the generated-language script.
image

MikeWallaceDev:
Ah yes. Obviously I tested it on my machine, but get pub had already ran..

We need to speed up the CI tests so we can run them locally before creating the PR. I have no clue why they're so slow...

MikeWallaceDev:
I totally screwed the pooch.  Going to start over.

Issue #31, opened 2021-11-18
[FR] Darkmode


rhinotastic:
**1~3 main use cases of the proposed feature**
Dark mode UI.

**what types of users can benefit from using your proposed feature**
Pretty much everyone.

**Additional context**
Some form of darkmode or "Themes" backed in.

SleepDaemon:
I agree, dark mode is more eye comforting. My eyes are very sensitive to bright lights and just using the app burns into my retinas. Having dark mode would make it more comfortable to use and allowing longer work time.

MostHated:
Yeah, I will keep an eye on this thread, because even with my monitor's brightness down, it's just too bright for me to try and use for any extended period of time.

fellahi-ali:
@annieappflowy  @appflowy Are you currently using a custom theme for the material app or just the default?
I'm just getting started with Flutter but I may work on this issue if I get some guidance 💙

notd5a-alt:
If anyone isnt working on this is can probably add this feature.  Just need to know what the devs / people may want in terms of accent colors.
Also any recommendations for the positioning / placement of the dark mode switch would be appreciated.

rileyhawk1417:
Well I think it would be nice to have custom themes, but for now maybe focusing on Light and Dark theme switches would be a starting point.

eqixcx:
A Dark/Light toggle would be nice, but honestly, most people just stick to one 'theme', so even if it was a config file that people can replace by dropping in another and even just a 'Theme Picker' to select a config file, that would be cool :)

Would also be very good in adoption of AppFlowy for people who enjoy ricing all of their programs/utilities like the folks on r/unixporn

notd5a-alt:
Ok, im going to try to work on this feature for you guys.

rileyhawk1417:
@notd5a-alt any progress? 
For now am trying to make sure everything matches like font and other backgrounds.

notd5a-alt:
Hi, ive been extremely busy and i actually forgot i was working on this xD.
I dont have the time rn, maybe someone else can / is working on it. I was
also thinking it might be best to implement dark modes and theming after
the extension system comes online. Then it should become easier to make
themes and have custom user made themes. Something like Typoras system? But
more refined and user friendly.

rileyhawk1417:
Well I am currently working on it was also wondering if anyone else had given it a shot

branhillsdesign:
[...]

PR #259, opened 2022-01-17
Refactor default themes, and start migrating from `Provider` to `Bloc (Cubit)`

hieugao:
#### Changes

- Instead of using a custom `AppTheme` to theming, we could use `Theme.of(context)...` instead, leads to a more consistent app where the app, SDK and package widgets could benefit from
- Using BLOC (Cubit) instead of Provider to control theming
- Persist theme setting with Hydrated BLOC
- Update the colors and terms according to https://github.com/AppFlowy-IO/appflowy/issues/31#issuecomment-995269689

#### Thought

We need to work more on the default theme currently because the colors are not working pretty well (using the same tints for both theme modes, etc). ~IMO I don't think we should put a lot of time into this aspect but rather spending time on implementing features, it seems like Material Design is the way to go at the current point.~ Seems like it's more practical if we could have an actual design (on Figma or something) so we could test out the colors and establish the default theme for AppFlowy.

~But maybe it will not take a lot of time since~ now we theming with `Theme.of(context)...`, changes will be applied throughout the app pretty easily :sweat_smile:.


frontend/app_flowy/lib/user/presentation/sign_up_screen.dart
```diff
@@ -1,203 +1,203 @@
-import 'package:app_flowy/startup/startup.dart';
```
appflowy, line 1: sign_in_screen and sign_up_screen will be used in the next version, I think it's not necessary to comment them out because the flutter_lint can't check them.

frontend/app_flowy/lib/user/presentation/skip_log_in_screen.dart
```diff
@@ -56,7 +54,7 @@ class _SkipLogInScreenState extends State<SkipLogInScreen> {
           logoSize: const Size.square(60),
         ),
         const VSpace(80),
-        GoButton(onPressed: () => _autoRegister(context)),
+        _GoButton(onPressed: () => _autoRegister(context)),
```
appflowy, line 57: Nice! "Private" the button is better

frontend/app_flowy/lib/workspace/presentation/widgets/menu/widget/app/header/header.dart
```diff
@@ -54,7 +52,7 @@ class MenuAppHeader extends StatelessWidget {
           theme: ExpandableThemeData(
             expandIcon: FlowyIconData.drop_down_show,
             collapseIcon: FlowyIconData.drop_down_hide,
-            iconColor: theme.shader1,
```
appflowy, line 57: The colors in AppTheme are specified by the design specs. I think keeping those names is good for development.

appflowy:
> #### Changes
> * Instead of using a custom `AppTheme` to theming, we could use `Theme.of(context)...` instead, leads to a more consistent app where the app, SDK and package widgets could benefit from
> * Using BLOC (Cubit) instead of Provider to control theming
> * Persist theme setting with Hydrated BLOC
> * Update the colors and terms according to [[FR] Darkmode #31 (comment)](https://github.com/AppFlowy-IO/appflowy/issues/31#issuecomment-995269689)
> 
> #### Thought
> We need to work more on the default theme currently because the colors are not working pretty well (using the same tints for both theme modes, etc). ~IMO I don't think we should put a lot of time into this aspect but rather spending time on implementing features, it seems like Material Design is the way to go at the current point.~ Seems like it's more practical if we could have an actual design (on Figma or something) so we could test out the colors and establish the default theme for AppFlowy.
> 
> ~But maybe it will not take a lot of time since~ now we theming with `Theme.of(context)...`, changes will be applied throughout the app pretty easily 😅.

Thank you! @hieugao.  I agree with your thought. I
 have commented on the PR.

frontend/app_flowy/lib/user/domain/i_splash.dart
```diff
@@ -17,7 +17,7 @@ abstract class ISplashUserWatch {
 }
 
 abstract class ISplashRoute {
-  void pushSignInScreen(BuildContext context);
+  // void pushSignInScreen(BuildContext context);
```
appflowy, line 20: It will be used later in the next version. Just keep it

frontend/app_flowy/lib/user/presentation/widgets/background.dart
```diff
@@ -49,11 +46,7 @@ class FlowyLogoTitle extends StatelessWidget {
           const VSpace(30),
           Text(
             title,
-            style: TextStyle(
-              color: theme.textColor,
-              fontWeight: FontWeight.w600,
-              fontSize: 24,
-            ),
+            style: Theme.of(context).textTheme.headline5,
```
appflowy, line 49: Is "Theme.of(context).textTheme.headline5"

 is equal to 

"TextStyle(
              color: theme.textColor,
              fontWeight: FontWeight.w600,
              fontSize: 24,
            )
"?

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
```diff
@@ -1,164 +0,0 @@
...
-  /// Default constructor
-  AppTheme({this.isDark = false});
-
-  /// fromType factory constructor
-  factory AppTheme.fromType(ThemeType t) {
-    switch (t) {
-      case ThemeType.light:
-        return AppTheme(isDark: false)
-          ..surface = Colors.white
-          ..hover = const Color(0xFFe0f8ff) //
```
appflowy, line 64: The color names in AppTheme are specified by the design specs. I think keeping those names is good for development. Or Any suggestions about how to map the color to the design specs.

hieugao: The problem seems like we naming the same thing with different terms, maybe those terms are already covered inside `ThemeData` (or Material Design in general) for around 80% cases (or even higher). Currently we can't add custom color names to it (I have to use `extension`), so I think we should follow Material Design for most of the colors names and having some custom colors for the app.

frontend/app_flowy/lib/user/presentation/widgets/background.dart
hieugao: I think so, https://api.flutter.dev/flutter/material/TextTheme-class.html (font weight is probably `w400`). Seems like `headline6` is more suitable. I think we need a more consistent text styling.

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
hieugao: Besides, naming colors with number is very confusing IMO. With `shaders`, `bg` we could use [`elevation`](https://api.flutter.dev/flutter/material/Material/elevation.html) instead (not all of the time), with `tints` we should use the actual color names, with `main1` we have `primary` and `main2` we have `secondary`.

appflowy: > The problem seems like we naming the same thing with different terms, maybe those terms are already covered inside `ThemeData` (or Material Design in general) for around 80% cases (or even higher). Currently we can't add custom color names to it (I have to use `extension`), so I think we should follow Material Design for most of the colors names and having some custom colors for the app.

Agreed! It seems like we need to map the names of the color to the design specs. 
For example:
bg(from design specs)  -> elevation (from the dart code)

So that each developer can find the color when building the UI from design specs.

rileyhawk1417:
Hi @hieugao just to ask the file "frontend/app_flowy/lib/common/theme/theme.dart". Can't it use the values from hydrate bloc instead of hard coding the values directly in the value?

frontend/app_flowy/lib/user/presentation/widgets/background.dart
rileyhawk1417: So the Theme.of(context).textTheme can handle the font style color without having to hard code it?

hieugao:
@rileyhawk1417, I thought like you too, if what you mean by Hydrated Bloc automatically know the `isDark` property to store it. I don't know much (or anything) about Bloc though, it's just AppFlowy uses Bloc as the main SM so I just replaced Provider with Cubit for consistency.

frontend/app_flowy/lib/user/presentation/widgets/background.dart
hieugao: @rileyhawk1417, sorry I didn't saw your comment. Yep, I mean that is the purpose of theming :smile_cat:, consistency and reduce unnecessary hard coded styles. We could follow Material Design preset (the link) or we could customize it with `textTheme` property inside `ThemeData`.

rileyhawk1417:
@hieugao it's alright am still new to Flutter and learning. Was just asking as it would be nice to have one component. Instead of having small components managing different things.

hieugao:
Hi @appflowy, I think it's good enough to be merged, let's me know in case anything is missing.

appflowy:
> Hi @appflowy, I think it's good enough to be merged, let's me know in case anything is missing.

Thanks! I will check it ASAP.

hieugao:
Hi @appflowy, is this good enough to be merged? I'm having some free time right now, so let me know in case anything need to be changed, I'll fix it ASAP.

frontend/app_flowy/lib/common/theme/theme.dart
```diff
@@ -0,0 +1,151 @@
...
+
+  ThemeData get themeData => ThemeData(
+        brightness: isDark ? Brightness.dark : Brightness.light,
+
+        primaryColor: const Color(0xFF00bcf0),
+
+        // TODO: Custom color with tins and shades
+        primarySwatch: Colors.blue,
+
+        scaffoldBackgroundColor: isDark ? const Color(0xFF292929) : const Color(0xFFf7f8fc),
```
appflowy, line 44: How can I extend the others theme type? For example, we have an enum for ThemeType.
enum ThemeType {
  light,
  dark,
}

frontend/app_flowy/lib/user/presentation/sign_in_screen.dart
```diff
@@ -191,16 +186,15 @@ class EmailTextField extends StatelessWidget {
...
       builder: (context, state) {
         return RoundedInputField(
           hintText: LocaleKeys.signIn_emailHint.tr(),
           style: const TextStyle(fontSize: 14, fontWeight: FontWeight.w500),
-          normalBorderColor: theme.shader4,
-          highlightBorderColor: theme.red,
-          cursorColor: theme.main1,
+          normalBorderColor: Colors.transparent,
+          highlightBorderColor: Theme.of(context).primaryColor,
+          cursorColor: Theme.of(context).primaryColor,
```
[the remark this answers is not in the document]
appflowy: I am afraid some changes will be missed in PR 2 and the colors will mix up. So one PR would be fine

appflowy:
> Hi @appflowy, I changed some of the code according to your suggestions.
> 
> The PR kind of complicated since Theme is using through out the app, and also apart of it is my fault when mixing up:
> 
> * Refactoring the `AppTheme` to `Theme.of(context)`, `Provider` to `Bloc`, etc
> * And changing the colors which supposed to be in another PR, because it also require the design specs, and could be done individually _(I didn't want to go through the same code twice, so I f-up lol)_ since we could use the new `Theme.of(context)` to change the colors whenever and wherever we want

Yep, I know. it's complicated!

hieugao:
> @appflowy - I have reviewed this PR. But I can't make sure the colors is right or not. Besides, I think we need a map that mapping the color to the design specs. Otherwise, I don't know which color should be used base on the design specs.

We could create a new PR to do that, the main point of this PR is refactoring `AppTheme` to `Theme.of(context)`. As long as we roll `Theme.of(context)` out we could do update (refactor) the app colors whenever we want.

According to what you said, it seemed like AppFlowy has its own custom Design System. If so then yes, we need to map it to Material Design since we using Material Design in our app. But IMO, AppFlowy should build its own custom styles on top of Material Design, instead of creating its own Design System which will cost of lot of time and resource.

frontend/app_flowy/lib/common/theme/theme.dart
hieugao: Yes we could get rid of the `isDark` (or `brightness`) since there will be more themes than just `light` and `dark`, so I think this `enum ThemeType` should becomes `class ThemeType extends ThemeData`, so we could reload `ThemeData` programmatically when user change the theme and reusing the `properties`.

Hmmm, so because of that I don't think what you want to do have any problem with this PR so far. This PR is solely for refactoring `AppTheme` to `Theme.of(context)`.

But I understand what you wanted and I don't have enough knowledge of how Flutter could load JSON (https://github.com/peiffer-innovations/json_theme) at runtime (using Rust instead?). But if we could found a way to load JSON then reloading `ThemeData` is not really a big deal with `ThemeCubit`.

PR #260, opened 2022-01-18
Update README.md

tsuiyuenhong:
I can't switch to the `macOS` platform in visual studio code when building the appflowy. And I later found the reason was that the following command was not executed.

Hope to add this comment to the README to help the others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,17 @@ flutter channel stable
 ```
 
 **Step 4:**
+* You should enable the specified platform first if you don't enable it before and then select the desktop device.
+```shell
+# for windows
+flutter config --enable-windows-desktop
 
+# for macos
+flutter config --enable-macos-desktop
+
+# for linux
+flutter config --enable-linux-desktop
+```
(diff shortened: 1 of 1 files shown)


MikeWallaceDev:
Thank you for that. 
In reality those installation instructions should probably be removed from the README. We have separate installation instructions for multiple cases in the documentation. (I'm not sure if that documentation is public yet, so I won't link it here.)
And indeed, this line was missing for the Windows instructions. So I added it, thanks :)

appflowy:
@tsuiyuenhong Thank you! Hope you have fun in AF

appflowy:
> Thank you for that. In reality those installation instructions should probably be removed from the README. We have separate installation instructions for multiple cases in the documentation. (I'm not sure if that documentation is public yet, so I won't link it here.) And indeed, this line was missing for the Windows instructions. So I added it, thanks :)

We can remove the macOS install instructions in a separate file like Windows and Linux does.

MikeWallaceDev:
I have updated the readme and sent in a PR. The readme now points to the official documentation.

PR #262, opened 2022-01-20
Settings Modal

Hari-07:
Settings Modal implemented and existing Theme Switch moved into it


rileyhawk1417:
It looks fine on my end going forward we will need to fix the language selector.

PR #293, opened 2022-02-01
Changes language names to their respective languages

### LINKED ISSUE #290: [FR] Changes language names to their respective languages
**1~3 main use cases of the proposed feature**
Ex: As a non English speaker, I want to able to change the language of the app to my preferred language. However, the languages in the application are currently written in English.

**what types of users can benefit from using your proposed feature**
Non English speakers who want to change the language of the application

**Additional context**
When going into the settings to change the language setting, the user is presented with language names in English. This makes it difficult for a non English speaker to find their language.

**To Reproduce**
1. Click on the 'gear' icon to go into settings
2. Click on the 'Language' menu option. (This has an icon, so it's not that bad if you don't understand English)
3. A menu appears with all the languages in English.

**Screenshots**
image

bolshoytoster:
Fixes #290

Some non-english speakers might find it hard to change their language
because they are all written in english.

This commit changes the language names in the selector to their
respective languages.

- Added two functions in `frontend/app_flowy/packages/flowy_infra/lib/language.dart`:
- - fullStringFromLanguage
- - languageFromFullString

These convert two and from the language's name and it's AppLanguage.

- '_LanguageSelectorDropdownState' now uses strings for the `value`
field, and converts using the afformentioned functions.

### DIFF
--- a/frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_language_view.dart
+++ b/frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_language_view.dart
@@ -31,16 +31,16 @@ class _LanguageSelectorDropdownState extends State<LanguageSelectorDropdown> {
   @override
   Widget build(BuildContext context) {
     return DropdownButton<AppLanguage>(
-      value: fullStringFromLanguage(context.read<AppearanceSettingModel>().language),
+      value: context.read<AppearanceSettingModel>().language,
       onChanged: (val) {
         setState(() {
-          context.read<AppearanceSettingModel>().setLanguage(context, languageFromFullString(val!));
+          context.read<AppearanceSettingModel>().setLanguage(context, val!);
         });
       },
       autofocus: true,
       items: AppLanguage.values.map((language) {
         return DropdownMenuItem<AppLanguage>(
-          value: fullStringFromLanguage(language),
+          value: language,
           child: Text(describeEnum(language)),
         );
       }).toList(),
--- a/frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_language_view.dart
+++ b/frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_language_view.dart
@@ -31,16 +31,16 @@ class _LanguageSelectorDropdownState extends State<LanguageSelectorDropdown> {
   @override
   Widget build(BuildContext context) {
     return DropdownButton<AppLanguage>(
-      value: context.read<AppearanceSettingModel>().language,
+      value: fullStringFromLanguage(context.read<AppearanceSettingModel>().language),
       onChanged: (val) {
         setState(() {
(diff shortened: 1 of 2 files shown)


appflowy:
Hi, @bolshoytoster, Thank you for your improvements. But the CI seems not passed.
image

bolshoytoster:
@appflowy sorry, I wrote `String` with a lowercase.

MikeWallaceDev:
Hi there @bolshoytoster! Thanks for the PR, but you didn't need to add all that code. 
There's two ways to do this: either directly put the original language in en.json (and then not translate in the other languages), or just hard code the strings. Hard coding might be the better approach here. The languages will never be translated, they will always be shown in the original language. (i.e. "Français" will always be "Français" no matter what language the app is in)

MikeWallaceDev (changes requested):
This must be changed to a simpler solution. There's no need for added functions.

bolshoytoster:
@appflowy
> or just hard code the strings

I can’t name the enums with non-ascii characters because dart doesn’t support it.

I’ll use en.json instead.

bolshoytoster:
@MikeWallaceDev

looking at [the way the translations are implemented](https://github.com/AppFlowy-IO/appflowy/blob/728d60239a34a73174e99c191299be0ab29b1f97/frontend/app_flowy/lib/workspace/domain/edit_action/view_edit.dart#L12-L34), wouldn't it be the same/worse than what I've already implemented.

> There's no need for added functions.

I would still have to add functions if I used the `en.json` method.

appflowy:
> @MikeWallaceDev
> 
> looking at [the way the translations are implemented](https://github.com/AppFlowy-IO/appflowy/blob/728d60239a34a73174e99c191299be0ab29b1f97/frontend/app_flowy/lib/workspace/domain/edit_action/view_edit.dart#L12-L34), wouldn't it be the same/worse than what I've already implemented.
> 
> > There's no need for added functions.
> 
> I would still have to add functions if I used the `en.json` method.

I think using enum is Ok, it is used to choose the language.

MikeWallaceDev:
Here is what I wrote to Nathan earlier, it should help you out : 

> However, easy_localization has an attribute called "supportedLocales" which contains all the Locales that we use in the app.  I suggest that we iterate through that, instead of having an enum.  That way, when a new language is added, if the dev forgets to update the enum, we won't be screwed.
> 
> For example, we just added German. It's not in the enum yet. The pseudo code would be: 
> iterate through "supportedLocales" {
>    if "en" then return "english"
>    if "fr" then return "Francais"
>    else return Locale.languageCode
> }
> 
> The user would see
> english
> Francais
> de
> 
> Obviously it's not perfect. But right now the German language just wouldn't appear.

MikeWallaceDev:
@bolshoytoster I refactored how the original language settings were handled. The enum no longer exists, all the functions have been removed.

You can see how I did it here : #306

Issue #296, opened 2022-02-02
[Bug] Unable to use current Docker image to start AppFlowy-IO


teras:
**Describe the bug**
When I follow the instructions to run the app, although the docker seems to build properly, it is not possible to bring up the service.
Instead I get an error:
```
Attaching to t-app-1
Error response from daemon: failed to create shim: OCI runtime create failed: runc create failed: unable to start container process: exec: "appflowy/frontend/app_flowy/product/0.0.2/linux/Release/AppFlowy/app_flowy": stat appflowy/frontend/app_flowy/product/0.0.2/linux/Release/AppFlowy/app_flowy: no such file or directory: unknown
```

**To Reproduce**
1. Download docker & docker compose files 
2. Build docker containers 
3. Bring them up with the command `docker-compose up`
4. The error appears

**Expected behavior**
To run the app

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser Firefox
 - Version 0.0.2

MikeWallaceDev:
Hi @teras , can you please check if the changes in the dockerfile work for you? So that we can close this issue.

Thank you @lays147 for the fix!

teras:
It works for me now. Thank you

PR #297, opened 2022-02-02
fix appflowy version in dockerfile

lays147:
Hi, based on #296, I debugged and found that the path is using the `0.0.2` version while the last tag is `0.0.3`. So this PR aims to fix that. 

Signed-off-by: Lays Rodrigues <laysrodriguessilva@gmail.com>


appflowy:
@lays147 Thank you! I forgot to update the dockerfile.

PR #311, opened 2022-02-08
Opti protobuf code gen

appflowy:


appflowy:
Optimaze protobuf code generation using build.rs

1. protobuf files will be generated  while compiling if xx.proto changed
2. Only regenerate the protobuf files if the corresponding crate was changed.

TODO: 
1. Auto-generate the x.proto files using build.rs
2. Mark auto-generated protobuf files as git ignore.

PR #319, opened 2022-02-10
Remove `freezed` generated files

hieugao:
Not sure where to update the CI with this change.


appflowy:
Hi, @hieugao, I think you can't just remove the freezed files, you have to set up a script that generates the freezed files while building. Otherwise, the build process will fail.

Let me know if you need help.

Hari-07:
Why is this necessary? Seems like an unnecessary step to remove freezed files only to be added back in CI. And it doesnt increase size that significantly either right?

PR #338, opened 2022-02-16
CI - Generate Freezed files

hieugao:
Additional PR for #319. The reason for the `sed...` line is because I faced this problem -- https://github.com/dart-lang/build/issues/2835#issuecomment-702511142 when generating the files.

I don't know if this is enough, lemme know if anything is missing @appflowy.

Edit: Ooops, didn't realize that Github Workflow uses the newly updated code for CI too, should have updated the code inside #319 instead.


appflowy:
Hi @hieugao, I think you should continue on #319 or close the #319 and bring all the commit to #338.

PR #345, opened 2022-02-17
Reduce docker image size by using docker multi-stage build

PabloCastellano:
Previous image was 7.69GB which is somewhat heavy. Now the resulting image is 1.20GB thanks to multi-stage builds.


appflowy:
@PabloCastellano docker-compose.yml was conflict with main after I merged #346.

PabloCastellano:
@appflowy conflicts solved!

appflowy:
Thank you!

PR #362, opened 2022-02-20
Add en-UK translation

jay-tau:


MikeWallaceDev:
Hi @jay-tau , thanks for contributing to AppFlowy!

Your new language only changes one letter, and would require updating a whole language every time we add a new string. It's too much work for not a lot of advantage.  But also, the original 'authorise' is the correct version for the UK as you can see here : https://grammarist.com/spelling/authorise-authorize/

I'm going to close this PR without merging. But we do thank you :)

PR #374, opened 2022-02-23
Remove `freezed` files

hieugao:
I messed up #319 so I open this new PR. Link #338.

~The reason for the `sed...` line is because I faced this problem -- https://github.com/dart-lang/build/issues/2835#issuecomment-702511142 when generating the files:~
~- [ ] Build Automation for Windows - I don't know what command is equivalent to `sed` in Windows~


appflowy:
Hi,@hieugao. I think you should add the freezed files to `.gitignore` otherwise the files will be added after running the 
`flutter pub run build_runner build --delete-conflicting-outputs`

For example:
image

hieugao:
@appflowy, oops I forgor :skull:, had a feeling that I was missing something, so it was right lol.

Issue #354, opened 2022-02-20
[Bug] Fix DocBanner widget layout issue


annieappflowy:
**Describe the bug**
DocBanner widget layout reports a warning message

**To Reproduce**
1. On debug mode
1. Delete a page
2. Resize the window

**Expected behavior**
No warning message 

**Screenshots**
![image](https://user-images.githubusercontent.com/12026239/154824118-fd433637-d828-418c-bedf-576ea82b9881.png)

PR #375, opened 2022-02-23
fix: DocBanner overflow issue: #354

Pualoo:
Fixed DocBanner widget overflow reported in issue #354 

I used the FittedBox with a BoxFit.scaleDown to preserve the aspect of the feature. Now the FittedBox by the Center Alignment, removing the necessity of Alignment inside of Row.


appflowy:
@Pualoo thanks for working on this. Please comment on https://github.com/AppFlowy-IO/AppFlowy/issues/354 to claim the issue so that we can assign it to you.

PR #393, opened 2022-02-27
Add catalan translation

ikipm:
I have added support for the Catalan language, used in some parts of Spain.


MikeWallaceDev:
Edited commit so it passes commitlint

MikeWallaceDev:
Thank you very much @ikipm !

PR #396, opened 2022-02-28
Add Hungarian translation

peter-gy:
Added support for the Hungarian language.


MikeWallaceDev:
Thank you very much for your translation @peter-gy !

PR #400, opened 2022-03-01
Put business logic in BLoC

MikeWallaceDev:
refactor: Put business logic in BLoC

remove ThrashRepo class
delete trash_repo.dart

add file : trash_listener.dark
move TrashListener class to trash_listener.dart


MikeWallaceDev:
This design was rejected and Nathan and I came up with a better plan.

Issue #481, opened 2022-05-04
[Bug] editing the doc title in the left-panel doesn't sync with the title shown in the editor


annieappflowy:
**Describe the bug**
title syncing between the left panel and the main editor doesn't work properly 

**Expected behavior"
title syncs between these two places

MikeWallaceDev:
Should have closed automatically with fix #485

Issue #480, opened 2022-05-04
[Bug] adding options the colors aren't generated randomly


annieappflowy:
**Describe the bug**
Multi-select / Single select options - the color background isn't generated randomly

**Expected behavior**
Randomize the color options

MikeWallaceDev:
Should have closed automatically with fix #485

Issue #479, opened 2022-05-04
[Bug] Number type float format doesn't work


annieappflowy:
**Describe the bug**
float doesn't work

MikeWallaceDev:
Should have closed automatically with fix #485

Issue #478, opened 2022-05-04
[Bug] Windows open row as a page pop up window too large


annieappflowy:
**Describe the bug**
Opening a row record as a page on Windows too large to block other stuff

**Expected behavior**
A pop-up window over the main editor is displayed. The user can click elsewhere to close the pop-up window

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: v0.0.4-beta.3

MikeWallaceDev:
Should have closed automatically with fix #485

PR #485, opened 2022-05-05
Fix 0.0.4 beta3

appflowy:


appflowy:
close #481 #480 #479 #478

PR #511, opened 2022-05-27
feat: frameless window for mac

vincentdchan:
This PR is implementing a frameless window for AppFlowy on macOS.

**Before**

Normal State:

image

Collapsed State:

image

**After**

Normal State:

image

Collapsed State:

image

## Implementation detail

- Hide the title bar of WindowDelegate in Swift
- Implement movable area in Flutter
  - Send messages through **MethodChannel** to Cocoa to move the window
- Handle collapsing state in Flutter by moving the state from *MenuBloc* to *HomeBloc*

### DIFF
--- a/frontend/app_flowy/lib/workspace/application/home/home_bloc.dart
+++ b/frontend/app_flowy/lib/workspace/application/home/home_bloc.dart
@@ -49,6 +49,9 @@ class HomeBloc extends Bloc<HomeEvent, HomeState> {
         unauthorized: (_Unauthorized value) {
           emit(state.copyWith(unauthorized: true));
         },
+        collapseMenu: (e) {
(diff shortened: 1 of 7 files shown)


appflowy:
Wow, the traffic light looks better now.

frontend/app_flowy/lib/workspace/application/home/home_bloc.dart
```diff
@@ -49,6 +49,10 @@ class HomeBloc extends Bloc<HomeEvent, HomeState> {
         unauthorized: (_Unauthorized value) {
           emit(state.copyWith(unauthorized: true));
         },
+        collapseMenu: (e) {
+          print("press menu");
```
appflowy, line 53: Do not use print directly, use Log.info or Log.trace instead.

PR #515, opened 2022-05-29
Fix various small issues

Poly-Pixel:
This quick PR just fixes a few issues that I noticed when I built and ran AppFlowy on my computer. These include:
 - Setting the color of the add page icon to the theme `iconColor`, which changes with light mode/dark mode.
 - Setting the color of the toolbar buttons in the workspace to the theme `iconColor`, which changes with light mode/dark mode.
 - Changing the background of the toolbar buttons in the workspace to the theme `hover`, which changes with light mode/dark mode.
 - Setting the font style of the primary and secondary button text to the `Btn` text style.
 - Changing the size of the `Btn` text style to `s16` from `s14`, as the text seemed a little small compared with the button size.


appflowy:
@Poly-Pixel Very much appreciate your work! That's awesome.
LGTM

Issue #509, opened 2022-05-26
[Bug] Select options


annieappflowy:
**Describe the bug**
When picking an option in the dropdown list by clicking the area in color, it doesn't work

**To Reproduce**
1. Create two options in a single-select column
2. Switch one option to another by clicking on its capsule 

**Expected behavior**
can select the option by clicking on the capsule area

**Screenshots**
<img width="298" alt="2022-05-26 (1)" src="https://user-images.githubusercontent.com/12026239/170465272-d6b53190-3c27-44bf-aec7-6c29651c7b5b.png">

Issue #508, opened 2022-05-26
[Bug] Number editing


annieappflowy:
**Describe the bug**
Editing an existing value displayed in a number format leads to the disappearance of the value 

**To Reproduce**
1. Set a Number column in US dollar
2. Input 1000000
3. Press enter and it will be displayed as $1,000,000
4. Then re-enter the edit mode and delete a '0'
5. then enter

**Expected behavior**
The number value gets updated and displayed in the same number format

PR #516, opened 2022-05-30
Fix: Grid UI bugs

appflowy:


appflowy:
close #508  
close #509

Issue #523, opened 2022-06-01
[Bug] Renaming window doesn't close when creating another new option


annieappflowy:
**Describe the bug**
While renaming an option value persists, I can create another new option value

**To Reproduce**
1. Rename an option value
2. Enter another value

**Expected behavior**
Renaming window should auto close

**Screenshots**
![image](https://user-images.githubusercontent.com/12026239/171420689-7bd728ef-2c49-405e-b11a-5c4530c51819.png)

appflowy:
> @annieappflowy I would like to work on this issue. @appflowy Can I work on this?

Hi, @moksh-mahajan Thank you for your contribution to AppFlowy. I assigned the ticker for you. You can DM me if you have any questions.

moksh-mahajan:
@appflowy Can we close this one?

annieappflowy:
Thank you for contributing to AppFlowy!

PR #524, opened 2022-06-01
fix: Close renaming window when creating another new option

moksh-mahajan:
The PR is meant to address the issue #523

### DIFF
--- a/frontend/app_flowy/lib/workspace/presentation/plugins/grid/src/widgets/cell/select_option_cell/select_option_editor.dart
+++ b/frontend/app_flowy/lib/workspace/presentation/plugins/grid/src/widgets/cell/select_option_cell/select_option_editor.dart
@@ -156,6 +156,7 @@ class _TextField extends StatelessWidget {
             selectedOptionMap: optionMap,
             distanceToText: _editorPannelWidth * 0.7,
             tagController: _tagController,
+            onClick: () => FlowyOverlay.of(context).remove(SelectOptionTypeOptionEditor.identifier),
             newText: (text) {
               context.read<SelectOptionCellEditorBloc>().add(SelectOptionEditorEvent.filterOption(text));
             },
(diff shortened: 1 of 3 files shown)


frontend/app_flowy/lib/workspace/presentation/plugins/grid/src/widgets/cell/select_option_cell/select_option_editor.dart
```diff
@@ -156,6 +156,7 @@ class _TextField extends StatelessWidget {
             selectedOptionMap: optionMap,
             distanceToText: _editorPannelWidth * 0.7,
             tagController: _tagController,
+            onClick: () => FlowyOverlay.of(context).remove((SelectOptionTypeOptionEditor).toString()),
```
appflowy, line 159: How about create a function to return the identifier then we can reuse the identifier. For example:
static String identifier() {
    return (DateCellEditor).toString();
  }

moksh-mahajan: @appflowy 
If I understood it correctly, you're asking to create a function in the _TextField class itself for returning this particular identifier (SelectOptionTypeOptionEditor) only, right? 
In that case, would the name of the function 'identifier' be appropriate, since it returns only a particular identifier?

appflowy: Create a function identifier() for class, SelectOptionTypeOptionEditor. For example.

static String identifier() {
   return (SelectOptionTypeOptionEditor).toString();
}

So, we can call SelectOptionTypeOptionEditor.identifier()

moksh-mahajan: Ah! Got it. 
Thanks!

PR #546, opened 2022-06-10
refactor: simplify launch settings

MikeWallaceDev:
delete duplicate .vscode folder in app_flowy
rename launches and tasks
Add new build configuration : clean and rebuild all


appflowy:
Awesome! That would be easier to understand. Good job, Mike.

PR #547, opened 2022-06-11
fix: Fix the launch configuration and tasks

MikeWallaceDev:
Fixed the "AF: Build Dart Only" launch configuration.

Renamed the "Flutter Pub" task to "Flutter Pub Get" so it's clearer

Fixed the default build task


appflowy:
Thanks, Mike. LGTM!

PR #568, opened 2022-06-19
Support debugging rust in vscode

tsuiyuenhong:


appflowy:
Hi @tsuiyuenhong. You could update this [doc](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/launcher-and-tasks) for AF: Debug Rust by updating the documentation [repo](https://github.com/AppFlowy-IO/AppFlowy-Docs).

MikeWallaceDev:
Hi, actually, this broke the Linux build :)  Sorry that I didn't see it earlier, but I couldn't check this PR because my code didn't compile...
I had to re-clone the repo in a different place to test this out.
What was wrong with the original code? I'm going to have to revert this locally to make it work, but obviously it was changed because someone else had a problem.

MikeWallaceDev:
I should have added that the reason that this doesn't work, as is, is because "uname -a" is returning a value that the script didn't expect.

PR #603, opened 2022-07-11
feat: add flowy_editor package

LucasXu0:


frontend/app_flowy/packages/flowy_editor/lib/document/node.dart
```diff
@@ -0,0 +1,55 @@
...
+  final LinkedList<Node> children;
+  final Map<String, Object> attributes;
+
+  Node({
+    required this.type,
+    required this.children,
+    required this.attributes,
+    this.parent,
+  });
+
```
appflowy, line 15: typedef xxx = <String, Object> would be better

PR #606, opened 2022-07-12
feat: position and selection

vincentdchan:


frontend/app_flowy/packages/flowy_editor/lib/document/text_delta.dart
```diff
@@ -0,0 +1,399 @@
...
+      return false;
+    }
+    return content == other.content &&
+        mapEquals(_attributes, other._attributes);
+  }
+
+  @override
+  int get hashCode {
+    final contentHash = content.hashCode;
+    final attrs = _attributes;
```
appflowy, line 52: It seems we can use _attributes directly?

frontend/app_flowy/packages/flowy_editor/lib/document/text_delta.dart
```diff
@@ -0,0 +1,399 @@
...
+    return _length.hashCode;
+  }
+}
+
+class _OpIterator {
+  final List<TextOperation> _operations;
+  int _index = 0;
+  int _offset = 0;
+
+  _OpIterator(List<TextOperation> operations) : _operations = operations;
```
appflowy, line 142: It seems like the _OpIterator will change the passed in value's length. Is this ok?
if yes, the iterator should not modify the passed in vec.

PR #607, opened 2022-07-12
Add render plugins

LucasXu0:


frontend/app_flowy/packages/flowy_editor/lib/render/render_plugins.dart
```diff
@@ -0,0 +1,36 @@
...
+  Map<String, NodeWidgetBuilderF> nodeWidgetBuilders = {};
+  // unused
+  // Map<String, NodeBuilder> nodeBuilders = {};
+
+  void register(String name, NodeWidgetBuilderF builder) {
+    nodeWidgetBuilders[name] = builder;
+  }
+
+  void unRegister(String name) {
+    nodeWidgetBuilders.removeWhere((key, _) => key == name);
```
appflowy, line 24: Just return if the name is empty

frontend/app_flowy/packages/flowy_editor/lib/render/render_plugins.dart
```diff
@@ -0,0 +1,36 @@
...
+// unused
+// typedef NodeBuilder<T extends Node> = T Function(Node node);
+
+class RenderPlugins {
+  Map<String, NodeWidgetBuilderF> nodeWidgetBuilders = {};
+  // unused
+  // Map<String, NodeBuilder> nodeBuilders = {};
+
+  void register(String name, NodeWidgetBuilderF builder) {
+    nodeWidgetBuilders[name] = builder;
```
appflowy, line 20: Empty name should not be allowed

frontend/app_flowy/packages/flowy_editor/lib/render/render_plugins.dart
```diff
@@ -0,0 +1,36 @@
...
+    nodeWidgetBuilders.removeWhere((key, _) => key == name);
+  }
+
+  Widget buildWidgetWithNode(Node node) {
+    final nodeWidgetBuilder = _nodeWidgetBuilder(node.type);
+    return nodeWidgetBuilder(node: node, renderPlugins: this)();
+  }
+
+  NodeWidgetBuilderF _nodeWidgetBuilder(String name) {
+    assert(nodeWidgetBuilders.containsKey(name));
```
appflowy, line 33: It would be better if the assert carry out the reason.
For example:
assert(xx, xx)

frontend/app_flowy/packages/flowy_editor/example/lib/plugin/text_node_widget.dart
```diff
@@ -0,0 +1,39 @@
...
+
+class TextNodeBuilder extends NodeWidgetBuilder {
+  TextNodeBuilder.create({required super.node, required super.renderPlugins})
+      : super.create();
+
+  String get content => node.attributes['content'] as String;
+
+  @override
+  Widget build() {
+    final childrenWidget = buildChildren();
```
appflowy, line 12: I prefer children here. Because we are in a 'widget' context.

frontend/app_flowy/packages/flowy_editor/lib/render/render_plugins.dart
LucasXu0: Maybe no, actually, the plugin name can be `''`, and we don't assume the plugin name style.

frontend/app_flowy/packages/flowy_editor/lib/render/render_plugins.dart
LucasXu0: Same answer as previous one.

frontend/app_flowy/packages/flowy_editor/example/lib/plugin/text_node_widget.dart
LucasXu0: done.

frontend/app_flowy/packages/flowy_editor/lib/render/render_plugins.dart
LucasXu0: done.

PR #613, opened 2022-07-14
chore: remove pubspec.lock generated by libraries. 

MikeWallaceDev:
Keep the one in the application


appflowy:
Very nice, We should only keep one.

PR #620, opened 2022-07-18
Fix githooks and remove Husky

MikeWallaceDev:
Our current implementation of githooks and Husky did not work. They did nothing.
My research showed that Husky is a JS script tool, and didn't really fit for our needs. So I removed it.
Native githooks will work fine, but require one line of code to be executed, I will document this in the Environment Setup documentation.

I enabled all of the hooks that @appflowy had setup, they all work now.  However, I think that the pre-push hook might be too much and might need to be removed.

CommitLint also now works, I will add the documentation for that too.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,0 +1,26 @@
+#!/bin/sh
+#
+# An example hook script to check the commit log message.
+# Called by "git commit" with one argument, the name of the file
+# that has the commit message.  The hook should exit with non-zero
+# status after issuing an appropriate message if it wants to stop the
+# commit.  The hook is allowed to edit the commit message file.
+
+echo "Running the AppFlowy commit-msg hook."
+
+# This example catches duplicate Signed-off-by lines.
+
+test "" = "$(grep '^Signed-off-by: ' "$1" |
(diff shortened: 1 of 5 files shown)


appflowy:
The husky should work by running the `cargo make install-commitlint` script, then you're all set. Also, I had documented it in our [guide](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/submitting-code/submitting-your-first-pull-request#commit-message-guidelines).

After running the `cargo make install-commitlint`,  my local hook is active.
image

appflowy:
Is there no need to install husky (`cargo make install-commitlint` does) by replacing it with githook? I mean the commit message will be checked automatically on local.

MikeWallaceDev:
You did document it, that's true, but in the wrong place. So people don't see that documentation when they install the environment.

Everything works now, and is simple.  :)

appflowy:
> You did document it, that's true, but in the wrong place. So people don't see that documentation when they install the environment.
> 
> Everything works now, and is simple. :)

Alright, I will try on my local

> You did document it, that's true, but in the wrong place. So people don't see that documentation when they install the environment.
> 
> Everything works now, and is simple. :)

Alright, that's awesome.

PR #705, opened 2022-07-25
chore: rename structs

appflowy:


appflowy:
Rename flowy-crate event struct's name

PR #706, opened 2022-07-25
fix: add indonesian translation

kakzaki:


appflowy:
Thank you for updating this.

PR #719, opened 2022-07-25
Feat replace commit lint with gitlint

MikeWallaceDev:
I was able to remove the dependencies to : npm, node.js and commitlint
I replaced commitlint with go-gitlint which is an executable
This change does not change the github action at all. Github will still use commitlint.
I made the gitlint parameters match the commitlint parameters.


appflowy:
> I was able to remove the dependencies to : npm, node.js and commitlint I replaced commitlint with go-gitlint which is an executable This change does not change the github action at all. Github will still use commitlint. I made the gitlint parameters match the commitlint parameters.

Awesome! I love that, merged!

PR #723, opened 2022-07-27
Fix YAML syntax in rust_lint workflow

Poly-Pixel:
Quick fix for the rust_lint GitHub workflow, since I noticed that the tests were failing due to invalid YAML syntax.


appflowy:
@Poly-Pixel  Thank you for contributing.  LGTM, merged.

PR #739, opened 2022-07-29
Add Windows development environment install script

Poly-Pixel:
This PR adds a development environment install script for Windows so that AF devs on Windows can get their environment set up easily.


appflowy:
@Poly-Pixel Thank you! I will test it on my computer today.

Issue #741, opened 2022-07-30
[Bug] Overlay demo example app has `ProviderNotFoundException` error


sky1095:
**Describe the bug**
In flowy_infra_ui's example app, under  `Overlay Demo` if you click on `Show List Overlay` and `Show Options Overlay`, both have `ProviderNotFoundException` for AppTheme. And there is a render issue in List Overlay

**To Reproduce**
1. Rund flowy_infra_ui's example app
2. On Demo's page, click on Overlay item
3. On Overlay Demo page, click on `Show List Overlay` and `Show Options Overlay` button.
4. See `ProviderNotFoundException` error

**Expected behavior**
Should render the List overlay and Options overlay
And show list overlay without any render issue

**Screenshots**
1. ProviderNotFoundException
<img width="856" alt="Screenshot 2022-07-31 at 12 19 07 AM" src="https://user-images.githubusercontent.com/16065933/181939604-949f4cac-6372-481d-b2da-b993796794a0.png">

2. Render overflow issue
<img width="856" alt="Screenshot 2022-07-31 at 12 20 05 AM" src="https://user-images.githubusercontent.com/16065933/181940021-c8ff4639-1298-4932-b742-2abb8e1527a6.png">

**Desktop:**
 - OS: MacOS 12.4

PR #744, opened 2022-07-31
fix: ProviderNotFound exception in overlay demo example

sky1095:
This PR is a fix for #741

### DIFF
--- a/frontend/app_flowy/packages/flowy_infra_ui/lib/src/flowy_overlay/overlay_container.dart
+++ b/frontend/app_flowy/packages/flowy_infra_ui/lib/src/flowy_overlay/overlay_container.dart
@@ -18,7 +18,7 @@ class OverlayContainer extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final theme = context.watch<AppTheme>();
+    final theme = context.watch<AppTheme?>() ?? AppTheme.fromType(ThemeType.light);
     return Material(
       type: MaterialType.transparency,
       child: Container(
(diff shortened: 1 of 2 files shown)


frontend/app_flowy/packages/flowy_infra_ui/lib/src/flowy_overlay/overlay_container.dart
```diff
@@ -18,7 +18,7 @@ class OverlayContainer extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final theme = context.watch<AppTheme>();
+    final theme = context.watch<AppTheme?>() ?? AppTheme.fromType(ThemeType.light);
```
thisisamank, line 21: Hardcoding a theme here might bring bugs in other parts of the apps.

sky1095: It is not hardcoded, I just set a default value so if the former comes as null it doesn't break the code.
The better way of doing this is passing the AppTheme from the Widget's constructor so it can be used anywhere without thinking about the provider's value above the widget tree

appflowy: Thank you for contributing. I think it makes sense.

thisisamank: > It is not hardcoded, I just set a default value so if the former comes as null it doesn't break the code. The better way of doing this is passing the AppTheme from the Widget's constructor so it can be used anywhere without thinking about the provider's value above the widget tree

got it, Thanks for explaining.

PR #746, opened 2022-07-31
chore: fix instal_macos.sh download glint error

appflowy:


appflowy:
replace wget with curl

Issue #634, opened 2022-07-19
[FR] Implement FlowyEditor Toolbar component.


LucasXu0:
- [ ] support change document style through selection in the toolbar.
- [ ] plugin system integrates with the toolbar. A developer or a user could add a plugin to the toolbar.

PR #754, opened 2022-08-02
feat: implement bold text in toolbar service

LucasXu0:


LucasXu0:
[#634](https://github.com/AppFlowy-IO/AppFlowy/issues/634)

PR #755, opened 2022-08-02
chore: fix install script errors

annieappflowy:
Case insensitive to accepting the installation of Rust
Use Source to refresh the current window


vincentdchan:
Add some explanations plz.

PR #761, opened 2022-08-03
Feat: copy html from document

vincentdchan:


frontend/app_flowy/packages/flowy_editor/lib/infra/html_converter.dart
```diff
@@ -63,18 +65,44 @@ class HTMLConverter {
...
+      delta.insert(element.text,
+          _getDeltaAttributesFromHtmlAttributes(element.attributes));
     } else if (element.localName == "a") {
       final hyperLink = element.attributes["href"];
       Map<String, dynamic>? attributes;
       if (hyperLink != null) {
         attributes = {"href": hyperLink};
       }
       delta.insert(element.text, attributes);
-    } else if (element.localName == "strong") {
```
appflowy, line 76: Would it be better to use const string?

vincentdchan: It would be a little strange.

PR #775, opened 2022-08-05
feat: optimize selection implement by binary search

LucasXu0:


frontend/app_flowy/packages/flowy_editor/example/lib/main.dart
```diff
@@ -205,4 +216,34 @@ class _MyHomePageState extends State<MyHomePage> {
       child: TextField(),
     );
   }
+
+  Widget _buildFlowyEditorWithBigDocument() {
+    return FutureBuilder<String>(
+      future: rootBundle.loadString('assets/big_document.json'),
+      builder: (context, snapshot) {
+        if (!snapshot.hasData) {
```
appflowy, line 224: I think we should write the normal path through the code first; then write the unusual cases.

if (snapshot.hasData) {
// normal path
} else {
}

frontend/app_flowy/packages/flowy_editor/lib/service/selection_service.dart
```diff
@@ -359,12 +359,14 @@ class _FlowySelectionState extends State<FlowySelection>
...
           panStartOffsetWithScrollDyGap.translate(0, panStartScrollDy! - dy);
     }
-    final nodes = getNodesInRange(panStartOffsetWithScrollDyGap, panEndOffset!);
-    if (nodes.isEmpty) {
-      return;
-    }
-    final first = nodes.first.selectable;
-    final last = nodes.last.selectable;
+
+    final sortedNodes =
```
appflowy, line 363: Nice try!

appflowy (approved):

frontend/app_flowy/packages/flowy_editor/example/lib/main.dart
LucasXu0: I think we should put the code which's body is shorter before.

appflowy: Well, extracting the else condition into a function call would be better.

appflowy: Also, I found the e-version of the code complete. https://nikola-breznjak.com/blog/books/programming/code-complete-2-steve-mcconnell-using-conditionals/

LucasXu0: Got it.

PR #778, opened 2022-08-06
Add icon setting 360

ENsu:


appflowy:
@ENsu LGTM, the icons seem very cute.

Issue #688, opened 2022-07-23
[FR] Tooltip for open row as page


appflowy:
Add a tooltip for opening row as a page.

https://user-images.githubusercontent.com/86001920/180592152-a3981436-de5c-484f-967e-3e57161487a8.mov

ChiragKr04:
Can I work on this issue !?

ChiragKr04:
And what to write on the tooltip!?

appflowy:
> Can I work on this issue !?

Sure. I think `open as a page` would be fine.

ChiragKr04:
> > Can I work on this issue !?
> 
> Sure. I think `open as a page` would be fine.

Created pull request https://github.com/AppFlowy-IO/AppFlowy/pull/781

richardshiue:
Closing as complete since #781 has been merged

PR #781, opened 2022-08-07
Fix: Added tooltip to open row as page

ChiragKr04:
Added tooltip for the open row as page button.


appflowy:
@ChiragKr04 Thank you for contributing so quickly. There is a lint error about the unnecessary import.

image

ChiragKr04:
> @ChiragKr04 Thank you for contributing so quickly. There is a lint error about the unnecessary import.
> 
> image

Alright fixing it.

ChiragKr04:
#688

Issue #767, opened 2022-08-05
[FR] Right click for context menu in sidebar tree


vincentdchan:
<img width="406" alt="image" src="https://user-images.githubusercontent.com/2352832/182989550-0fc0922a-9f3f-47d3-97a1-b914c2170576.png">

<img width="285" alt="image" src="https://user-images.githubusercontent.com/2352832/182989405-e1c416ac-f6a6-46ae-95ca-43dcf2c624ad.png">

**Current:**
When the "more button" is clicked, the dropdown menu is shown. But when the user right clicks on the button, nothing happens.

**Expected:**
When the user right clicks on the button, the dropdown menu shows at the position of the cursor.

aryaman31:
Hey, can I work on this? I have worked on small flutter apps before.

vincentdchan:
Absolutely

vincentdchan:
If you need help. Join our discord group and DM me.

PR #808, opened 2022-08-10
Added right click context menu feature request. 

aryaman31:
This PR is in response to FR https://github.com/AppFlowy-IO/AppFlowy/issues/767. I created a new widget "ViewDisclosureRegion" which is a wrapper for a listener widget. The listener widget is used to detect right clicks on the ActionCell and subsequently display the context menu at the position of the mouse.


vincentdchan (approved):
LGTM

Issue #809, opened 2022-08-10
[FR] Handle `Esc` key in rename dialog


vincentdchan:
1. Right click on the sidebar item.

<img width="265" alt="image" src="https://user-images.githubusercontent.com/2352832/183892998-752e7632-3469-4738-a6cf-7e95464dbddb.png">

2. Click on "rename" button.

<img width="720" alt="image" src="https://user-images.githubusercontent.com/2352832/183893055-a850983c-a1d2-47ce-8d82-4291254c11cc.png">

**Expect:**
After the dialog appears, press the "Esc" key, and the dialog should close.

ChiragKr04:
Can I work on this?

vincentdchan:
Sure

PR #823, opened 2022-08-11
fix: closing rename dialog on escape key press

ChiragKr04:
#809


frontend/app_flowy/packages/flowy_infra_ui/lib/widget/dialog/styled_dialogs.dart
```diff
@@ -3,12 +3,29 @@ import 'package:flowy_infra/size.dart';
...
+          }
+        },
+        child: this,
+      ),
+      context,
+    ).then(
+      (value) {
+        dialogFocusNode.dispose();
+      },
+    );
```
vincentdchan, line 28: Maybe removing the `.then` handler can be simpler:

```dart
    FocusNode dialogFocusNode = FocusNode();
    await Dialogs.show(
      RawKeyboardListener(
        focusNode: dialogFocusNode,
        onKey: (value) {
          if (value.isKeyPressed(LogicalKeyboardKey.escape)) {
            Navigator.of(context).pop();
          }
        },
        child: this,
      ),
      context,
    );
    dialogFocusNode.dispose();
```

vincentdchan:
LGTM

PR #826, opened 2022-08-11
Android build

rileyhawk1417:
This branch is meant to allow AppFlowy to run on android devices.

Currently the app can run, but UI elements need to be resized to cater for mobile devices.


vincentdchan:
Incredible!

rileyhawk1417:
There's a README file in app_flowy/android.
That's where I wrote the documentation, thought it would be best to keep it seperate.

LucasXu0:
I noticed that the OpenSSL library needs to be compiled separately. 

I guess the main reason is that `cargo-ndk` will fail to compile the RustSDK. I found another solution in the `cargo-ndk` issues.

 Maybe try it instead of compiling the OpenSSL library separately.

https://github.com/bbqsrc/cargo-ndk/issues/65

LucasXu0:
Also, Could you integrate the current flow into the compile options of VSCode?

rileyhawk1417:
> I noticed that the OpenSSL library needs to be compiled separately.
> 
> I guess the main reason is that `cargo-ndk` will fail to compile the RustSDK. I found another solution in the `cargo-ndk` issues.
> 
> Maybe try it instead of compiling the OpenSSL library separately.
> 
> [bbqsrc/cargo-ndk#65](https://github.com/bbqsrc/cargo-ndk/issues/65)

Will give it a try, have met that error a lot. That's why I opted to building it seperately.

rileyhawk1417:
> Also, Could you integrate the current flow into the compile options of VSCode?
> 
> 

Will try get this done as soon as possible.

rileyhawk1417:
Just created some tasks in VSCode, so I duplicated the normal ones and changed them to work with android.

frontend/rust-lib/.cargo/config.toml
```diff
@@ -5,4 +5,20 @@
 rustflags=["-C", "link-arg=-mmacosx-version-min=10.11"]
 
 [target.aarch64-apple-darwin]
-rustflags=["-C", "link-arg=-mmacosx-version-min=10.11"]
\ No newline at end of file
+rustflags=["-C", "link-arg=-mmacosx-version-min=10.11"]
+
+[target.aarch64-linux-android]
```
rileyhawk1417, line 10: The NDK location needs to be specified here for cargo ndk to work

frontend/rust-lib/dart-ffi/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 name = "dart_ffi"
 # this value will change depending on the target os
 # default static lib
-crate-type = ["staticlib"]
+crate-type = ["staticlib", "cdylib"]
```
rileyhawk1417, line 11: If cdylib is not specified it will not build the SDK with the .so extension. Would maybe need to modify the config in Makefile.toml

appflowy: > If cdylib is not specified it will not build the SDK with the .so extension. Would maybe need to modify the config in Makefile.toml

image
Check out the frontent/Makefile.toml, the CRATE_TYPE will be modified on a different platforms.

rileyhawk1417:
Tried to make it work with cargo make. Might still have to do some more testing.

rileyhawk1417:
Will close this PR and open a new one with the commits squashed.

PR #828, opened 2022-08-11
fix: #827

LucasXu0:


appflowy:
You can use any of these keywords to close an issue via commit message:
`close, closes, closed, fix, fixes, fixed, resolve, resolves, resolved`

The message MUST contain a string matching the following pattern: KEYWORD #ISSUE. For example: close #1.

appflowy:
fix #827

PR #391, opened 2022-02-26
chore: use actions-cache to speed up CI.

LucasXu0:

Issue #359, opened 2022-02-20
[CI] Speed up CI using GitHub build matrix


annieappflowy:
Speed up CI using GitHub [build matrix](https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#using-a-build-matrix)

annieappflowy:
> I will give a try

Awesome. I have assigned it to you. If you need help, please reach out to @tsuiyuenhong on Discord.

rubiagatra:
Could you give me some context @annieappflowy which GitHub actions do you need to use the GitHub matrix?
```
.github/workflows
├── ci.yaml
├── commit_lint.yml
├── dart_lint.yml
├── dart_test.yml
├── release.yml
├── release_docker.yml
├── rust_lint.yml
├── rust_test.yml
└── translation_notify.yml
```

And what kind of matrix? Rust version? flutter version? or CPU architecture matrix?

tsuiyuenhong:
> Could you give me some context @annieappflowy which GitHub actions do you need to use the GitHub matrix?
> 
> ```
> .github/workflows
> ├── ci.yaml
> ├── commit_lint.yml
> ├── dart_lint.yml
> ├── dart_test.yml
> ├── release.yml
> ├── release_docker.yml
> ├── rust_lint.yml
> ├── rust_test.yml
> └── translation_notify.yml
> ```
> 
> And what kind of matrix? Rust version? flutter version? or CPU architecture matrix?

Hi, rubiagatra. I thought the title `[CI] Speed up CI using GitHub build matrix` is a little bit puzzling. In the first part, we want to manage the build scripts by matrix, like platforms. And as the second part, we want to reduce the cost time of some scripts, like `ci.yaml`, `dart_test.yaml`, and `rust_test.yaml`. But now I still have no idea how to optimize them. Would you mind doing some research and we keep discussing it after that?

rubiagatra:
Ahh Thank you @tsuiyuenhong, I think its little bit more clear.
Sure I'll explore something this weekend, Maybe on Monday I'll let you know

annieappflowy:
Hi @rubiagatra, any update on this?

rubiagatra:
I have been a little bit busy in the past 3 months, I'll check this on Sunday. If anyone wants to take a stab at it, go ahead!

rubiagatra:
I think for `build` matrix is already done yeah @annieappflowy @LucasXu0? or do you want windows support too?
```
    strategy:
[...]

PR #829, opened 2022-08-11
chore: speed up ci by respecting the cargo deps cache

Binlogo:
Noticed #359 issue, and try to fix. Then, it comes here.

Since #391 uses actions-caches, it speed up a lot. Even better, we could respect the cache of the dependency, it may save a lot of time.

|  **Before**   | **After**  |
|  ----  | ----  |
| image  | image |
| image | image |

Hope it may help. If the ci & dart job works well, I could optimize others deeply.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,11 +3,11 @@ name: CI
 on:
   push:
     branches:
-      - 'main'
-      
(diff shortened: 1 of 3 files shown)


LucasXu0:
LGTM. Thanks for contributing.

Issue #685, opened 2022-07-23
[FR] The color of the highlighted background box is inconsistent


appflowy:
The color of the highlighted background box is inconsistent. The highlight color of `New column` and `New row` is not the same as the column. It should be the same as `Done button `, aka, gray color.

<img width="705" alt="image" src="https://user-images.githubusercontent.com/86001920/180591296-9880591e-12dc-48a0-b740-f7aff51c6de3.png">

<img width="694" alt="image" src="https://user-images.githubusercontent.com/86001920/180591309-3cff4dd0-f52a-480a-a0b0-2595bfe66981.png">

<img width="684" alt="image" src="https://user-images.githubusercontent.com/86001920/180591304-1a0b406e-9c4e-4b08-bd5d-cccf5414dea0.png">

aarhusgregersen:
I personally think it makes sense to keep the buttons in that color, as they are actions, not 'details', as is true for 'Done' here.

In other words: Clicking either, do different things, and therefore it makes sense to keep different colors.

annieappflowy:
I second to @aarhusgregersen

Nike682631:
@appflowy is this issue available to work on?

annieappflowy:
> @appflowy is this issue available to work on?

Assigned

Nike682631:
@annieappflowy I've gone through the docs and have been able to build appflowy. I had a question though. Is it possible to do live editing in appflowy?

Nike682631:
https://github.com/AppFlowy-IO/AppFlowy/pull/842

Nike682631:
@appflowy Please close this issue

PR #842, opened 2022-08-12
fix: fixed hover color #685

Nike682631:
image

Both New Column and New Row buttons now have the same background shade as other column headers on hover


appflowy:
@Nike682631 Thanks for contributing

PR #847, opened 2022-08-14
Feat/grid view lens

appflowy:


appflowy:
WIP PR, merge first in case of giant PR

PR #859, opened 2022-08-16
Feat: document model in rust

vincentdchan:
This PR is implementing the document model in Rust.

Missing:
- Unit tests
- Documentations
- Operation transforming.

### DIFF
--- a/shared-lib/lib-ot/src/core/document/position.rs
+++ b/shared-lib/lib-ot/src/core/document/position.rs
@@ -1,4 +1,4 @@
-#[derive(Clone)]
+#[derive(Clone, serde::Serialize, serde::Deserialize)]
 pub struct Position(pub Vec<usize>);
 
 impl Position {
@@ -20,7 +20,7 @@ impl Position {
             return b.clone();
         }
         // check the prefix
-        for i in 0..(pre_insert_path.len()) {
+        for i in 0..(pre_insert_path.len() - 1) {
             if pre_insert_path.0[i] != b.0[i] {
                 return b.clone();
             }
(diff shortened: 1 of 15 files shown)


shared-lib/lib-ot/tests/main.rs
```diff
@@ -1 +1,147 @@
...
+        let mut tb = TransactionBuilder::new(&document);
+        tb.delete_node_at_path(&vec![0].into());
+        tb.finalize()
+    };
+    document.apply(transaction).unwrap();
+    assert!(document.node_at_path(&vec![0].into()).is_none());
+}
+
+#[test]
+fn test_inserts_nodes() {
```
appflowy, line 46: Try using the script-style test.  For example:
```rust
  let scripts = vec![
        MoveRow {
            from_group_index: 0,
            from_row_index: 0,
            to_group_index: 1,
            to_row_index: 1,
        },
        AssertGroupRowCount {
            group_index: 0,
            row_count: 1,
        },
        AssertGroupRowCount {
            group_index: 1,
            row_count: 3,
        },
        AssertRow {
            group_index: 1,
            row_index: 1,
            row: group.rows.get(0).unwrap().clone(),
        },
    ];
    test.run_scripts(scripts).await;
```

You could check out the enum `GroupScript` for more information.

shared-lib/lib-ot/src/core/document/position.rs
```diff
@@ -0,0 +1,46 @@
...
+        self.0.is_empty()
+    }
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+}
+
+impl Position {
+    // delta is default to be 1
+    pub fn transform(pre_insert_path: &Position, b: &Position, delta: i64) -> Position {
```
appflowy, line 15: Would it be better to replace delta with offset? Because we have a struct named delta.

PR #866, opened 2022-08-17
chore: add shortcut to exit the FlowyOverlay

appflowy:


appflowy:
Press Esc to exist the FlowyOverlay

PR #874, opened 2022-08-18
Update fr-FR transation to support grid

yodatak:


DevNono:
Hi, I would like to suggest modifications : use the singular of "nombre" instead of "nombres" (l. 160) and thus, change "Format des nombres" into "Format du nombre" (l. 164). Indeed, I think that the case will only store 1 number and not several.

yodatak:
I juste updated with your changes @DevNono  thanks !

PR #882, opened 2022-08-20
Feat/board configuration

appflowy:


appflowy:
This PR contains the following changes: 
1. Save the board column order
2. Fix bugs when dragging the card to other columns
3. Add more unit tests for group operations.
4. Delete some unused codes

PR #903, opened 2022-08-25
Feat: appflowy_popover

vincentdchan:
A Popover can be used to display some content on top of another.

## Features

> A popover is a transient view that appears above other content onscreen when you tap a control or in an area. Typically, a popover includes an arrow pointing to the location from which it emerged. Popovers can be nonmodal or modal. A nonmodal popover is dismissed by tapping another part of the screen or a button on the popover. A modal popover is dismissed by tapping a Cancel or other button on the popover.

Source: [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/).

- Basic popover style
- Nested popover support
- Exclusive popover API


IMPranshu:
Hey, @vincentdchan, I see that you are currently working on this PR.

I wanted to ask does this PR closes any specific issue from the issue section. Can you please mention that issue here so that the issue will be tracked and closed automatically?

frontend/app_flowy/lib/plugins/board/presentation/toolbar/board_toolbar.dart
```diff
@@ -47,14 +48,22 @@ class _SettingButton extends StatelessWidget {
...
+      child: FlowyIconButton(
+        hoverColor: theme.hover,
+        width: 22,
+        onPressed: () {},
+        icon: Padding(
+          padding: const EdgeInsets.symmetric(vertical: 3.0, horizontal: 3.0),
+          child: svgWidget("grid/setting/setting"),
+        ),
       ),
+      popupBuilder: (BuildContext popoverContext) {
```
appflowy, line 62: The popupBuilder can provide a built-in OverlayContainer, develop can decide whether to use it.
 If so, there is no need to wrap every popup child with OverlayContainer.

frontend/app_flowy/lib/plugins/board/presentation/toolbar/board_setting.dart
```diff
@@ -166,3 +135,48 @@ extension _GridSettingExtension on BoardSettingAction {
...
+    Key? key,
+    required this.settingContext,
+  }) : super(key: key);
+
+  @override
+  State<StatefulWidget> createState() => _BoardSettingListPopoverState();
+}
+
+class _BoardSettingListPopoverState extends State<BoardSettingListPopover> {
+  bool _showGridPropertyList = false;
```
appflowy, line 152: Seems like we created implicit dependencies here. We assume the BoardSettingListPopover must be used as a popover. Would it be better if we create a BoardSettingList and let the developer decide how to show it?

appflowy: I think there is no need to use the stateful widget. We can let the `popupBuilder` decide whether to rebuild it or not. Just like the `BlocBuilder` does. 

Also, I saw lots of stateless widgets are refactored to stateful widgets in order to keep the  `PopoverController`. So we can use   _popover.show(). But it seems very invasive.

I prefer something like context.read<PopoverController>.show(). Just like the `BlocProvider` does.

frontend/app_flowy/lib/plugins/board/presentation/toolbar/board_toolbar.dart
vincentdchan: The popover widget provides the functionalities that the popover needs.
It's better not to define the style of the popover. 
`OverlayContainer` embeds the Theme and the style of the popover.
Especially the `AppTheme`, this class can't be added to the third-party package.
If the user doesn't want to warp every into a container, they can warp the popover into another popover such as a `AppFlowyStylePopover` in the `flowy_infra_ui` package.

frontend/app_flowy/lib/plugins/board/presentation/toolbar/board_setting.dart
vincentdchan: Most of the time, the parent widget doesn't need to pass a `PopoverController`. They use `triggerActions` to control the popover.
The controller's style matched the style of flutter such as `InputController`.

And it can't be refactored into a `context.read<PopoverController>` because the `PopoverController` is held by a parent widget to open a child widget. It's not used in the popover child. If a child popover wants to close itself, it can call:

```
PopoverContainerState.of(popoverContext).closeAll();
```
Example: `app_flowy/lib/plugins/grid/presentation/widgets/header/type_option/date.dart`

frontend/app_flowy/lib/plugins/board/presentation/toolbar/board_toolbar.dart
appflowy: > AppFlowyStylePopover

`AppFlowyStylePopover` is exactly what I mean, we should use AppFlowyStylePopover which is just a wrapper of `Popover`, but it implements the appflowy style. The caller does not need to wrap the OverlayContainer anymore.

frontend/app_flowy/lib/plugins/board/presentation/toolbar/board_setting.dart
appflowy: > Most of the time, the parent widget doesn't need to pass a `PopoverController`. They use `triggerActions` to control the popover. The controller's style matched the style of flutter such as `InputController`.
> 
> And it can't be refactored into a `context.read<PopoverController>` because the `PopoverController` is held by a parent widget to open a child widget. It's not used in the popover child. If a child popover wants to close itself, it can call:
> 
> ```
> PopoverContainerState.of(popoverContext).closeAll();
> ```
> 
> Example: `app_flowy/lib/plugins/grid/presentation/widgets/header/type_option/date.dart`

Got it.

appflowy: Would it be better if we use `PopoverContainer.of(context).closeAll()` because the state of PopoverContainer should be private

vincentdchan: Reasonable

PR #944, opened 2022-08-30
chore: add edit card button

appflowy:


appflowy:
Enable click the edit button on the card to enter edit mode.

Issue #911, opened 2022-08-25
[FR] Implement embed code 


LucasXu0:
**Feature Scope**
- [ ] Implement embed code
  - [ ] [Figma UI](https://www.figma.com/file/X1R9i5HAFtJSNIKOPPW7yK/AppFlowy-%5BUI%5D%5BMaster%5D?node-id=862%3A411799)
  - [ ] change raw text into embed code by
    - [ ] clicking the icon in the toolbar menu
    - [ ] pressing the command or control + E
 - [ ] Optional: convert text surrounded by \`\` to code, like \`abc\` to `abc`

**Test cases**
- [ ] add test cases

rileyhawk1417:
Hi can I try and take this issue?

LucasXu0:
#985

PR #976, opened 2022-09-03
Implement Embed code

rileyhawk1417:
This PR is meant to solve some of the tasks on issue #911.

### What has been done:

- [x] Implement the icon in the toolbar.
- [x] Should apply style after option has been clicked.
- [x] "Embed Code" has been added in "/" menu command.

### What's left:

- [ ] Keyshortcuts.
- [ ] Wrapping the text in "``abc``" then applying the style.


frontend/app_flowy/packages/appflowy_editor/lib/src/render/rich_text/embed_code.dart
```diff
@@ -0,0 +1,75 @@
...
+import 'package:appflowy_editor/src/editor_state.dart';
+import 'package:appflowy_editor/src/infra/flowy_svg.dart';
+import 'package:appflowy_editor/src/render/rich_text/default_selectable.dart';
+import 'package:appflowy_editor/src/render/rich_text/flowy_rich_text.dart';
+import 'package:appflowy_editor/src/render/rich_text/rich_text_style.dart';
+import 'package:appflowy_editor/src/render/selection/selectable.dart';
+import 'package:appflowy_editor/src/service/render_plugin_service.dart';
+import 'package:flutter/material.dart';
+
+class EmbedCodeTextNodeWidgetBuilder extends NodeWidgetBuilder<TextNode> {
```
LucasXu0, line 11: I don't think we need to create a new node widget to process the embed code. The embed code style should be like the bold, italic style. We should integrate this feature into rich_text.dart.

LucasXu0:
Hi Sean, embed code should not exist in the "/" menu because it is a style like bold or italic, not heading or quote.

PR #977, opened 2022-09-04
Feat/support grouping by different field

appflowy:


appflowy:
1. Create an aggregate GridFieldContext to wrap any related field objects.
2. Support grouping by Multi-Select field.
3. Add a default group if the FieldType doesn't support grouping.

PR #978, opened 2022-09-04
chore: support group by checkbox field

appflowy:


appflowy:
Support group by checkbox

PR #979, opened 2022-09-04
Embed code test

rileyhawk1417:
This is a test case that tests the toolbar widget  and if the style is applied.


rileyhawk1417:
Will close this PR and open another PR that has more tests for the toolbar_items.

PR #1014, opened 2022-09-09
Refactor/document tree

appflowy:


appflowy:
1. Remove Box<> in DocumentOperation
2. Impl Into<Path> for &[usize], size, etc. to support accept generic params
3. Rename Position to Path
4. Private the inner type Arena and root node id

vincentdchan (approved):

PR #1015, opened 2022-09-09
Fix: UI issues of popover

vincentdchan:
Merge this after: #991


frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_type_list.dart
```diff
@@ -25,7 +26,7 @@ class FieldTypeList extends StatelessWidget with FlowyOverlayDelegate {
         fieldType: fieldType,
         onSelectField: (fieldType) {
           onSelectField(fieldType);
-          FlowyOverlay.of(context).remove(FieldTypeList.identifier());
+          PopoverContainerState.of(context).closeAll();
```
appflowy, line 29: Would it be better if we use PopoverContainer?

vincentdchan: Sounds better

PR #1017, opened 2022-09-10
fix: empty create option in select/multi-select properties

richardshiue:
An empty create option sometimes appears in grid and kanban select/multi-select. To reproduce, type something into the text box, clear it, and then choose an option.

https://user-images.githubusercontent.com/71320345/189468722-56fbebab-7b61-457f-9bd6-263d391f82ec.mp4


appflowy:
@richardshiue Thank you for contributing! 

LGTM.

appflowy:
> An empty create option sometimes appears in grid and kanban select/multi-select. To reproduce, type something into the text box, clear it, and then choose an option.
> 
>  bug.mp4

Very nice demonstration. 👍🏻

PR #1018, opened 2022-09-10
refactor: replace delta with node body

appflowy:


appflowy:
Replace Delta with NodeBody in Node struct to obtain flexibility when using the other body type.

PR #1032, opened 2022-09-11
fix: prevent select property from having identical options

richardshiue:
When using the edit property menu (from the grid header or the grid settings buttons at the top) to edit single or multi-select properties, identical options can be created, as shown below:

https://user-images.githubusercontent.com/71320345/189540529-cf73f8f2-aa7d-4316-8647-29376186cf1d.mp4


appflowy:
@richardshiue LGTM!

PR #1080, opened 2022-09-17
fix: some improvements for when including time in a date cell

richardshiue:
- I think one o'clock is better than noon for time format hinting
- the cell's time should show/hide immediately according to the switch
- I added autofocus to the text field when the switch is toggled
- prevent the date from disappearing when the time input is invalid. (see below)
- prevent the time field from showing a whitespace at the front after submitting (see below). The cell's date and time spacing is actually reduced by one space as well, but it can be re-added separately.

https://user-images.githubusercontent.com/71320345/190865573-0af932d2-135b-47ea-9139-f450d9bbaecc.mp4

### DIFF
--- a/frontend/app_flowy/assets/translations/en.json
+++ b/frontend/app_flowy/assets/translations/en.json
@@ -222,8 +222,8 @@
   "document": {
     "menuName": "Doc",
     "date": {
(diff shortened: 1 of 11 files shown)


annieappflowy:
Thanks @richardshiue . LGTM

PR #1083, opened 2022-09-18
chore: update issue templates to forms

Poly-Pixel:
This PR replaces the issue templates with the new issue forms, which will improve the bug reporting and feature requesting process.


annieappflowy:
Thanks @Poly-Pixel  for making the templates better

PR #1087, opened 2022-09-18
chore: move vscode config to repo root

Poly-Pixel:
This PR moves the vscode config folder to the repository root so that people can use the launch configurations without having to open a nested folder within the repository.


appflowy:
Hi @Poly-Pixel , thank you for contributing.

For the moment, the root just contains two folders and some other config files.
- AppFlowy
   -  frontend
   -  shared-lib

But, that might be some other folders. For example, if we implement AppFlowy using JS. In my option, keeping the config file in the frontend is ok.

LucasXu0:
Thanks for your proposal. 

At this time,  I don't think it's necessary to move the `.vscode` config to the root directory. If we add the `.vscode` config for the `shared-lib` in the future, we will have to add many configs in the root directory. I think keeping the current is acceptable.

Poly-Pixel:
Ok, thanks for your feedback!

Issue #700, opened 2022-07-23
[Bug] Language selector does not save language when reopening


sunny7dusk:
**Describe the bug**
When using the language selector to use a language, UI changes to reflect it. However, if you close the app and reopen it, the language selector remains at  the previously chosen language but the language of the UI returns to English.

**To Reproduce**
1. Go to settings through the settings icon on the top left, then pick the language option on the sidebar
2. Pick any non-English language. You should see the UI updating the locale.
3. Close and reopen the app.
4. When the app reopens, you should see that the language has returned to English.
5. Navigate to settings -> language again, you should see the language selector is at the language previously selected.

**Expected behavior**
When choosing a new language, it should have saved the locale even after closing and opening a new session.

**Screenshots**
This is when I reopen Flowly after changing language.
![image](https://user-images.githubusercontent.com/71191616/180616808-6432896f-24f4-45a9-bbdc-80a7ebdcf7fc.png)

**Desktop (please complete the following information):**
 - OS -> Windows 10 with WSL2

annieappflowy:
@sunny7dusk would you like to work on this?

sunny7dusk:
I can take a look at this @annieappflowy

annieappflowy:
@sunny7dusk , any update on this? If you no longer work on this issue, please let me know so I can release it to other contributors.

sunny7dusk:
Sorry about that, can look into this soon

sunny7dusk:
I have a PR up for this issue @annieappflowy

annieappflowy:
@sunny7dusk , thank you. It is merged? Can you share the link?

annieappflowy:
Great work. Closing this issue.

PR #1089, opened 2022-09-18
fix: Fixed locale bug when oppening - 

sunny7dusk:
#700 
If case before in `readLocaleWhenAppLaunch` and `setLocale` stopped `AppearanceSettingModel` from loading the saved locale during opening load.

## Testing
1. Go to `Settings` and then `Language`
2. Pick any language, let's say `Bahasa`
3. Close and reopen the app
4. The app should now load with `Bahasa` as the starting language


appflowy:
@sunny7dusk  Thank you for fixing this. LGTM

PR #1097, opened 2022-09-20
Refactor/appflowy overlay

appflowy:


appflowy:
1. Refactor the appflowy_popover by following the package [guide line](https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package)
2. Enable remove the popover layer one by one
3. Fix some bugs

PR #1108, opened 2022-09-21
chore: move russian translation file from CRLF to LF

richardshiue:


appflowy:
Is there any issues if we don't translate to LF?

richardshiue:
> Are there any issues if we don't translate to LF?

No it's just for consistency's sake.

appflowy:
If someone uses Windows to modify the file, then it will convert to CRLF. I think It is hard to maintain consistency.

richardshiue:
Ahh I see. I was thinking that that wouldn't happen. Closing as completed.

PR #1109, opened 2022-09-21
fix: tap event conflict when dragging grid header cell

appflowy:


appflowy:
When set needsLongPressDraggable to false, the drag event conflicts with the AppFlowyPopover tap gesture. Set the popover trigger flag to `none` and let the `GridFieldCell` show the popover manually.

PR #1115, opened 2022-09-22
chore: add install gitlint script & ignore body length

appflowy:


appflowy:
1. Add enable git hook script
2. ignore the body length of the commit message

PR #1132, opened 2022-09-22
Fix blue tint

richardshiue:
This fixes the blue tint, which is used in single/multi-select option labels and is currently the same color as aqua.The design doc seems to have made a mistake. I used a screen color picker to get the color, and have tested on the other colors beforehand to make sure that the code is correct.

Screenshot_20220923_012230


appflowy:
Wow, I didn't notice the design specs also have bugs. LOL

Thank you for fixing this.

PR #1133, opened 2022-09-22
fix: remove keyword when click selection menu item

flappyBug:
When clicking a selection menu item instead of press `enter` key, the keyword (e.g. `/bul`) was not removed. This pr fixes that and adds tests for it.


LucasXu0:
Hi, @flappyBug . Thanks for contributing. 

Moving the `_deleteToSlash` function into the `SelectionMenuItem` doesn't seem appropriate, making the `SelectionMenuItem` difficult to test and redundant with business logic. Would you mind finding a better way to fix this issue?

flappyBug:
Hi @LucasXu0 , thanks for the comment. I just saw the widget test was broken. It's obvious that we need to apply a decorator pattern to wrap keyword deletion on top of the original handler.  I have considered putting the deletion process to

1.  `SelectionMenuItemWidget`'s `onPressed` handler
2. `SelectionMenuItem`'s handler (as shown in my commit)
3. every function passed as a handler (that will cause lots of code redundancy)

The test is failed due to the improper `editorState`, and the `SelectionMenuItemWidget` is already depend on it `editorState` before my pr. The test is illness because the handler of the test cannot interact with the service under `editorState` without pump `AppflowyEditor` in the test.
 
My suggestion is to remove the unit test for `SelectionMenuItemWidget`, as `selection_menu_widget_test.dart` should already covered it, unless we can mock the `editorState`.

LucasXu0:
It sounds reasonable. The `SelectionMenuItemWidget` is not functionally pure already. Removing the test is acceptable.

PR #1140, opened 2022-09-23
grid: fix pressing enter on an option that already exists recreates it

richardshiue:
https://user-images.githubusercontent.com/71320345/191922686-d486dcaf-893e-4f8a-9974-92237ed05092.mp4

The new logic is as follows:
If there aren't any matches, then create it.
If there is an unselected option that matches, then select it and clear filter.
If there is a selected option that matches, then clear filter.

The logic works as intended, but it seems messy. How can I make it better?

### DIFF
--- a/frontend/app_flowy/lib/plugins/grid/application/cell/select_option_editor_bloc.dart
+++ b/frontend/app_flowy/lib/plugins/grid/application/cell/select_option_editor_bloc.dart
@@ -54,6 +54,9 @@ class SelectOptionCellEditorBloc
           selectOption: (_SelectOption value) {
             _onSelectOption(value.optionId);
           },
+          trySelectOption: (_TrySelectOption value) {
+            _trySelectOption(value.optionName, emit);
+          },
           filterOption: (_SelectOptionFilter value) {
             _filterOption(value.optionName, emit);
           },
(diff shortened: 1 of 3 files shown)


rileyhawk1417:
@richardshiue looks great thanks for contribution. Am sure there's a way to try refactor the code and improve it somehow :)

appflowy:
> 

Well, I think it would be better to add documentation in the code.
```dart
// If there aren't any matches, then create it.
if (matchingOption == null && !optionExistsButSelected) {
  context
      .read<SelectOptionCellEditorBloc>()
      .add(SelectOptionEditorEvent.newOption(tagName));
  return;
}

// If there is an unselected option that matches, then select it and clear filter.
 if (matchingOption != null) {
  context.read<SelectOptionCellEditorBloc>().add(
      SelectOptionEditorEvent.selectOption(matchingOption.id));
}

// If there is a selected option that matches, then clear filter.
context
      .read<SelectOptionCellEditorBloc>()
      .add(SelectOptionEditorEvent.newOption(tagName));
      .add(const SelectOptionEditorEvent.filterOption(''));
```

Or we can put all the logic from lines 157 to 181 in the SelectOptionCellEditorBloc to keep the UI simple. You could DM me if you want to know the details.

appflowy:
LGTM!

PR #1142, opened 2022-09-23
Fix/UI adjust

appflowy:


appflowy:
1. Adjust the delete button UI of FieldEditor
2. Set color to the create view panel cell's  icon
3. Rounded button hover

PR #1168, opened 2022-09-26
fix: improve look of select option list

richardshiue:
Before:
Screenshot_20220926_195939

After:
Screenshot_20220926_200023


appflowy:
@richardshiue That's awesome!

PR #1187, opened 2022-09-27
feat: Support markdown to bold text

tekdel:
To bold text, add two asterisks or underscores before and after a word or phrase. To bold the middle of a word for emphasis, add two asterisks without spaces around the letters.

image

https://user-images.githubusercontent.com/5417051/192629293-c8a421ab-c1df-48b7-ac9c-ab89f2b4c931.mp4

Reference:
https://github.com/AppFlowy-IO/AppFlowy/pull/1099


frontend/app_flowy/packages/appflowy_editor/test/infra/test_raw_key_event.dart
```diff
@@ -139,6 +139,12 @@ extension on LogicalKeyboardKey {
     if (this == LogicalKeyboardKey.keyZ) {
       return PhysicalKeyboardKey.keyZ;
     }
+    if (this == LogicalKeyboardKey.asterisk) {
+      return PhysicalKeyboardKey.digit8;
+    }
+    if (this == LogicalKeyboardKey.underscore) {
+      return PhysicalKeyboardKey.minus;
```
tekdel, line 146: I'm not sure what should be here, could you advise with that?

frontend/app_flowy/packages/appflowy_editor/lib/src/operation/transaction_builder.dart
```diff
@@ -114,21 +115,17 @@ class TransactionBuilder {
...
   }) {
-    var newAttributes = attributes;
-    if (index != 0 && attributes == null) {
-      newAttributes =
-          node.delta.slice(max(index - 1, 0), index).first.attributes;
-      if (newAttributes != null) {
-        newAttributes = Attributes.from(newAttributes);
-      }
-    }
+    final newAttributes = attributes ?? _getAttributesAt(node, index);
```
tekdel, line 127: I don't think it's the best solution, it's good for now and it works. Maybe we need to introduce a new object to sealing formatting of a previous content. So far this works okay, but I'm open do do it in a different way.

frontend/app_flowy/packages/appflowy_editor/test/infra/test_raw_key_event.dart
LucasXu0: The `minus` is fine. Since I often encountered some error reports, like https://github.com/flutter/flutter/issues/96022, when simulating keyboard pressing in the test before, I didn't use the Flutter built-in keyboard testing function directly but passed the keyboard event to the `keyboard_service`. The mock keyboard events need to correspond to the logical keys and physical keys, so there is no problem to use `minus` here.

PR #1192, opened 2022-09-28
feat: use appflowy styled toggle widget in date settings

richardshiue:
Screenshot_20220928_214319


appflowy:
Using `DaysOfWeekStyle` is a very nice try.

LGTM!

Issue #1071, opened 2022-09-16
[FR] Translate into fr-FR


annieappflowy:
French translation

Reference:
frontend/app_flowy/packages/appflowy_editor/documentation/translation.md

ankur12-1610:
I would like to work on this issue ;)

Defkon1:
Sorry, I just found out this issue; I pushed the fr_FR a few minutes ago

ankur12-1610:
Do we have any other translation that needs to be covered?

Defkon1:
For the editor: https://github.com/AppFlowy-IO/AppFlowy/tree/22d7ade18b8844ff879f692b6037b2348aa82090/frontend/app_flowy/packages/appflowy_editor/lib/l10n

ankur12-1610:
@Defkon1 we don't have issues for them right?
Should I create or are you going to create the issues collectively for all the remaining translations?

Defkon1:
I don't know what is the preferred way in this repo, I am a first time contributor here...

Probably you should create a new issue

PR #1201, opened 2022-10-01
feat: Added fr_FR AppFlowy editor translation

Defkon1:
resolve [#1071](https://github.com/AppFlowy-IO/AppFlowy/issues/1071)

Added fr_FR AppFlowy editor translation


Defkon1:
This should close #1071

PR #1205, opened 2022-10-01
fix: Theme toggle not updating UI

luketg8:
Fix the toggle not updating live in the settings UI when switching between Light/Dark mode


appflowy:
Thank you for contributing!

PR #1209, opened 2022-10-02
feat: double tildes to strikethrough text

alemoreau:


LucasXu0:
LGTM. Thanks for contributing!

Issue #1068, opened 2022-09-16
[FR] Enable using '/' to insert a numbered list


annieappflowy:
**1~3 main use cases of the proposed feature**
As a user who gets used to '/' to insert elements into a page, I would like to use it to insert a numbered list

citricacid2:
Can I work on this one? Thanks!

rileyhawk1417:
Since it's been merged will be closing this issue.

PR #1220, opened 2022-10-03
feat: added numbered list to the slash menu

citricacid2:
solves #1068


LucasXu0:
LGTM. Thanks for contributing!

PR #1228, opened 2022-10-05
Update README.md

raihan71:
- adding logo appflowy that made look nicer
- changing heading little smaller

<strong>Before</strong>
image

<strong>After</strong>
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,11 @@
-<h1 align="center" style="border-bottom: none">
+<h2 align="center" style="border-bottom: none">
     <b>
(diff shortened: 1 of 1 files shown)


raihan71:
my 1st pr hack22 wdyt @appflowy 🙏

annieappflowy:
Asset 6@4x
Asset 10
^ some logo assets for you to use.
Can you make "AppFlowy" next to the logo on the same line? Also a little bit bigger? I don't see a focus in your revised header. Not sure if this looks better than the original one. I may be too subjective.

LucasXu0:
@annieappflowy. Should we update the readme title? Or close it if not necessary.

Issue #1038, opened 2022-09-13
[FR] AppFlowy Editor Roadmap


LucasXu0:
**0.0.5**
- [x] Implement embed code #911
- [x] Implement numbered list #991
- [x] #1053
- [x] Theme switcher (Dark mode)
- [x] #1055

**0.0.6**
- [x] Implement saving and data export
- [x] Desktop side functional validation
  - [x] macOS
  - [x] #1149
- [x] Code block Plugin
- [x] TeX Plugin
- [x] Horizaontal rule Plugin
- [x] Markdown conversation

**0.0.7**
- [x] Desktop side functional validation
  - [x] Linux
  - [x] Windows
- [ ] Optimize the render time for a large document
- [ ] Design scalable copy-paste extensions
  - [ ] For text-plugins: support copying and pasting both the plain text and styled text
  - [ ] For other-plugins: support copying as the customized text

**0.0.8**
- [ ] Mobile platform support
  - [ ] Android
  - [ ] iOS
- [ ] TBD ...

annieappflowy:
Need an updated 6-month roadmap for appflowy_editor
cc: @LucasXu0  @Xazin

PR #1239, opened 2022-10-07
feat: implement horizontal rule

LucasXu0:


LucasXu0:
https://github.com/AppFlowy-IO/AppFlowy/issues/1038

PR #1246, opened 2022-10-07
feat: limit max character length on select/multiselect options

### LINKED ISSUE #1128: [Bug]: No hard limit on maximum length of options in single/multi-select properties
### Bug Description

Title says all. Currently, I can write an essay in the text field and set it as an option. It would be a lot better imo if there were a limit to how long the option name can be. The exact number of this limit, however, deserves more discussion.

### How to Reproduce

1. Open up a grid with a single or multi-select property.
2. Type a lot of text into the text field and press enter
3. Observe that the option is created

### Expected Behavior

The long input should not be allowed.

### Operating System

Linux

### AppFlowy Version(s)

0.0.5.2

### Screenshots

image

### Additional Context

_No response_

vortixhead:
Fixes #1128
Limit is set to 30 characters


appflowy:
Hi, @vortixhead Thank you for contributing.

PR #1261, opened 2022-10-10
fix: Extra spacing in multiselect cell

### LINKED ISSUE #1186: [Bug]: Extra spacing in multiselect cell
### Bug Description

When I select 2 or more options in a multi-select grid cell, the cell wraps to a second line even though there is more than enough to fit everything in the first row.

### How to Reproduce

1. Go to a grid with a multi-select property
2. Select 2 options when that should fit in the first row
3. Observe that there is extra spacing

### Expected Behavior

The tags should fit on the same row

### Operating System

Linux

### AppFlowy Version(s)

v0.0.5.2

### Screenshots

https://user-images.githubusercontent.com/71320345/192566299-f9b28dfb-3db1-4d80-b88b-1e58c974a637.mp4

### Additional Context

_No response_

sahej-dev:
This fixes #1186 .

`spacing` attribute of the `Wrap` being used in `SelectOptionWrap` seems to be causing the issue. I have just removed the use of the `spacing` from the wrap and instead I have used a `Padding` to provide spacing to the children of the `Wrap`.

### DIFF
--- a/frontend/app_flowy/lib/plugins/grid/presentation/widgets/cell/select_option_cell/select_option_cell.dart
+++ b/frontend/app_flowy/lib/plugins/grid/presentation/widgets/cell/select_option_cell/select_option_cell.dart
@@ -210,15 +210,17 @@ class _SelectOptionWrapState extends State<SelectOptionWrap> {
     } else {
       final children = widget.selectOptions.map(
         (option) {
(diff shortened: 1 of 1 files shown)


appflowy:
@sahej-dev  Thank you for contributing!
LGTM

PR #1267, opened 2022-10-12
fix: no status groups order

appflowy:


appflowy:
1. The initial index of the` No status` group is 0.
2. Clear all the groups when grouping by a new field
3. Update unit tests

PR #1282, opened 2022-10-14
horizontal scroll on multiselect options

### LINKED ISSUE #1223: [Bug]: Scrolling in multi-select edit text fields
### Bug Description

The selected tags show up at the start of the search field as shown below. However, the only way to scroll through them is by using the touchpad horizontal scroll (no mouse scrolling and horizontal mouse scroll is very rare, and vertical scroll for touchpad doesn't work either). We can also explore adding mouse drag to scroll through the selected tags.

I am also of the view that there are too many ways of conveying the selected options (the cell itself, the search field and the list view). While I think that on mobile the cell would probably not be visible when editing the cell, on desktop 3 is too many imo.

### How to Reproduce

1. Open a grid cell with multi-select property
2. Select multiple tags or a few tags with long option names
3. Observe that the only way to scroll through the tags is via a horizontal mouse swipe

### Expected Behavior

The tags should be easily scrollable

### Operating System

Fedora Linux 36

### AppFlowy Version(s)

v0.0.5.3

### Screenshots

image

### Additional Context

_No response_

ChiragKr04:
Resolve #1223

### DIFF
--- a/frontend/app_flowy/lib/plugins/grid/presentation/widgets/cell/select_option_cell/text_field.dart
+++ b/frontend/app_flowy/lib/plugins/grid/presentation/widgets/cell/select_option_cell/text_field.dart
@@ -3,6 +3,7 @@ import 'dart:collection';
 import 'package:flowy_infra/size.dart';
 import 'package:flowy_infra/theme.dart';
 import 'package:flowy_sdk/protobuf/flowy-grid/select_type_option.pb.dart';
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:easy_localization/easy_localization.dart';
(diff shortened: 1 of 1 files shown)


appflowy:
@ChiragKr04 Thank you for contributing.
Hi @richardshiue, you can review the code. I tested it, and it works

richardshiue:
The mouse drag works perfectly, though I would've liked to see a mouse vertical scroll being accepted as well. Another small nitpick would be that once I start dragging, the cursor turns into a I-beam.

With that being said, I still wonder if rendering the selected tags in the text field is necessary at all in the first place, as we already have the cell depicting the currently selected options, accompanied by checked option tags in the list below. If this is the design that we want, then LGTM.

Issue #1255, opened 2022-10-09
[Bug]: Better color selection for new options in single/multi-select properties


richardshiue:
### Bug Description

Currently, the color for new tags can often overlap, especially after a previous option is deleted.

### How to Reproduce

1. Go to a multi-select property
2. Create 3 option and remember the colors for each 
3. Delete the second option
4. Create a new option
5. Observe that the new option's color overlaps

### Expected Behavior

The new option's color isn't the same as an existing option's

### Operating System

Fedora Linux 36

### AppFlowy Version(s)

v0.0.5.3

### Screenshots

https://user-images.githubusercontent.com/71320345/194760832-899c854b-8c04-4d94-b90f-ce36e9f563c5.mp4

annieappflowy:
Can this be a good issue for contributors? @appflowy

PR #1318, opened 2022-10-19
fix: avoid overlapping option colors

richardshiue:
fix: #1255


appflowy:
@richardshiue Wow, good job. You modified the right backend code.

LGTM

PR #1333, opened 2022-10-21
Add Swedish translation

eson57:
Hi,
Here's Swedish translation if wanted.


appflowy:
@eson57  Thank you for contributing!

PR #1347, opened 2022-10-24
fix: several grid ui improvements

richardshiue:
- Add separator in edit property for date and number properties
image
image

- Hide "Add option button" when the input field is shown. Also changed text color of "Options"
image

- Increase size of appflowy popover for date editor because it gets cut off when "include time" is enabled
image


appflowy:
@richardshiue Awesome! That looks very nice

Issue #1325, opened 2022-10-21
[Bug] undo Quote doesn't work


annieappflowy:
### Bug Description

I marked some text as code and formatted as Quote then I can't undo Quote via clicking on the Quote icon. 
Same as I inserted a Quote block through '/' 

### How to Reproduce

<img width="512" alt="image" src="https://user-images.githubusercontent.com/12026239/197158189-2bd0ea9b-ae05-4fcb-b0fd-553854af6e98.png">

### Expected Behavior

I can undo the Quote format by clicking on the Quote icon again 

### Operating System

macOS

### AppFlowy Version(s)

integrate_appflowy_editor

alemoreau:
Hi ! I took the liberty of trying to fix this issue in this PR. https://github.com/AppFlowy-IO/AppFlowy/pull/1353 :innocent: 
I'm open to any feedback.

LucasXu0:
> Hi ! I took the liberty of trying to fix this issue in this PR. #1353 😇 I'm open to any feedback.

Thanks. I merged it and added some tests.

PR #1353, opened 2022-10-24
fix: remove attribute if it already exists on the node

alemoreau:
I tried to tackle this issue https://github.com/AppFlowy-IO/AppFlowy/issues/1325.
Hopefully it is the correct way to fix it.


LucasXu0:
LGTM. I added some tests by the way

LucasXu0:
Resolved #1325

PR #1354, opened 2022-10-25
chore: add reorder bloc test

appflowy:


appflowy:
Add bloc tests:
1. Reorder apps
2. Reorder views

PR #1377, opened 2022-10-26
Replaced social media hyperlinked text with icons.

Abubakrce19:
0-I replaced the simple Hyperlinked Website, Discord, Twitter text with with their icons. 
1-This makes our readme.md file more interactive. 
2-Now the user can go to appflowy social media handles with a single click on the particular social media icon, where the user want go!!
Any kind of improvement idea is appreciated
Rich difference:
image
Thank you!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ You are in charge of your data and customizations.
 </p>
 
 <p align="center">
-    <a href="http://www.appflowy.io"><b>Website</b></a> •
-    <a href="https://discord.gg/9Q2xaN37tV"><b>Discord</b></a> •
-    <a href="https://twitter.com/appflowy"><b>Twitter</b></a>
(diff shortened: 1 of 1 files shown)


Abubakrce19:
@annieappflowy 
I request you to review this pr.
Thankyou

annieappflowy:
Hi @Abubakrce19 , thanks for your work! 
I agree that using icons rather than text gives more visual clues. However, I don't think the improvement adds enough benefits to the reader. In my view, the message the text links convey is more succinct, and the look is clean. 
Because of the reason stated above, I decided not to merge this PR.

Abubakrce19:
> Hi @Abubakrce19 , thanks for your work! I agree that using icons rather than text gives more visual clues. However, I don't think the improvement adds enough benefits to the reader. In my view, the message the text links convey is more succinct, and the look is clean. Because of the reason stated above, I decided not to merge this PR.

oh
Its ok

PR #1407, opened 2022-10-31
refactor: appflowy theme system pt. 1

richardshiue:
This is a replacement for PR #259. Among the tasks outlined in the above PR, the first task (port from `context.watch<AppTheme>` to `Theme.of(context)` syntax) and second task (port from provider to bloc) were implemented, while the hydrated bloc task was skipped due to AppFlowy already storing this information inside the DB. I’ve also added some default colors to shared widgets such as `FlowyHover` and `FlowyButton` to reduce duplicate code.

Applying this PR also makes the app’s transition between light and dark themes seem a bit smoother as well.

There is now more usage of Material Design’s color tokens throughout the application. These are mapped from the `AppTheme` tokens that made up the original design specs. Visually, there shouldn’t be any major changes (exception: grid border color to shader 6, divider width to 1 px to match the design specs) or regressions.

Please test abundantly before merging and please let me know if there is something I can improve here.

~~Blocking: UI regression from scrollbars on `AppFlowyPopovers`~~ (fixed)


appflowy:
> This is a replacement for PR #259. Among the tasks outlined in the above PR, the first task (port from `context.watch<AppTheme>` to `Theme.of(context)` syntax) and second task (port from provider to bloc) were implemented, while the hydrated bloc task was skipped due to AppFlowy already storing this information inside the DB. I’ve also added some default colors to shared widgets such as `FlowyHover` and `FlowyButton` to reduce duplicate code.
> 
> Applying this PR also makes the app’s transition between light and dark themes seem a bit smoother as well.
> 
> There is now more usage of Material Design’s color tokens throughout the application. These are mapped from the `AppTheme` tokens that made up the original design specs. Visually, there shouldn’t be any major changes (exception: grid border color to shader 6, divider width to 1 px to match the design specs) or regressions.
> 
> Please test abundantly before merging and please let me know if there is something I can improve here.
> 
> Blocking: UI regression from scrollbars on `AppFlowyPopovers`

Hi @richardshiue. Thank you so much for this, it's not easy to do the theme refactoring. That is a big PR, It takes time to review. I will DM you on Discord If I have any questions.

Issue #1413, opened 2022-11-03
[Bug] Azrty keyboard not opening menu when typing "/".


DonTomasDeLaVega:
### Bug Description

When i type "/", it doesn't open the menu it's supposed to open, but if i switch to a QWERTY keyboard, it suddenly works.

### How to Reproduce

Switch to an azerty keyboard (i.e : french) et type "/" (it's "shift+ ." on a qwerty keyboard).

### Expected Behavior

![image](https://user-images.githubusercontent.com/47381187/199737216-6e7a95b4-9ef2-46b2-b1cb-cba23742518e.png)
Opening this menu

### Operating System

Windows

### AppFlowy Version(s)

Version 0.0.6.2 - 10/30/2022

PR #1419, opened 2022-11-07
fix: #1413 [Bug] Azrty keyboard not opening menu when typing /

LucasXu0:


LucasXu0:
https://user-images.githubusercontent.com/11863087/200308933-68902932-c20b-425f-b048-694d404ca37a.mov

Issue #1436, opened 2022-11-10
[FR]  Support local image import


rileyhawk1417:
### Description

The feature will allow the user to import images from their computer.

### Impact

The user can take advantage of using local images if they don't wish to use network images.

### Additional Context

A separate copy of the image will be placed in the applications directory

annieappflowy:
Thanks @Xazin for your feedback. We'll keep it in mind when designing this feature.

bmcgonag:
I agree, I'd like to have a local image option since AppFlowy runs locally, seems odd to force me to upload an image to a server just to add a url for it to load from.

PR #1440, opened 2022-11-11
[FR] Img file picker WIP

rileyhawk1417:
This PR is to solve the issue #1436


rileyhawk1417:
Will update the tests on this PR and the codebase when the custom storage path feature is implemented.

LucasXu0:
Hey, @rileyhawk1417. I suggest you draw a flow diagram, which helps to sort out the logic. At present, it is obviously not good to save an image in the `build` function. It is better to save the image after selecting it. 

1. Select a local image and filter the unsupported types.
2. Save it into the AppFlowy database folder.
3. Update the image node's attributes.
4. Handle the error logic like the image is deleted by the user from a folder.

LucasXu0:
Same as https://github.com/AppFlowy-IO/AppFlowy/pull/1702

PR #1463, opened 2022-11-17
chore: use theme.of(context) text theme in FlowyText

richardshiue:


appflowy:
Cool, that looks better.

PR #1465, opened 2022-11-20
fix: fix colors for buttons and allow more customization

richardshiue:
The difference is subtle, but notice how the hover color only flashes when the mouse is exiting the button. This patch also defers default font properties of the TextButtons from the child FlowyText if needed.

main:

https://user-images.githubusercontent.com/71320345/202879622-02dfcbd7-7125-497d-b436-915283832b67.mp4

this patch:

https://user-images.githubusercontent.com/71320345/202879633-fa40799e-4054-4044-be8c-1a9c0f53751b.mp4


appflowy:
Wow, I didn't notice this. Nice!

PR #1466, opened 2022-11-20
chore: use theme.of(context) text theme in settings and around app

richardshiue:


frontend/app_flowy/lib/user/presentation/skip_log_in_screen.dart
```diff
@@ -61,22 +60,29 @@ class _SkipLogInScreenState extends State<SkipLogInScreen> {
           mainAxisAlignment: MainAxisAlignment.spaceEvenly,
           children: [
             InkWell(
+              hoverColor: Colors.transparent,
+              onTap: () =>
+                  _launchURL('https://github.com/AppFlowy-IO/appflowy'),
               child: Text(
                 LocaleKeys.githubStarText.tr(),
-                style: TextStyles.general(color: Colors.blue).underline,
+                style: Theme.of(context)
```
appflowy, line 68: This seems a bit verbose. Can we do something like this?
`AFThemeExtension.of(context).textTheme.underline`

richardshiue: I've added a commit to address this!

PR #1477, opened 2022-11-24
chore: port more const textstyles to theme provider styles

richardshiue:


appflowy:
Cool, I just found out I make a mistake I used the TextStyle. HAHA.  (Already replace with TextTheme)

PR #1485, opened 2022-11-25
fix: ghost scrollbar in scrollviews using theme's ScrollBarTheme

richardshiue:
There is a ghost scrollbar even after making the thumb transparent. This seems to be the way to do it, as I've also tried setting the thickness to zero, trackVisibility to false and thumbVisibility to false, all to no avail.

https://user-images.githubusercontent.com/71320345/203930676-eb6dcb05-c513-4505-b9fa-e97d6235b3b6.mp4


appflowy:
I think it's ok to have the scrollbar. It would be better if the color of the scroll bar is darker

Issue #1479, opened 2022-11-24
[FR] Customize Font Size In AppFlowy


LucasXu0:
## Brief
Support customizing the font size in AppFlowy.

## Details
> Please link each commit to the relevant task below.

**Tasks**
- [x] support switch font size with small, medium, and large.
- [x] add a `more` button to the title bar beside the `share` button.
- [x] add an entry for font size switching after clicking the `more` button.

**Test cases**
- [ ] include the relevant tests that match the tasks mentioned above.

### Impact

The end-users who want to switch the font size.

annieappflowy:
Should test across mainstream operating systems

PR #1495, opened 2022-11-28
Customize Font Size In AppFlowy #1479

LucasXu0:


LucasXu0:
Screenshot 2022-11-28 at 18 24 44

frontend/app_flowy/packages/flowy_infra/lib/image.dart
```diff
@@ -1,15 +1,26 @@
 import 'package:flutter/widgets.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 
-Widget svgWidget(String name, {Color? color}) {
-  final Widget svg = SvgPicture.asset('assets/images/$name.svg', color: color);
-
-  return svg;
-}
-
 Widget svgWithSize(String name, Size size) {
```
appflowy, line 4: this function is duplicated with the following function, I think we can remove this function.

LucasXu0: I will remove this function through a new PR. Because there are so many codes had referenced this function.

PR #1539, opened 2022-12-06
[FR] Create a new yellow theme for AppFlowy

MayurSMahajan:
#### Tasks

- [x] Create a mechanism to select from multiple theme options
- [x] Create a new all-colored theme that utilizes all the colors in AppFlowy's logo.

#### Impact

Adding more themes will become much easier.

#### Flow

The previous AppFlowys app only allowed two themes. This theme control was taken care of by the following approach:

    User opens the settings page where they can toggle between the light theme and dark theme.

After the PR is merged, changing the theme will done as follows

    User opens the settings page where they see a list of different themes, from which they select any one.

#### Naming

This PR adds a new ThemeType to the app. It adds the requested yellow theme. I have named the theme "Anne" because when I showed her a bunch of themes, this was the theme she liked the most. Also, she has been very supportive and welcoming during my journey of making my first contribution to AppFlowy. This is my way of saying Thanks.


MayurSMahajan:
I hope it will pass all the tests now, I have taken care of the commit lint.

MayurSMahajan:
Very excited to see all checks have passed.

annieappflowy:
Hi @MayurSMahajan ,

First of all, thanks a lot for creating a community theme for AppFlowy! Great work :D
I expect that this PR will get merged soon! 
Secondly, it's our pleasure to support you, and hope you've received sufficient guidance to move forward so far.
In addition, I'm very flattered to see you would like to name it Anne. Although I love it, I believe it would be much easier for our users to pick if the name gives a hint of the actual theme palette.

Once this PR gets merged, we should create another issue to add more details to community themes. For example, the user should be able to tell which one is a community theme rather than an official AppFlowy theme. It'd be better to show some metadata about the theme, eg: author, published date, and latest updated date. Maybe later, we should add the usage data.

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
```diff
@@ -17,27 +14,46 @@ class SettingsAppearanceView extends StatelessWidget {
...
-    } else {
-      context.read<AppearanceSettingsCubit>().setTheme(Brightness.dark);
+  String getThemeName(ThemeType ty) {
+    switch (ty) {
+      case ThemeType.light:
+        return LocaleKeys.settings_appearance_lightLabel.tr();
+      case ThemeType.dark:
+        return LocaleKeys.settings_appearance_darkLabel.tr();
+      case ThemeType.anne:
+        return "Anne Mode";
```
LucasXu0, line 50: Localize the string by using `en.json`.

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
```diff
@@ -17,27 +14,46 @@ class SettingsAppearanceView extends StatelessWidget {
...
+      default:
+        return "Try Me";
```
LucasXu0, line 52: Throw an exception is better.

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
MayurSMahajan: okay

MayurSMahajan:
I am making some more changes in my PR after discussing with Richard https://github.com/richardshiue. The idea is to implement two theme modes for each separate theme. @LucasXu0 I think I should base my work in top of Richard's PR.

MayurSMahajan:
After seeing the current development in the themes of AppFlowy, I have adjusted my PR accordingly. 

The latest version of flowy has these options in the appearance settings:
image

I have added an option to select a theme type which updates the appearance settings as follows:
image

Now the above one is consistent with the existing option in the settings. But initially, I had planned a listview for the themes which looks like this:
image

@LucasXu0 what do you think is better: the listview or the popover?

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
```diff
@@ -28,6 +30,124 @@ class SettingsAppearanceView extends StatelessWidget {
...
+        Expanded(
+          child: FlowyText.medium(
+            LocaleKeys.settings_appearance_themeType_label.tr(),
+            overflow: TextOverflow.ellipsis,
+          ),
+        ),
+        AppFlowyPopover(
+          direction: PopoverDirection.bottomWithRightAligned,
+          child: FlowyTextButton(
+            _getThemeName(currentThemeType),
```
LucasXu0, line 50: It's better to add an extension function for `ThemeType` to get the name.

```dart
extension ThemeTypeExtension on ThemeType {
  String get name => ...
}
```

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
```diff
@@ -28,6 +30,124 @@ class SettingsAppearanceView extends StatelessWidget {
...
+        return LocaleKeys.settings_appearance_themeType_defaultTheme.tr();
+      case ThemeType.dandelion:
+        return LocaleKeys.settings_appearance_themeType_dandelionCommunity.tr();
+      default:
+        throw Exception("Unknown ThemeType");
+    }
+  }
+}
+
+// class ThemeNameSetting extends StatelessWidget {
```
LucasXu0, line 103: Please remove those commented codes if they won't be used anymore.

frontend/app_flowy/packages/flowy_infra/lib/colorscheme/colorscheme.dart
```diff
@@ -67,6 +68,14 @@ abstract class FlowyColorScheme {
   });
 
   factory FlowyColorScheme.builtIn(String themeName, Brightness brightness) {
+    if (themeName == 'Dandelion Community Theme') {
```
LucasXu0, line 71: Please don't hardcode the name. If the translation has been modified then we will find it hard to locate the error.

LucasXu0:
Hi, @MayurSMahajan. I added some comments. 

By the way, which flutter version do you use? I found that this PR modified the `pubspec.lock`. Under normal circumstances, this part should not be updated

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
MayurSMahajan: okay

MayurSMahajan:
About the flutter version, I use the latest 3.3.8. I know you told me that appflowy uses a previous version, but silly me I thought it won't be a problem if I don't delete any existing code.

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
MayurSMahajan: That depends on the way we want to give choice of selecting themes to the user. The uncommented code is for the consistent popover choice of theme selection and the commented code is for a list view of themes as I have stated in my earlier comments in this PR. So what do you think is better, a popover or a listview. According to your reply, I will have to delete the other one.

LucasXu0: I prefer to popover style. Or you can add a comment above the code to explain why we should keep them.

frontend/app_flowy/packages/flowy_infra/lib/colorscheme/colorscheme.dart
MayurSMahajan: sure

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
MayurSMahajan: So what I have done here is I have created two copies of this method. One is an extension function as you've suggested, that gets a constant string value independent of the translation logic (since we can't access the LocaleKey file from the flowy-infra package). And another variant of the same method that shows the translated string.

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
MayurSMahajan: I too prefer the popover, because it is consistent with the rest of the User Experience on AppFlowy. No problem I have removed the listview comment.

richardshiue (commented):
I don't think we should add a new `ThemeType` enum to do this, since the appearance settings in the backend stores the current theme as a string. We should either make the backend store it as an enum and make the cubit/`AppTheme` use maybe `ThemeTypePB`, or store `themeName` in `AppTheme` in its factory constructor (better choice imo). If we want to have user-facing theme names that are translatable, we should store them in `AppTheme` as well.

LucasXu0:
Hi, @richardshiue. Actually, I have a question about the relationship between `theme` and `theme mode`.  Why do we need to use theme mode? 

IMO, there isn't every theme has dart or light styles. Is it enough that we only use the `theme`, and dark and light as separate themes?

richardshiue:
Hey @LucasXu0! Hope you're doing well

> Is it enough that we only use the `theme`, and dark and light as separate themes?

Right. This was something I decided on in favor of `X light`, `X dark`, `Y light`, `Z dark`, etc. like it was before.

For one, many well-known themes have light and dark versions (monokai, solaris, nord, etc) and the dark variant is something that a lot of designers like to explore, so I think coupling the dark and light variants together is a good thing. Another advantage of this is that it makes implementing the `adapt to system theme` easier and less likely to fail. (e.g. a person have a dark theme and a light theme, but the theme name/identifier somehow don't match. Then, the app doesn't know which theme is the current theme's dark/light counterpart)

> IMO, there isn't every theme has dart or light styles.

Absolutely. We need to allow this behavior, and then communicate to the user that the selected theme doesn't have the light variant or the dark variant when that happens

MayurSMahajan:
> I don't think we should add a new `ThemeType` enum to do this, since the appearance settings in the backend stores the current theme as a string. We should either make the backend store it as an enum and make the cubit/`AppTheme` use maybe `ThemeTypePB`, or store `themeName` in `AppTheme` in its factory constructor (better choice imo). If we want to have user-facing theme names that are translatable, we should store them in `AppTheme` as well.

I am working on this

LucasXu0:
Hi, @richardshiue. Would you like to help to review the latest code?

LucasXu0:
@MayurSMahajan I try to refactor the code, changing some hardcode text to const string, https://github.com/AppFlowy-IO/AppFlowy/commit/a6c770e712f2b03580cc614d8c971712b24170c0. Please check it.

frontend/app_flowy/assets/translations/en.json
```diff
@@ -163,8 +163,14 @@
         "light": "Light Mode",
         "dark": "Dark Mode",
         "system": "Adapt to System"
+      },
+      "themeType": {
```
richardshiue, line 167: Can we use appTheme or theme? ThemeType seems to just appear from nowhere

frontend/app_flowy/assets/translations/en.json
```diff
@@ -163,8 +163,14 @@
...
+        "label": "Theme Type",
+        "defaultTheme": "Default Theme",
+        "dandelionCommunity": "Dandelion Community Theme"
       }
     }
+
```
richardshiue, line 173:
```suggestion

```

frontend/app_flowy/test/unit_test/settings_appearance_view_test.dart
```diff
@@ -0,0 +1,26 @@
...
+          themeTypeSetting.getThemeNameForDisplaying(BuiltInTheme.light);
+      expect(defaultThemeDisplayName,
+          'settings.appearance.themeType.defaultTheme');
+      defaultThemeDisplayName =
+          themeTypeSetting.getThemeNameForDisplaying(BuiltInTheme.dandelion);
+      expect(defaultThemeDisplayName,
+          'settings.appearance.themeType.dandelionCommunity');
+    },
+  );
+}
```
richardshiue, line 26: Testing translation keys is kind of unnecessary imo?

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
```diff
@@ -28,6 +30,76 @@ class SettingsAppearanceView extends StatelessWidget {
   }
 }
 
+class ThemeTypeSetting extends StatelessWidget {
```
richardshiue, line 33: Same as above, ThemeType seems out of place

frontend/app_flowy/packages/flowy_infra/lib/colorscheme/colorscheme.dart
```diff
@@ -1,6 +1,22 @@
+import 'package:flowy_infra/theme.dart';
 import 'package:flutter/material.dart';
-
```
richardshiue, line 2: Please revert. A newline separates external and internal packages

richardshiue (changes requested):
At this moment, I think it's okay to have the display name for the themes be localized using easy localization. However, I'm still a proponent for storing the localized theme names in AppTheme (perhaps in a future PR). Json-derived themes will benefit greatly from this.

frontend/app_flowy/assets/translations/en.json
MayurSMahajan: Okay sure

frontend/app_flowy/test/unit_test/settings_appearance_view_test.dart
MayurSMahajan: Yeah, I agree, actually I have never done bloc testing, so didn't do it.

frontend/app_flowy/packages/flowy_infra/lib/colorscheme/colorscheme.dart
MayurSMahajan: Sure

frontend/app_flowy/assets/translations/en.json
MayurSMahajan: What do u want it to be, appTheme or just theme. I prefer theme.

frontend/app_flowy/lib/workspace/presentation/settings/widgets/settings_appearance_view.dart
MayurSMahajan: Okay

LucasXu0:
> However, I'm still a proponent for storing the localized theme names in AppTheme (perhaps in a future PR). Json-derived themes will benefit greatly from this.

I don't understand this point. Why should we store the localized theme name instead of the hardcode name, like `light` or `dandelion`? And how JSON-derived themes would benefit from that?

richardshiue:
> Why should we store the localized theme name instead of the hardcode name, like `light` or `dandelion`? And how JSON-derived themes would benefit from that?

Sorry, I should have been more distinct. What I had in mind was adding the localized theme names in addition to the hard coded name as part of the theme's metadata (created, modified, author, version, etc.). JSON-derived themes don't have access to localization keys generated during compilation, so there isn't a way for them to localize their theme names.

frontend/app_flowy/assets/translations/en.json
richardshiue: theme sounds good to me

LucasXu0:
Do you mean that we need to add additional properties in the metadata to describe the information of the theme, instead of using the method to convert the `display name`? If so, it makes sense.

richardshiue:
> Do you mean that we need to add additional properties in the metadata to describe the information of the theme, instead of using the method to convert the `display name`? If so, it makes sense.

Exactly. On a second thought, maybe localization isn't the right term, it's just the display name that will be displayed to the user.

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
```diff
@@ -1,35 +1,27 @@
 import 'package:flowy_infra/colorscheme/colorscheme.dart';
 import 'package:flutter/material.dart';
 
-const List<String> builtInThemes = [
-  'light',
-];
+class BuiltInTheme {
+  static const String light = 'light';
+  static const String dandelion = 'dandelion';
+}
```
richardshiue, line 7: What are the advantages of using a class over a list or enum? I initially wanted this to be used to check whether a theme is built in or not

MayurSMahajan:
@LucasXu0 note, me and Richard have decided to remove the translation logic, because it seems unnecessary and cumbersome as the number of themes increases. Instead in a new PR, all the metadata related to the theme will be stored and fetched from a separate class. That will contain a display name for the theme.

PR #1560, opened 2022-12-09
test: fix select transform test name

ENsu:


frontend/rust-lib/flowy-grid/src/services/field/type_options/selection_type_option/single_select_type_option.rs
```diff
@@ -121,7 +121,7 @@ mod tests {
     }
 
     #[test]
-    fn single_select_transform_with_multiselect_option_test() {
+    fn single_select_transform_with_multi_select_type_option_test() {
```
appflowy, line 124: It sounds better!

PR #1566, opened 2022-12-11
chore: deprecate InputTextField

richardshiue:
This PR deprecates the `InputTextField` wrapper around `RoundedInputField` which itself will be deprecated in favor of `FlowyTextField`. It also makes the following changes to `FlowyTextField`:

- change the non-focused border color (check by pressing return after filtering through number formats) from blue to grey
- allow supplying a `FocusNode` instead of creating one in the widget.
- add optional `onEditingComplete` field to override its [default behavior](https://api.flutter.dev/flutter/material/TextField/onEditingComplete.html) if needed. It's overridden in the add select option because focus shouldn't be given up.
- call `setState()` after submitting an input and autoclearing. Without this, the suffix text won't update if the parent widget doesn't get rebuilt, since changing the controller text [doesn't call `onChanged`](https://github.com/flutter/flutter/issues/83171#issuecomment-1126878806). (check by submitting an existing option in the select option header editor)
- don't submit the existing text if focus is lost.


frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
```diff
@@ -135,11 +141,7 @@ class FlowyTextFieldState extends State<FlowyTextField> {
 
   void notifyDidEndEditing() {
     if (!focusNode.hasFocus) {
-      if (controller.text.isEmpty) {
```
appflowy, line 138: If remove these codes here, how to save the text to disk when switching to another textField?

image

richardshiue: Added `submitOnLeave` flag to determine whether to submit when focus is lost.

PR #1567, opened 2022-12-11
refactor: appflowy themes

richardshiue:
- reduce responsibility of AppTheme to only contain colorscheme (and theme metadata in the future). I've moved the font family and the getThemeData function out of the class.
- make the colorscheme more easily extendable


frontend/app_flowy/lib/startup/tasks/app_widget.dart
```diff
@@ -82,8 +83,10 @@ class ApplicationWidget extends StatelessWidget {
         builder: (context, state) => MaterialApp(
           builder: overlayManagerBuilder(),
           debugShowCheckedModeBanner: false,
-          theme: state.theme.getThemeData(state.locale),
-          darkTheme: state.darkTheme.getThemeData(state.locale),
+          theme: getThemeData(
```
appflowy, line 86: Would it be better if call `state.darkTheme `or `state.lightTheme` directly? Just by moving these constructions to the Cubit.

frontend/app_flowy/packages/flowy_infra/lib/colorscheme/colorscheme.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:flutter/material.dart';
+
+import 'default_colorscheme.dart';
+
+@immutable
+abstract class FlowyColorScheme {
```
appflowy, line 6: Yep, `abstract class `is the right tool here.

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
```diff
@@ -1,267 +1,131 @@
...
+  final FlowyColorScheme? lightTheme;
+  final FlowyColorScheme? darkTheme;
+  // static final Map<String, dynamic> _cachedJsonData = {};
+
+  const AppTheme({
+    this.lightTheme,
+    this.darkTheme,
+  });
+
+  factory AppTheme.fromName({required String themeName}) {
```
appflowy, line 25: Create a new factory method for loading JSON in the future.  Maybe `AppTheme.fromJson`

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
```diff
@@ -1,267 +1,131 @@
...
-          caption: textTheme.getFontStyle(
-            fontSize: FontSizes.s11,
-            fontWeight: FontWeight.w400,
-            fontColor: shader3,
-          ),
-        )
-      ],
+  // metadata member
+  final FlowyColorScheme? lightTheme;
+  final FlowyColorScheme? darkTheme;
```
appflowy, line 17: We can remove the `?` from the lightTheme and the darkTheme. Because we must have these.

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
richardshiue: Will this be a public factory? I was thinking that since the backend will only store the theme name, it will be the fromName factory that will determine whether to use a built-in theme or load from json.

frontend/app_flowy/lib/startup/tasks/app_widget.dart
richardshiue: I've added a commit to address this

frontend/app_flowy/packages/flowy_infra/lib/theme.dart
richardshiue: Done!

Issue #1183, opened 2022-09-27
[Test] Implement unit test in rust when switching from a Multi-select property type to a Text property type


appflowy:

0xkelvin:
hello can i take this @annieappflowy @appflowy

annieappflowy:
Thanks @0xkelvin for taking this issue. Just assigned it to you.
In case you need help, please reach out @appflowy on Discord

0xkelvin:
it done and merged #1585

PR #1585, opened 2022-12-19
test: UnitTest Switching from Multi to text

0xkelvin:
#1183


0xkelvin:
> It seems like you don't use the `cargo fmt`. You can enable the` cargo fmt` then it will format your files in Rust style when you save the file.

ok, let me try

appflowy (approved):

Issue #1609, opened 2022-12-26
[Bug] Scrolling to the bottom via the controller doesn't work on macOS


Dev-Owl:
### Bug Description

Using the `scrollToBottom` doesn't work on macOS, the board view doesn't scroll

### How to Reproduce

Use the example of the appflowy_board and click on the + button inside the board footer

### Expected Behavior

The view scrolls the last item into the visible area of the scroll container

### Operating System

macOS

### AppFlowy Version(s)

main

### Additional Context

The problem is caused by the following line inside the `reorder_flex.dart` file:

`alignmentPolicy: ScrollPositionAlignmentPolicy.keepVisibleAtStart,`

Removing the line and use the Flutter default (`ScrollPositionAlignmentPolicy alignmentPolicy = ScrollPositionAlignmentPolicy.explicit`) lets the view work as expected.

PR #1610, opened 2022-12-26
Fix scrollToBottom for macos see #1609

Dev-Owl:
See #1609


appflowy:
Thank you for contributing! 
LGTM

PR #1625, opened 2022-12-31
chore: add spacing between a popover and the edge of the window

richardshiue:
Flutter doesn't currently allow popups to go outside the application window. Because of this, popovers have an unpolished look when they are about go off-canvas. This PR adds a spacing between the popover and the edge of the window.

We can improve this even further by automatically switching to the other side if there isn't enough space on side where the popover is supposed to spawn.

image


appflowy:
Nice try!

Issue #1624, opened 2022-12-30
[FR] Support Shift + Option + Left/Right Arrow


Kaijiro:
### Bug Description

On MacOS, I usually select one or many words with the combination "Shift + Option + Left/Right Arrow". This is a basic text manipulation.
Nothing happens when I use this key combination on AppFlowy.

### How to Reproduce

Write many characters / words and try to select them using only your keyboard.

### Expected Behavior

The words should be highlighted by the selection

### Operating System

MacOs

### AppFlowy Version(s)

0.0.8 (1)

LucasXu0:
Hi, @Kaijiro. Thanks for your suggestion. I implemented the shortcuts for shift+option+arrow left/right in this [PR](https://github.com/AppFlowy-IO/AppFlowy/pull/1648) and will release it in the next version.

emmggi:
Can we perhaps have it select one word instead all until the end/beginning when we do CTRL+SHIFT+ARROW LEFT/RIGHT?
Selection all the way to the end/beginning could be done using SHIFT+HOME/END.

PR #1648, opened 2023-01-04
#1624 Support Shift + Option + Left/Right Arrow

LucasXu0:
https://user-images.githubusercontent.com/11863087/210517210-a0f181aa-2599-44dd-9b07-9610cfc13597.mov

### DIFF
--- a/frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
+++ b/frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
@@ -23,8 +23,8 @@ class SettingsLocation {
 
   String? get path {
     if (Platform.isMacOS) {
-      // remove the prefix `/Volumes/Macintosh HD/Users/`
-      return _path?.replaceFirst('/Volumes/Macintosh HD/Users', '');
+      // remove the prefix `/Volumes/*`
+      return _path?.replaceFirst(RegExp(r'^/Volumes/[^/]+'), '');
     }
     return _path;
   }
--- a/frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
+++ b/frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
@@ -12,10 +12,22 @@ const String kSettingsLocationDefaultLocation =
 
 class SettingsLocation {
   SettingsLocation({
-    this.path,
-  });
+    String? path,
+  }) : _path = path;
 
-  String? path;
+  String? _path;
+
+  set path(String? path) {
+    _path = path;
+  }
+
+  String? get path {
+    if (Platform.isMacOS) {
+      // remove the prefix `/Volumes/Macintosh HD/Users/`
+      return _path?.replaceFirst('/Volumes/Macintosh HD/Users', '');
+    }
+    return _path;
+  }
 
   SettingsLocation copyWith({String? path}) {
     return SettingsLocation(
--- a/frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/arrow_keys_handler.dart
+++ b/frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/arrow_keys_handler.dart
@@ -289,8 +289,50 @@ ShortcutEventHandler cursorRight = (editorState, event) {
   return KeyEventResult.handled;
 };
 
+ShortcutEventHandler cursorLeftWordSelect = (editorState, event) {
+  final nodes = editorState.service.selectionService.currentSelectedNodes;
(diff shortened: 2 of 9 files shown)


frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
```diff
@@ -12,10 +12,22 @@ const String kSettingsLocationDefaultLocation =
...
+  String? _path;
+
+  set path(String? path) {
+    _path = path;
+  }
+
+  String? get path {
+    if (Platform.isMacOS) {
+      // remove the prefix `/Volumes/Macintosh HD/Users/`
+      return _path?.replaceFirst('/Volumes/Macintosh HD/Users', '');
```
appflowy, line 27: Is `Macintosh HD` the same with different computer settings?

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/arrow_keys_handler.dart
```diff
@@ -322,11 +388,30 @@ extension on Position {
       }
       return null;
     }
-    if (node is TextNode) {
-      return Position(path: path, offset: node.delta.nextRunePosition(offset));
-    } else {
-      return Position(path: path, offset: offset);
+    switch (selectionRange) {
+      case _SelectionRange.character:
+        if (node is TextNode) {
```
appflowy, line 393: Would it be better if we create an extension for this? Then we can use node.isTextNode()

frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
LucasXu0: Yes. If the user didn't change the name when reinstalling the system. It will be `Macintosh HD` always.

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/arrow_keys_handler.dart
LucasXu0: I don't think so. Using `node.isTextNode()` needs to import the extension file. I think that using TextNode will be more concise.

appflowy: Just declare the extension in node.dart file.

richardshiue:
This is awesome! Because I can't see the keystrokes, I'm not too sure how many times the shortcut is performed. But did the result of one trigger select the white-space separately when going backward? In my experience, the function would select all the spaces/tabs from a cursor up until the boundary of the previous word, in addition to that word. The forward behavior looks like the right one.

lorem |

 then becomes

|`lorem `

frontend/app_flowy/lib/workspace/application/settings/settings_location_cubit.dart
appflowy: So using regex: `volums/*/User.` would be better.

LucasXu0: Done, using `/Volumes/[^/]+` to match the prefix now.

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/arrow_keys_handler.dart
LucasXu0: Good suggestion. However, if we use `node.isTextNode`, then the compiler won't infer the below node is `TextNode`. So we have to use `is TextNode`.

For example.
```dart
class Foo {
  void bar() {}
}

if (object is Foo) {
  object.bar(); // good.
}

if (object.isFoo) {
  object.bar(); // bad.
}
```

LucasXu0:
> This is awesome! Because I can't see the keystrokes, I'm not too sure how many times the shortcut is performed. But did the result of one trigger select the white-space separately when going backward? In my experience, the function would select all the spaces/tabs from a cursor up until the boundary of the previous word, in addition to that word. The forward behavior looks like the right one.
> 
> lorem |
> 
> then becomes
> 
> |`lorem `

Cool. You're right. I fixed it.

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/arrow_keys_handler.dart
appflowy: Cool

Issue #1603, opened 2022-12-22
[Bug] Inputting the text was accepted under "Number" type property


0xkelvin:
### Bug Description

Switching field property to "Number", but still can input the text in the cell under number property
<img width="410" alt="image" src="https://user-images.githubusercontent.com/56182747/209160251-1f809f7a-ef99-4e58-a4de-e7eddec13f45.png">

### How to Reproduce

Switching the field to "Number" property, then just fill in the text instead of number

### Expected Behavior

shouldn't accept text input under Number property, might be the cell just display empty after press inter, or raise the error message

### Operating System

MacOS M1

### AppFlowy Version(s)

latest

0xkelvin:
let me fix it after i finish the unit test @annieappflowy @appflowy

PR #1650, opened 2023-01-04
feat: handling same as Notion when user fills in the mixing of number and text

0xkelvin:
@appflowy could you please help me review this change
i tried to modify the code to achieve the same behaviours like Notion app for Number type.

the only problem is when i fill in the text under Number type, the backend in Rust processes as an empty string which is correct, But the frontend still display the string on the cell instead of empty the cell.

### DIFF
--- a/frontend/rust-lib/flowy-grid/src/services/field/type_options/number_type_option/number_type_option.rs
+++ b/frontend/rust-lib/flowy-grid/src/services/field/type_options/number_type_option/number_type_option.rs
@@ -7,6 +7,7 @@ use crate::services::field::{
     TypeOptionCellDataCompare, TypeOptionCellDataFilter, TypeOptionTransform,
 };
 use bytes::Bytes;
+use fancy_regex::Regex;
 use flowy_derive::ProtoBuf;
 use flowy_error::FlowyResult;
 use grid_rev_model::{FieldRevision, TypeOptionDataDeserializer, TypeOptionDataSerializer};
@@ -97,10 +98,14 @@ impl NumberTypeOptionPB {
 
     pub(crate) fn format_cell_data(&self, s: &str) -> FlowyResult<NumberCellData> {
         match self.format {
-            NumberFormat::Num => match Decimal::from_str(s) {
-                Ok(value, ..) => Ok(NumberCellData::from_decimal(value)),
-                Err(_) => Ok(NumberCellData::new()),
-            },
+            NumberFormat::Num => {
+                let re = Regex::new(r"[^\d\.]").unwrap();
+                let strnum = re.replace_all(s, "");
+                match Decimal::from_str(&strnum) {
+                    Ok(value, ..) => Ok(NumberCellData::from_decimal(value)),
+                    Err(_) => Ok(NumberCellData::new()),
+                }
+            }
             _ => NumberCellData::from_format_str(s, self.sign_positive, &self.format),
         }
     }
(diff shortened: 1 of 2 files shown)


frontend/rust-lib/flowy-grid/src/services/grid_editor.rs
```diff
@@ -510,6 +510,7 @@ impl GridRevisionEditor {
                 // Update the changeset.data property with the return value.
                 let type_cell_data =
                     apply_cell_data_changeset(cell_changeset, cell_rev, field_rev, Some(self.cell_data_cache.clone()))?;
+                tracing::trace!("xxxxxxxxxxxx:: {}", type_cell_data);
```
appflowy, line 513: Please remove this.

frontend/rust-lib/flowy-grid/src/services/field/type_options/number_type_option/number_type_option.rs
```diff
@@ -97,10 +98,14 @@ impl NumberTypeOptionPB {
 
     pub(crate) fn format_cell_data(&self, s: &str) -> FlowyResult<NumberCellData> {
         match self.format {
-            NumberFormat::Num => match Decimal::from_str(s) {
-                Ok(value, ..) => Ok(NumberCellData::from_decimal(value)),
-                Err(_) => Ok(NumberCellData::new()),
-            },
+            NumberFormat::Num => {
+                let re = Regex::new(r"[^\d\.]").unwrap();
```
appflowy, line 102: You can try lazy_static. For example:
```
lazy_static! {
    static ref URL_REGEX: Regex = Regex::new(
        "[(http(s)?):\\/\\/(www\\.)?a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
    )
    .unwrap();
}
```

frontend/rust-lib/flowy-grid/src/services/grid_editor.rs
0xkelvin: removed

frontend/rust-lib/flowy-grid/src/services/field/type_options/number_type_option/number_type_option.rs
0xkelvin: ah, thanks for suggestion, this is the good point, i was consider lazy_static, but then i think the Regex is not to expensive to compute, so we don't really need initialized lazily. i think so.

appflowy: That depends. 
```
#[cfg(test)]
mod tests {
    use fancy_regex::Regex;
    use std::time::Instant;

    #[test]
    fn time_cost_test() {
        let start = Instant::now();
        let re = Regex::new(r"[^\d\.]").unwrap();
        let duration = start.elapsed();
        println!("Time elapsed in expensive_function() is: {:?}", duration);
    }
}

```
The result is `3ms` test on my computer.
image


So if we want to get 1000 cells' content using the `format_cell_data`, that will be expensive.

0xkelvin: nice, i see, let me switch to use lazy_static!

0xkelvin: done, @appflowy please help me review again

0xkelvin:
fixed : #1603

PR #1651, opened 2023-01-04
Fix mobile sdk

rileyhawk1417:
Fixed broken rust sdk build task for mobile


frontend/app_flowy/lib/main.dart
```diff
@@ -16,11 +17,32 @@ class FlowyApp implements EntryPoint {
 }
 
 Future<void> main() async {
```
LucasXu0, line 19: Since the two functions only differ in the hotkey manager, try to use the below function instead.

```dart
Future<void> main() async {
  WidgetsFlutterBinding.ensureInitialized();
  await EasyLocalization.ensureInitialized();

  if (Platform.isAndroid || Platform.isIOS) {
    // Mobile platform is not supported the hotKeyManager
  } else {
    WidgetsFlutterBinding.ensureInitialized();
    await hotKeyManager.unregisterAll();
  }

  await FlowyRunner.run(FlowyApp());
}
```

frontend/rust-lib/flowy-database/Cargo.toml
```diff
@@ -16,8 +16,8 @@ r2d2 = "0.8.9"
 libsqlite3-sys = { version = ">=0.8.0, <0.24.0", features = ["bundled"] }
 scheduled-thread-pool = "0.2.5"
 error-chain = "=0.12.0"
-openssl = { version = "0.10.38", optional = true, features = ["vendored"] }
-openssl-sys = { version = "0.9.69", optional = true, features = ["vendored"] }
+openssl = { version = "0.10.38",  features = ["vendored"] }
+openssl-sys = { version = "0.9.69",  features = ["vendored"] }
```
rileyhawk1417, line 20: Using the optional flag causes the android build to fail completely. It throws an error saying can't find openssl.

rileyhawk1417:
Updated all android dependencies across the package to use the same version. This is to avoid projects/plugins using different versions and clashing together causing the build to fail.

LucasXu0:
Hey, @rileyhawk1417. I suggest you create a new file called android ci.yml to run the Android task. Don't modify the existing CI files.

rileyhawk1417:
Added task to setup sqlite crate for mobile devices.

rileyhawk1417:
Closing this PR as it's going to be solved by #1850

PR #1652, opened 2023-01-05
fix: update the cell content if input is not valid data

appflowy:
If the initial input content is: "", then after entering an invalid number string, the cell content will be empty. The cellDataNofitier doesn't notify the listener because the new value is the same as the old one. aka, "", the empty string.

Add `listenWhenOnCellChanged` callback to let the caller design how to handle the value change notification.


0xkelvin (approved):
lgtm

PR #1680, opened 2023-01-07
CleanUp: removed unused PhantomSnapshotPersistence-import

itinance:
As written in title. It just removes `use PhantomSnapshotPersistence` as it is not used (anymore).


appflowy:
@itinance Thank you for reporting this.

Issue #1061, opened 2022-09-16
[FR] Support markdown to create a blockquote


annieappflowy:
To create a blockquote, add a > in front of a paragraph.

> Dorothy followed her through many of the beautiful rooms in her castle.

Reference: 
* https://github.com/AppFlowy-IO/AppFlowy/discussions/1100
* https://www.markdownguide.org/basic-syntax/

Lakshikuga:
Hi! Would like to work on this issue.

annieappflowy:
Hi @Lakshikuga , thanks for offering the help. I assigned the issue to you. Your mentor is @LucasXu0  for this task. If you need help from him, please reach him out on [Discord](https://discord.gg/9Q2xaN37tV).

LucasXu0:
Hi @Lakshikuga. How's the progress of this issue? Feel free to DM me on Discord if you need any help.

abichinger:
Hello everyone, I would like to take over this issue, if @Lakshikuga is no longer working on it.

annieappflowy:
Assigned @abichinger 
Thanks for offering the help. Your mentor for this issue is @LucasXu0 . You can reach out him for help on Discord.

PR #1696, opened 2023-01-11
feat: greater to blockquote

abichinger:
solves #1061


LucasXu0:
Hi, @abichinger. I think we need to convert `> ` to quote text, not `>` to quote text. There is a whitespace after `>`.

You can refer to the `whitespace_handler.dart`.

PR #1698, opened 2023-01-12
chore: grid field action sheet layout improvement

richardshiue:
Implements vision of #1673

- different widths for general field actions and type options
- unbounded maxHeight
- fix extra padding
- in the future, if necessary, move SizedBox to FieldEditor to use different widths for each FieldType

### DIFF
--- a/frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_cell_action_sheet.dart
+++ b/frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_cell_action_sheet.dart
@@ -12,6 +12,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:easy_localization/easy_localization.dart';
 import 'package:app_flowy/generated/locale_keys.g.dart';
+import 'package:styled_widget/styled_widget.dart';
 
 import '../../layout/sizes.dart';
 
@@ -31,12 +32,15 @@ class _GridFieldCellActionSheetState extends State<GridFieldCellActionSheet> {
   Widget build(BuildContext context) {
     if (_showFieldEditor) {
       final field = widget.cellContext.field;
-      return FieldEditor(
-        gridId: widget.cellContext.gridId,
-        fieldName: field.name,
-        typeOptionLoader: FieldTypeOptionLoader(
+      return SizedBox(
+        width: 400,
+        child: FieldEditor(
           gridId: widget.cellContext.gridId,
-          field: field,
+          fieldName: field.name,
+          typeOptionLoader: FieldTypeOptionLoader(
+            gridId: widget.cellContext.gridId,
+            field: field,
+          ),
         ),
       );
     }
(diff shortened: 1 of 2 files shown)


frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_cell_action_sheet.dart
```diff
@@ -31,38 +32,39 @@ class _GridFieldCellActionSheetState extends State<GridFieldCellActionSheet> {
...
-                    _showFieldEditor = true;
-                  });
-                },
-              ),
-              VSpace(GridSize.typeOptionSeparatorHeight),
-              _FieldOperationList(widget.cellContext),
-            ],
-          ),
+    return SizedBox(
+      width: 200,
```
appflowy, line 48: If we set the width to 200, a fixed width. then It can not auto-expand to the text width. For example
image

frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_cell_action_sheet.dart
```diff
@@ -31,38 +32,39 @@ class _GridFieldCellActionSheetState extends State<GridFieldCellActionSheet> {
...
+            gridId: widget.cellContext.gridId,
+            field: field,
+          ),
         ),
       );
     }
-    return BlocProvider(
-      create: (context) =>
-          getIt<FieldActionSheetBloc>(param1: widget.cellContext),
-      child: IntrinsicWidth(
```
appflowy, line 46: It provides auto expand width and height

frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_cell_action_sheet.dart
```diff
@@ -31,38 +32,39 @@ class _GridFieldCellActionSheetState extends State<GridFieldCellActionSheet> {
   Widget build(BuildContext context) {
     if (_showFieldEditor) {
       final field = widget.cellContext.field;
-      return FieldEditor(
-        gridId: widget.cellContext.gridId,
-        fieldName: field.name,
-        typeOptionLoader: FieldTypeOptionLoader(
+      return SizedBox(
+        width: 400,
```
appflowy, line 36: I think 300 would be better for display.

frontend/app_flowy/lib/plugins/grid/presentation/widgets/header/field_cell_action_sheet.dart
```diff
@@ -92,18 +93,30 @@ class _EditFieldButton extends StatelessWidget {
...
+          HSpace(GridSize.typeOptionSeparatorHeight),
+          cells[1],
+        ],
       ),
-    );
+      VSpace(GridSize.typeOptionSeparatorHeight),
+      Flex(
+        direction: Axis.horizontal,
+        children: [
+          cells[2],
```
appflowy, line 114: I think hardcoding the cell index would be a problem unless there is a compile-time check.

richardshiue: Can we move away from List completely since we're no longer using Wrap/View?

appflowy: Do you mean  using Column instead of Wrap?

richardshiue: Yes, by using a couple of Wraps in a Column. I've added a commit to not use List

PR #1705, opened 2023-01-14
chore: improve number format display strings

richardshiue:


richardshiue:
Or should we localize them?

appflowy:
Wow, that's awesome!

appflowy:
> Or should we localize them?

I check out how notion handles this. It seems the same as AppFlowy.

richardshiue:
> > Or should we localize them?
> 
> I check out how notion handles this. It seems the same as AppFlowy.

Awesome, then I think this is good to go.

Issue #690, opened 2022-07-23
[FR] Save left side width setting to disk


appflowy:
The width of the left sidebar is not saved to disk, so it will get reset on the next launch. 
https://user-images.githubusercontent.com/86001920/180593248-5707c634-7f99-41cd-ae92-b71107d5b4fb.mov

1. Extend the UserSettingsService to save the width. 
2. Restore the width on the next launch.

Nike682631:
@appflowy is this issue available to work on?

annieappflowy:
I've assigned it to you. Thanks, @Nike682631 , for offering the help! If you need help, please feel free to contact @appflowy on Discord.

annieappflowy:
Hi @Nike682631 , can you update the status on this issue? Thanks

abichinger:
@annieappflowy I would like to work on this

annieappflowy:
@abichinger , assigned. thanks!

PR #1707, opened 2023-01-14
feat: save menu appearance

abichinger:
solves #690

extended `AppearanceSettingsPB` with
- `is_menu_collapsed` - the current visibility state of the menu
- `menu_offset` - the current offset of the default menu width


appflowy:
LGTM

PR #1747, opened 2023-01-27
feat: on enter exit editing property field #1295

### LINKED ISSUE #1295: [FR] Grid: Support pressing enter to exit editing the property filed
### Description

For now, we can only exit editing the text field by using the mouse. It's not high efficiency.

### Impact

It will be higher efficient to edit the property if we implement this feature.

### Additional Context

_No response_

zoli:
Introduces `onFieldSubmitted` on `RoundedInputField` type that will get used as `TextFormField` property.

Closes #1295.

### DIFF
--- a/frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
+++ b/frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
@@ -19,6 +19,7 @@ class FlowyTextField extends StatefulWidget {
   final bool autoClearWhenDone;
   final bool submitOnLeave;
   final Duration? debounceDuration;
+  final String errorText;
 
   const FlowyTextField({
     this.hintText = "",
@@ -34,6 +35,7 @@ class FlowyTextField extends StatefulWidget {
     this.autoClearWhenDone = false,
     this.submitOnLeave = false,
     this.debounceDuration,
+    this.errorText = "",
     Key? key,
   }) : super(key: key);
 
@@ -117,6 +119,7 @@ class FlowyTextFieldState extends State<FlowyTextField> {
         ),
         isDense: true,
         hintText: widget.hintText,
+        errorText: widget.errorText.isNotEmpty ? widget.errorText : null,
         hintStyle: Theme.of(context)
             .textTheme
             .bodySmall!
@@ -126,6 +129,13 @@ class FlowyTextFieldState extends State<FlowyTextField> {
         focusedBorder: OutlineInputBorder(
           borderSide: BorderSide(
             color: Theme.of(context).colorScheme.primary,
+            width: 1.0,
+          ),
+          borderRadius: Corners.s8Border,
+        ),
+        focusedErrorBorder: OutlineInputBorder(
+          borderSide: BorderSide(
+            color: Theme.of(context).colorScheme.error,
             width: 1.0,
           ),
           borderRadius: Corners.s8Border,
--- a/frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
+++ b/frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
@@ -131,14 +131,21 @@ class FlowyTextFieldState extends State<FlowyTextField> {
             color: Theme.of(context).colorScheme.primary,
             width: 1.0,
           ),
-          borderRadius: Corners.s8Border,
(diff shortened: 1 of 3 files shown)


richardshiue (commented):
From a QA perspective, LGTM! To make it even better, can you replace the `RoundedInputField` with `FlowyTextField`? We are trying to deprecate it. You would need to add the `errorText` to FlowyTextField though.

zoli:
> From a QA perspective, LGTM! To make it even better, can you replace the `RoundedInputField` with `FlowyTextField`? We are trying to deprecate it. You would need to add the `errorText` to FlowyTextField though.

Sure, I will make the changes.

zoli:
Switched from `RoundedInputField` to `FlowyTextField`. Although going through the `FlowyTextField` code I have two things to note:
1. In `FlowyTextField` normal border radius and [focused border radius](https://github.com/AppFlowy-IO/AppFlowy/blob/b34aac3f07435175a95f41c5bb300f6e2319edde/frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart#L131) are different (10 and 8). Is this intentionally?
2. Using `FlowyTextField` there is another different with `RoundedInputField` that on starting to edit the property cursor position is at the beginning of the text which before was at the end of the text.

frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
```diff
@@ -19,6 +19,7 @@ class FlowyTextField extends StatefulWidget {
   final bool autoClearWhenDone;
   final bool submitOnLeave;
   final Duration? debounceDuration;
+  final String errorText;
```
richardshiue, line 22:
```suggestion
  final String? errorText;
```

I think this will be better (the default value for this field can also be removed), and then in the TextField, you can simply do `errorText: widget.errorText,`

frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
```diff
@@ -130,6 +133,13 @@ class FlowyTextFieldState extends State<FlowyTextField> {
           ),
           borderRadius: Corners.s8Border,
```
richardshiue, line 134: Yeah I think this was a mistake, can you change it to s10?

richardshiue:
>   Using `FlowyTextField` there is another different with `RoundedInputField` that on starting to edit the property cursor position is at the beginning of the text which before was at the end of the text.

Yeah, that's true. I looked into it a bit, it's likely due to the TextEditingController used here. I don't think it's necessary to have a MultiBlocListener  in _FieldNameTextFieldState (only the first one is needed), so you can remove that and the TextEditingController, and move the child (BlocBuilder) under the remaining BlocListener. Finally, in the FlowyTextField, specify the text to be state.name

frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
zoli: Tried it, didn't work. It seems that `_FieldNameTextFieldState` is passing empty text and also maybe related to that `TextField` will check on not being `null` for `_hasError`.

zoli:
> Finally, in the FlowyTextField, specify the text to be state.name

You mean by creating a `BlocListener` in `FlowyTextField`?

frontend/app_flowy/packages/flowy_infra_ui/lib/style_widget/text_field.dart
richardshiue: Ahh I see, so there is a need for the check. Could you do `state.errorText.isEmpty ? null : state.errorText` in `_FieldNameTextFieldState`? The parameter format for the `TextField` and `FlowyTextField` should be similar

richardshiue:
> You mean by creating a `BlocListener` in `FlowyTextField`?

No, I meant in `_FieldNameTextFieldState`, specify the argument for `text` under `FlowyTextField` to be `state.name`.

zoli:
> > You mean by creating a `BlocListener` in `FlowyTextField`?
> 
> No, I meant in `_FieldNameTextFieldState`, specify the argument for `text` under `FlowyTextField` to be `state.name`.

That seems to work.

richardshiue (approved):
LGTM

PR #1687, opened 2023-01-08
feat: support group by url in kanban board

zoli:
Closes #1175

Issue #1175, opened 2022-09-27
[FR]: Support Group by the URL property in Kanban Board


appflowy:

appflowy:
Support group by URL in Kanban board.

Currently, AppFowy didn't support grouping by URL property, as shown below.

![group_by_url](https://user-images.githubusercontent.com/86001920/192419291-1b950daa-ee88-4799-bbe9-a1786fa1a2ff.gif)

zoli:
Hi,
I would like to take on this.

annieappflowy:
Thanks for claiming the issue @zoli 
Assigned it to you. Your mentor is @appflowy  for this task. You can DM him on Discord for help.

PR #1755, opened 2023-01-29
fix: empty url group bug

zoli:
Moving a row to empty url (no status) group would change the row's url to the group id which is not empty, its a random string. fixed it with puting a check to fill the url with the empty string when the group id is equal to empty url group id.

I'm not sure this is a good way to fix it though.

Continuing from https://github.com/AppFlowy-IO/AppFlowy/pull/1687 and #1175.


zoli:
Marked this as draft because I see that adding new url doesn't add the new group immediately. You  have to change the group by to other field and again go back to group by url to see the new group.

zoli:
> Marked this as draft because I see that adding new url doesn't add the new group immediately. You have to change the group by to other field and again go back to group by url to see the new group.

OK. I saw that even for group by select removing an option won't remove the group, Is this intended behavior? Also it seems generate group is called (probably from flutter?) for select option changes each time but not for url changes.

appflowy:
> > Marked this as draft because I see that adding new url doesn't add the new group immediately. You have to change the group by to other field and again go back to group by url to see the new group.
> 
> OK. I saw that even for group by select removing an option won't remove the group, Is this intended behavior? Also it seems generate group is called (probably from flutter?) for select option changes each time but not for url changes.

The behaviors of the group are different between different FieldType. I am checking it now.

zoli:
OK, It seems https://github.com/AppFlowy-IO/AppFlowy/commit/069519589e45bea8704801bf1155993d6a50ac0c has fixed the new group and remove group problem.

Still this pr fixes another bug which I mentioned at first: 
> Moving a row to empty url (no status) group would change the row's url to the group id which is not empty, its a random string. fixed it with puting a check to fill the url with the empty string when the group id is equal to empty url group id.

appflowy:
> OK, It seems [0695195](https://github.com/AppFlowy-IO/AppFlowy/commit/069519589e45bea8704801bf1155993d6a50ac0c) has fixed the new group and remove group problem.
> 
> Still this pr fixes another bug which I mentioned at first:
> 
> > Moving a row to empty url (no status) group would change the row's url to the group id which is not empty, its a random string. fixed it with puting a check to fill the url with the empty string when the group id is equal to empty url group id.

Yep. Thank you for contributing.
FYI. I merged the main branch and this PR will be merged after the CI checks.

Issue #1763, opened 2023-01-30
[Bug] Mouse unable to click a certain area


annieappflowy:
### Bug Description

I'm not able to click the area between the checked item and the unchecked item "Oh"
<img width="323" alt="image" src="https://user-images.githubusercontent.com/12026239/215461661-f51f65cf-f806-4ee9-b8af-3fd42e6c2ee4.png">

### How to Reproduce

Create a checkbox item
then press enter twice
Create another checkbox item
Click the area between these two items

### Expected Behavior

I can click and enter text in the area between two checkboxes

### Operating System

Windows 10

### AppFlowy Version(s)

0.0.9.1

PR #1771, opened 2023-01-31
fix: #1763 [Bug] Mouse unable to click a certain area

LucasXu0:


LucasXu0:
The newly added checkbox text node lacks the 'checkbox' field which makes it can't be parsed.

LucasXu0:
Close #1763

Issue #1753, opened 2023-01-29
[FR] Refactor overlay menu of node widgets


abichinger:
### Description

The menu component will be added to appflowy_editor, then every node widget can use it if they need it.

### Impact

The goal is to provide a menu with actions, which looks the same for every node widget.

### Additional Context

discussed in #1732

PR #1783, opened 2023-02-01
feat: node widget action menu

abichinger:
solves #1753


frontend/app_flowy/packages/appflowy_editor/lib/src/render/action_menu/action_menu.dart
```diff
@@ -0,0 +1,159 @@
...
+
+  bool _isHover = false;
+  bool _isPinned = false;
+
+  bool get isPinned => _isPinned;
+  bool get isHover => _isHover;
+  bool get isVisible => ActionMenuArena.instance.isVisible(path);
+
+  set isPinned(bool value) {
+    _isPinned = value;
```
LucasXu0, line 80: if (_isPinned == value) {
      return;
    }

LucasXu0, line 80: if (_isPinned == value) {
      return;
    }

frontend/app_flowy/packages/appflowy_editor/lib/src/render/action_menu/action_menu.dart
```diff
@@ -0,0 +1,159 @@
...
+    notifyListeners();
+  }
+
+  set isHover(bool value) {
+    _isHover = value;
```
LucasXu0, line 85: if (_isHover == value) {
      return;
    }

LucasXu0 (changes requested):

LucasXu0:
https://user-images.githubusercontent.com/11863087/216226406-edcc7272-a9f7-4d69-a900-5e6f22832ed3.mov

If you wrap the node widget with an action menu overlay, the overlay won't update immediately when updating the value in the node's attributes.

So, I think there will be a better solution to manage the action menu status. I will comment on this issue once I find a better solution.

abichinger:
> If you wrap the node widget with an action menu overlay, the overlay won't update immediately when updating the value in the node's attributes.

Moving the `ActionMenuOverlay` inside the `ChangeNotifierProvider` ([here](https://github.com/AppFlowy-IO/AppFlowy/blob/8ac4c21a5d0bba73da712d50ed2b18f52dea0e07/frontend/app_flowy/packages/appflowy_editor/lib/src/service/render_plugin_service.dart#L133)) should work.

frontend/app_flowy/packages/appflowy_editor/lib/src/render/action_menu/action_menu.dart
```diff
@@ -0,0 +1,180 @@
...
+        }).toList(),
+      ),
+    );
+  }
+}
+
+class ActionMenuOverlay extends StatelessWidget {
+  final Widget child;
+  final List<ActionMenuItem> items;
+  final Positioned Function(BuildContext context, List<ActionMenuItem> items)?
```
LucasXu0, line 142: How about move this customize function  to `ActionProvider`?

LucasXu0: Before
```dart
child: ActionMenuOverlay(
          items: builder.actions(context),
          customActionMenuBuilder: customActionMenuBuilder,
          child: builder.build(context),
        ),
```

After
```dart
child: ActionMenuOverlay(
          items: builder.actions(context),
          customActionMenuBuilder: builder.customActionMenuBuilder,
          child: builder.build(context),
        ),

abichinger: The intention behind `customActionMenuBuilder` is that projects using appflowy_editor can easily change all action menu appearances.

If we add `ActionProvider.customActionMenuBuilder`, then again there could be different menu designs.
That's why I believe we shouldn't add this function to `ActionProvider`. If a node widget really needs a custom menu, it can always create it itself.

PR #1791, opened 2023-02-03
fix: use window_manager to set minimum window size

### LINKED ISSUE #762: [Bug] App doesn't open fully / cant be resized when using fractional scaling on linux
**Describe the bug**
Running the app (```./app_flowy```), opens the app but it cannot be resized or maximized. The side bars are missing as well. I have added pictures for the same (screenshot 1). I managed to fix the issue by using either 100% or 200% scaling. Everytime I used 125% or 150% i would get the same issue as shown in the image. 

**To Reproduce**
1. Set fractional scaling to 125 or 150. 
2. Run the app by running ```./app_flowy```. 

**Expected behavior**
I would expect the app to look the way it does on the website with the side bars and the full page shown. I've added an image to show the expected behavior on my machine (screenshot 2). 

**Screenshots**
screenshot 1:
image
screenshot 2:
image

**Desktop**
 - OS: Pop!_OS 22.04 LTS
 - Browser: Chrome
 - Graphics: radeon graphics
 - Hardware model: Lenovo Legion 5 Pro 16ACH6H
 - Gnome version: 42.2

**Additional context**
 - Screen resolution: 2560 x 1600 (16:10)
 - Refresh rate: 60Hz

### LINKED ISSUE #1559: [Bug] Scaling error 4k monitor Linux
### Bug Description

scaling error when switching to more than 100% on 4K monitor

### How to Reproduce

1. open settings

2. Browse display settings

3. change scale to more than 100%

### Expected Behavior

I expected the user interface to adapt to the larger scale, as well as other software

### Operating System

Pop OS, Linux Mint Xfce, Manjaro Genome

### AppFlowy Version(s)

0.0.8.1 and previous

### Screenshots

Screenshot from 2022-12-09 13-38-28

Screenshot from 2022-12-09 13-38-48

### Additional Context

I'm happy for the software, congratulations on everyone's work.

Merrit:
Resolves #762
Resolves #1559

Fixes being unable to resize the tiny default window on Linux.

Having `window_manager` available also gives us the ability to do [many window-related operations](https://pub.dev/packages/window_manager#api) in the future if desired.

If this is merged I'd next suggest we look at the ability to remember window size & position between launches.

### DIFF
--- a/frontend/app_flowy/lib/startup/tasks/app_widget.dart
+++ b/frontend/app_flowy/lib/startup/tasks/app_widget.dart
@@ -73,7 +73,7 @@ class ApplicationWidget extends StatelessWidget {
   Widget build(BuildContext context) {
     const ratio = 1.73;
     const minWidth = 600.0;
-    windowManager.setSize(const Size(minWidth, minWidth / ratio));
+    windowManager.setMinimumSize(const Size(minWidth, minWidth / ratio));
 
     final cubit = AppearanceSettingsCubit(appearanceSetting)
       ..readLocaleWhenAppLaunch(context);
--- a/frontend/app_flowy/lib/startup/tasks/app_widget.dart
+++ b/frontend/app_flowy/lib/startup/tasks/app_widget.dart
@@ -5,7 +5,6 @@ import 'package:appflowy_backend/log.dart';
 import 'package:appflowy_backend/protobuf/flowy-user/protobuf.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
-import 'package:window_manager/window_manager.dart';
 
 import '../../user/application/user_settings_service.dart';
 import '../../workspace/application/appearance.dart';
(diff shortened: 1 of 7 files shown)


frontend/app_flowy/lib/startup/tasks/app_widget.dart
```diff
@@ -73,7 +73,7 @@ class ApplicationWidget extends StatelessWidget {
   Widget build(BuildContext context) {
     const ratio = 1.73;
     const minWidth = 600.0;
-    setWindowMinSize(const Size(minWidth, minWidth / ratio));
+    windowManager.setSize(const Size(minWidth, minWidth / ratio));
```
LucasXu0, line 76: The previous one is `setWindowMinSize` and now is `setSize`. We should keep the same logic as before.

Merrit: Previously (and now) we are just calling a library directly from a widget, so the choice of name is because of their implementation.

My suggestion would be to have a service that abstracts the implementation away. Then from outside the service we can call things predictably, like:

_Initialize on app startup_

```dart
AppWindow.instance.initialize();
```

_Show / hide to tray_

```dart
AppWindow.instance.hide();
```

```dart
AppWindow.instance.show();
```

etc.

Merrit: My bad, I misread; fixed that now. :+1:

LucasXu0:
`windowManager.setMinimumSize(const Size(minWidth, minWidth / ratio));` will cause a crash on macOS.

LucasXu0:
I have addressed the problem. The width and height should be integers in this case.

Issue #1762, opened 2023-01-30
[FR] Support shortcuts for checking on/off a checkbox


annieappflowy:
### Description

As described in the title, I would like to use keyboard shortcuts to check off a checkbox and uncheck it.

<img width="235" alt="image" src="https://user-images.githubusercontent.com/12026239/215460912-29d13128-7554-420e-9044-844f70227079.png">

### Impact

Users who want to quickly check off or uncheck a checkbox

MayurSMahajan:
Working on it!

PR #1817, opened 2023-02-07
[FR] Shortcut for toggling checkbox 

MayurSMahajan:
Solves #1762

Users can now use the key combination: **Ctrl+Enter** on Windows and Linux to quickly on/off a checkbox in AppFlowy. Mac Users have to press **Meta+Enter** in order to achieve the same.

This also works with multiple checkboxes. Thus as a user, you can select multiple todos and then press Ctrl+Enter to toggle them.

### DIFF
--- a/frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/checkbox_event_handler.dart
+++ b/frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/checkbox_event_handler.dart
@@ -1,22 +1,49 @@
 import 'package:appflowy_editor/src/service/shortcut_event/shortcut_event_handler.dart';
 import 'package:appflowy_editor/src/core/document/attributes.dart';
 import 'package:appflowy_editor/src/core/document/node.dart';
+import 'package:appflowy_editor/appflowy_editor.dart';
 import 'package:flutter/material.dart';
 
 ShortcutEventHandler toggleCheckbox = (editorState, event) {
   final selection = editorState.service.selectionService.currentSelection.value;
   final nodes = editorState.service.selectionService.currentSelectedNodes;
-  final textNodes = nodes.whereType<TextNode>().toList(growable: false);
-  if (selection == null || textNodes.isEmpty) {
+  final checkboxTextNodes = nodes
+      .where(
+        (element) =>
+            element is TextNode &&
+            element.subtype == BuiltInAttributeKey.checkbox,
+      )
+      .toList(growable: false);
+
+  if (selection == null || checkboxTextNodes.isEmpty) {
     return KeyEventResult.ignored;
   }
-  for (Node node in textNodes) {
-    if (node.attributes.containsKey('checkbox') &&
-        node.attributes.containsValue('checkbox')) {
-      bool currentStatus = !(node.attributes['checkbox']);
-      Attributes checkboxAttribute = {'checkbox': currentStatus};
-      node.updateAttributes(checkboxAttribute);
+
+  //If any one of the checkboxes is unchecked then make all checkboxes checked
+
+  bool isAllCheckboxesChecked = true;
+  final transaction = editorState.transaction;
+  for (final node in checkboxTextNodes) {
+    if (node.attributes[BuiltInAttributeKey.checkbox] == false) {
+      isAllCheckboxesChecked = false;
+      transaction.updateNode(node, {
+        BuiltInAttributeKey.checkbox:
+            !node.attributes[BuiltInAttributeKey.checkbox]
+      });
     }
   }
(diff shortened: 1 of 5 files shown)


frontend/app_flowy/packages/appflowy_editor/lib/src/service/shortcut_event/built_in_shortcut_events.dart
```diff
@@ -159,6 +160,13 @@ List<ShortcutEvent> builtInShortcutEvents = [
     linuxCommand: 'ctrl+u',
     handler: formatUnderlineEventHandler,
   ),
+  ShortcutEvent(
+    key: 'Toggle Checkbox',
```
LucasXu0, line 164: `CTRL + Q`? Normally, `CTRL + Q` means quitting the app. 

I suggest using `CTRL + ENTER` or `META + ENTER` instead.

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/checkbox_event_handler.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:appflowy_editor/src/service/shortcut_event/shortcut_event_handler.dart';
+import 'package:appflowy_editor/src/core/document/attributes.dart';
+import 'package:appflowy_editor/src/core/document/node.dart';
+import 'package:flutter/material.dart';
+
+ShortcutEventHandler toggleCheckbox = (editorState, event) {
+  final selection = editorState.service.selectionService.currentSelection.value;
+  final nodes = editorState.service.selectionService.currentSelectedNodes;
+  final textNodes = nodes.whereType<TextNode>().toList(growable: false);
```
LucasXu0, line 9:
```dart
final checkboxTextNodes = nodes
      .where(
        (element) =>
            element is TextNode &&
            element.subtype == BuiltInAttributeKey.checkbox,
      )
      .toList(growable: false);
```

I would like to use the above code to get the checkbox text nodes.

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/checkbox_event_handler.dart
```diff
@@ -0,0 +1,22 @@
...
+  if (selection == null || textNodes.isEmpty) {
+    return KeyEventResult.ignored;
+  }
+  for (Node node in textNodes) {
```
LucasXu0, line 13: It's not allowed to use `updateAttributes` function directly.

You should use transactions to update the document.

For example
```dart
final transaction = editorState.transaction;
  for (final node in checkboxTextNodes) {
    transaction.updateNode(node, {
      BuiltInAttributeKey.checkbox:
          !node.attributes[BuiltInAttributeKey.checkbox]
    });
  }
  editorState.apply(transaction);
```

LucasXu0:
Hey, @MayurSMahajan. For the logic of the handler, I think this would be better.

1. If any one of the checkboxes is unchecked, then make all of the checkboxes checked after pressing the shortcut.
2. Only if all the checkboxes are checked, then make all of the checkboxes unchecked after pressing the shortcut.

frontend/app_flowy/packages/appflowy_editor/lib/src/service/internal_key_event_handlers/checkbox_event_handler.dart
```diff
@@ -0,0 +1,46 @@
...
+      )
+      .toList(growable: false);
+
+  if (selection == null || checkboxTextNodes.isEmpty) {
+    return KeyEventResult.ignored;
+  }
+
+  //If any one of the checkboxes is unchecked then make all checkboxes checked
+
+  bool isAllCheckboxesChecked = true;
```
LucasXu0, line 21:
```dart
  bool isAllChcked = checkboxTextNodes
      .every((e) => node.attributes[BuiltInAttributeKey.checkbox] == true);
  final transaction = editorState.transaction;
  if (isAllChcked) {
    // if all the checkboxes are checked, then make all of the checkboxes unchecked

    // apply the transaction to make all checkboxes unchecked.
    // ...
  } else {
    // If any one of the checkboxes is unchecked then make all checkboxes checked

    // apply the transaction to make all checkboxes checked.
    // ...
  }
```


No need to use `!` to get the opposite value. Keep all checkboxes as the same status, all the check statuses of them are both true, or just both false.

MayurSMahajan: makes sense

LucasXu0:
LGTM. I will merge it after releasing v0.1.0.

PR #1819, opened 2023-02-07
chore: remove appflowy-board

appflowy:


appflowy:
Extract appflowy-board to https://github.com/AppFlowy-IO/appflowy-board

PR #1831, opened 2023-02-09
Feat/appflowy tauri

mikiastilahun:


frontend/appflowy_tauri/src/appflowy_app/redux/grid/slice.ts
```diff
@@ -0,0 +1,151 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { nanoid } from 'nanoid';
+
+// eslint-disable-next-line no-shadow
+export enum FieldType {
```
appflowy, line 5: You can use the FieldType defined in services/backend/classes/flowy-database/field_entities.ts. I already export them in services/backend/index.ts.

mikiastilahun: @appflowy I've changed the FieldType to be imported from the backend generated enum.

PR #1842, opened 2023-02-10
Refactor app window logic

Merrit:
This is much cleaner and better encapsulated. Sets up an area for further window management in the future.

Does not try to initialize on mobile or web, which would crash.


appflowy:
@Merrit Thank you for contributing. It's better. 
LGTM

Issue #1684, opened 2023-01-08
[FR] Tab for lists


annieappflowy:
### Discussed in https://github.com/AppFlowy-IO/AppFlowy/discussions/1645

<div type='discussions-op-text'>

<sup>Originally posted by **Docfips** January  4, 2023</sup>
Good afternoon,

I am curious if there was anyway to be able to add the "tab" feature to subtask when using the To-Do?

- [ ]  1 here
     - [ ]  2 here</div>

MayurSMahajan:
Hey Annie! currently, the tab feature is not supported. I could implement it, but I had one doubt. 
##### Current Behaviour:
pressing the tab in a checkbox causes the text field to create a tabbed space
##### Expected Behaviour:
pressing the tab in a checkbox causes the checkbox to shift right, denoting sub checkbox or todo

My question is, should we add a **combination** of keys to support this, like pressing __tab and LShift will create a subtask__ and pressing only **Tab will work like before** OR should we override the Tab key and always create a subtask instead of the original behavior.

annieappflowy:
Hi @MayurSMahajan ,

By default, pressing Tab should indent the checkbox like you described. This will incur overheads for users who want to create a tabbed space but won't prevent them using Space to accomplish the same goal. 

We should enable users to customize their shortcuts as a complete solution.

MayurSMahajan:
I absolutely agree!

PR #1847, opened 2023-02-12
Pressing tab on a todo creates a sub-todo

MayurSMahajan:
Solves #1684 

**Impact:**
Users can now press the tab key to create a sub todo or a sub checkbox.

Pressing tab here:
image

will result in:
image


LucasXu0:
LGTM.

PR #1852, opened 2023-02-13
feat: add open ai key to database

LucasXu0:


frontend/app_flowy/assets/translations/en.json
```diff
@@ -191,6 +191,12 @@
       "create": "Create",
       "folderPath": "Path to store your folder",
       "locationCannotBeEmpty": "Path cannot be empty"
+    },
+    "user": {
+      "name": "Name",
+      "icon": "Icon",
+      "selectAnIcon": "Select an Icon",
```
appflowy, line 198: I think Select an icon would be better

frontend/app_flowy/assets/translations/en.json
```diff
@@ -191,6 +191,12 @@
...
+      "pleaseInputYourOpenaiKey": "please input your openai key"
```
appflowy, line 199: OpenAI

shared-lib/user-model/src/parser/user_openai_key.rs
```diff
@@ -0,0 +1,16 @@
+use crate::errors::UserErrorCode;
+
+#[derive(Debug)]
+pub struct UserOpenaiKey(pub String);
+
+impl UserOpenaiKey {
+  pub fn parse(s: String) -> Result<UserOpenaiKey, UserErrorCode> {
+    Ok(Self(s))
```
appflowy, line 8: Should not be empty string

LucasXu0: The OpenAI key can be an empty string.

LucasXu0:
On the dart side, I rename the `openai` to `openAI`. On the rust side, I  rename the `openai` to `openAI` if the code is camel-case style, and keep `openai` if the code is underscore case style.

shared-lib/user-model/src/parser/user_openai_key.rs
appflowy: Wow, interesting

PR #1853, opened 2023-02-13
Catch platform errors

Merrit:
Adds handling for platform errors that are not caught by the
Flutter framework. Doing so will log the error properly and
prevent the app from crashing.


appflowy:
Nice! LGTM

PR #1854, opened 2023-02-13
fix(network_monitor): add missing case for `other` network type

Merrit:
Missing case causing static analysis error that is blocking the build.


Merrit:
That's weird - can't build locally because the switch statement doesn't handle the "other" case.

Also the "other" case doesn't exist during CI build..? :thinking:

appflowy:
> That's weird - can't build locally because the switch statement doesn't handle the "other" case.
> 
> Also the "other" case doesn't exist during CI build..? 🤔

The version of the connectivity_plus might be different.

Merrit:
> The version of the connectivity_plus might be different.

Looks like you are right. Thanks!

appflowy:
LGTM

PR #1855, opened 2023-02-13
Add generated Flutter files to gitignore

Merrit:
Flutter regenerates these files frequently, including when there are minor package upgrades, and they do not need to be tracked in version control where it only creates unnecessary noise.


LucasXu0:
Hey, @Merrit. Thanks for contributing. Would it be better to add the below code to `.gitignore`

https://github.com/flutter/flutter/blob/master/.gitignore
```
# Flutter/Dart/Pub related
**/doc/api/
.dart_tool/
.flutter-plugins
.flutter-plugins-dependencies
**/generated_plugin_registrant.dart
.packages
.pub-cache/
.pub/
build/
flutter_*.png
linked_*.ds
unlinked.ds
unlinked_spec.ds

# Android related
**/android/**/gradle-wrapper.jar
.gradle/
**/android/captures/
**/android/gradlew
**/android/gradlew.bat
**/android/local.properties
**/android/**/GeneratedPluginRegistrant.java
**/android/key.properties
*.jks

# iOS/XCode related
**/ios/**/*.mode1v3
**/ios/**/*.mode2v3
**/ios/**/*.moved-aside
**/ios/**/*.pbxuser
**/ios/**/*.perspectivev3
**/ios/**/*sync/
**/ios/**/.sconsign.dblite
**/ios/**/.tags*
**/ios/**/.vagrant/
**/ios/**/DerivedData/
**/ios/**/Icon?
**/ios/**/Pods/
**/ios/**/.symlinks/
**/ios/**/profile
**/ios/**/xcuserdata
**/ios/.generated/
**/ios/Flutter/.last_build_id
**/ios/Flutter/App.framework
**/ios/Flutter/Flutter.framework
**/ios/Flutter/Flutter.podspec
**/ios/Flutter/Generated.xcconfig
**/ios/Flutter/ephemeral
**/ios/Flutter/app.flx
**/ios/Flutter/app.zip
**/ios/Flutter/flutter_assets/
**/ios/Flutter/flutter_export_environment.sh
**/ios/ServiceDefinitions.json
**/ios/Runner/GeneratedPluginRegistrant.*

# macOS
**/Flutter/ephemeral/
**/Pods/
**/macos/Flutter/GeneratedPluginRegistrant.swift
**/macos/Flutter/ephemeral
**/xcuserdata/

# Windows
**/windows/flutter/generated_plugin_registrant.cc
**/windows/flutter/generated_plugin_registrant.h
**/windows/flutter/generated_plugins.cmake

# Linux
**/linux/flutter/generated_plugin_registrant.cc
**/linux/flutter/generated_plugin_registrant.h
**/linux/flutter/generated_plugins.cmake

# Coverage
coverage/

# Symbols
app.*.symbols

# Exceptions to above rules.
!**/ios/**/default.mode1v3
!**/ios/**/default.mode2v3
!**/ios/**/default.pbxuser
!**/ios/**/default.perspectivev3
!/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
!/dev/ci/**/Gemfile.lock
```

Merrit:
Hey @LucasXu0.

> Otherwise, if we use generated as the named filename, they will be ignored.

I was looking at their gitignore as well, and I decided to go with the glob pattern instead for two reasons:

1. The Flutter version will almost certainly change. If they add a new generated file, we won't know about it until it starts showing up in random PRs.
2. If we use "generated" in a filename, I would presume we would want to treat it as such files usually are; that is to say not version controlled and regenerated on build. If for some reason though we had such a file we *did* want tracked we could whitelist that specific file with a negate pattern in the gitignore, for example:

```
!flowy_generated_interface.c
```

or even:

```
!flowy_generated_*
```

Which seems more manageable to me than potential whack-a-mole with the Flutter framework and potential 3rd party sources.

Curious to hear your thoughts :)

LucasXu0:
It sounds reasonable. LGTM. I will merge it after the CI pass.

PR #1873, opened 2023-02-20
fix: error happens when parsing readme content

LucasXu0:
Screenshot 2023-02-20 at 17 48 04


LucasXu0:
The `create_view` function accepts raw document content instead of document transactions.

PR #1879, opened 2023-02-21
chore: nav items persist, board layout

Ascarbek:
This PR includes:
- load current workspace items
- if there's no workspace, then create new one
- persist to backend folder(app) create, rename and delete
- persist to backend new document, grid and board. persist rename
- folder open/close animation
- database store and reducers for future use in board and grid pages
- board(kanban) page layout (readonly) based on sample data


frontend/appflowy_tauri/src/appflowy_app/components/auth/auth.hooks.ts
```diff
@@ -1,14 +1,30 @@
 import { currentUserActions } from '../../stores/reducers/current-user/slice';
 import { useAppDispatch, useAppSelector } from '../../stores/store';
+import { SignInPayloadPB, UserEventSignIn, UserProfilePB } from '../../../services/backend/events/flowy-user';
 
 export const useAuth = () => {
   const dispatch = useAppDispatch();
 
   const currentUser = useAppSelector((state) => state.currentUser);
 
+  async function login(email: string, password: string, name: string): Promise<UserProfilePB> {
```
appflowy, line 10: Would it be better if we use the Promise<Result<x,x>>

frontend/appflowy_tauri/src/appflowy_app/components/board/Board.hooks.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ICellData, IDatabase, IDatabaseRow, ISelectOption } from '../../stores/reducers/database/slice';
+
+export const useBoard = () => {
+  const dispatch = useAppDispatch();
+  const groupingFieldId = useAppSelector((state) => state.board);
+  const database = useAppSelector((state) => state.database);
+  const [title, setTitle] = useState('');
+  const [boardColumns, setBoardColumns] =
+    useState<(ISelectOption & { rows: (IDatabaseRow & { isGhost: boolean })[] })[]>();
+  const [movingRowId, setMovingRowId] = useState<string | undefined>(undefined);
```
appflowy, line 13: I think using the Option<String> instead of `string | undefined` would be better.  The option concept comes from Rust and it was introduced in the ts-result package.

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
```diff
@@ -30,58 +56,130 @@ export const useFolderEvents = (folder: IFolder) => {
...
 
-  const deleteFolder = () => {
+  const deleteFolder = async () => {
     closePopup();
+    await FolderEventDeleteApp(
+      AppIdPB.fromObject({
+        value: folder.id,
+      })
+    );
     appDispatch(foldersActions.deleteFolder({ id: folder.id }));
```
appflowy, line 81: I think the backend service(for example, AppBackendService) should go to the foldersActions' handler. Just like the `Bloc`(Flutter) does.

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
```diff
@@ -30,58 +56,130 @@ export const useFolderEvents = (folder: IFolder) => {
...
-  const onAddNewBoardPage = () => {
+  const onAddNewBoardPage = async () => {
     closePopup();
-    appDispatch(
-      pagesActions.addPage({
-        folderId: folder.id,
-        pageType: ViewLayoutTypePB.Board,
-        title: 'New Board 1',
-        id: nanoid(6),
+    const createViewResult = await FolderEventCreateView(
```
appflowy, line 136: Is the call chain like this?
appDispatch -> action -> handler (interact with the backend using AppBackendService) .

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/NewFolderButton.hooks.ts
```diff
@@ -1,12 +1,28 @@
...
+import { CreateAppPayloadPB, FolderEventCreateApp } from '../../../../services/backend/events/flowy-folder';
+import { workspaceSlice } from '../../../stores/reducers/workspace/slice';
 
 export const useNewFolder = () => {
   const appDispatch = useAppDispatch();
+  const workspace = useAppSelector((state) => state.workspace);
 
-  const onNewFolder = () => {
-    appDispatch(foldersActions.addFolder({ id: nanoid(8), title: 'New Folder 1' }));
+  const onNewFolder = async () => {
```
appflowy, line 11: I think OnNewFolder should be moved to the addFolder handler. Then the UI will wait for the response comes from the backend and handler the errors if there are any.

image.

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/PageItem.hooks.ts
```diff
@@ -17,12 +25,23 @@ export const usePageEvents = (page: IPage) => {
     closePopup();
   };
 
-  const changePageTitle = (newTitle: string) => {
+  const changePageTitle = async (newTitle: string) => {
+    await FolderEventUpdateView(
```
appflowy, line 29: Please use `ViewBackendService` and move the logic to the renamePage handler.

frontend/appflowy_tauri/src/appflowy_app/components/auth/auth.hooks.ts
Ascarbek: refactored to use backend service

frontend/appflowy_tauri/src/appflowy_app/components/board/Board.hooks.ts
Ascarbek: in this particular case it's normal to use `string | undefined` because this variable(state) is used in UI. In other cases I will use Option and Result types

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
Ascarbek: refactored to use backend service

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
Ascarbek: refactored to use backend service

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/NewFolderButton.hooks.ts
Ascarbek: refactored to use backend service.
To complete error handling we need to make a window that can show error messages. That will be a different task that I will create soon.
Right here addFolder is in correct place because it is called only on success.

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
```diff
@@ -30,57 +47,81 @@ export const useFolderEvents = (folder: IFolder) => {
...
 
   const closePopup = () => {
     setShowFolderOptions(false);
     setShowNewPageOptions(false);
   };
 
-  const onAddNewDocumentPage = () => {
+  const onAddNewDocumentPage = async () => {
     closePopup();
+    const newView = await appBackendService.createView({
```
appflowy, line 81: Are there any global error page? If we have, then the errors can be shown there. Every call to the backend service, it comes with a result. It might success or failed. I think we should display the errors first.

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
```diff
@@ -30,57 +47,81 @@ export const useFolderEvents = (folder: IFolder) => {
...
+      name: 'New Document 1',
```
appflowy, line 82: How can we support i18 by using the static string?

appflowy: AppFlowy uses https://pub.dev/packages/easy_localization in flutter. Can we reuse the translation files?
image

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/PageItem.hooks.ts
Ascarbek: I've already updated this file. You should "select changes from all commits"

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
Ascarbek: I've created Error Handler page Error modal and corresponding redux functions to show error messages. Now any error should be properly handled.

frontend/appflowy_tauri/src/appflowy_app/components/layout/NavigationPanel/FolderItem.hooks.ts
Ascarbek: I've installed react-i18next and now it is using exact same translation files. So we should be good.

PR #1901, opened 2023-02-28
chore: include time per cell

### LINKED ISSUE #695: [FR] Improve the grid date unit test in the backend
The existing tests for the grid date field are in [here](https://github.com/AppFlowy-IO/AppFlowy/blob/main/frontend/rust-lib/flowy-grid/src/services/field/type_options/date_type_option/date_tests.rs) out.

Expected:
Add more unit tests to cover more corner cases.

### LINKED ISSUE #1888: [FR] Move include time setting for date fields from type-option to cell
### Description

Title says all. This allows a higher level of flexibility for users since cells without time time will still show an unwanted, arbitrary time when the bug outlined in #1800 is fixed.

### Impact

Users of the date option type in database

### Additional Context

_No response_

richardshiue:
Change the option of including a time for date cells in the appflowy database from an option that applies to the entire field to one that you can toggle for each cell. Toggling the include time option on the header will toggle the setting for all the individual cells in the database.

todo

- [x] fix test
- [x] connect to frontend
- [ ] header option
- [ ] more tests (?)

resolves #1888, resolves #695

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Passes existing tests

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules


frontend/rust-lib/flowy-database/src/services/field/type_options/date_type_option/date_type_option_entities.rs
```diff
@@ -71,36 +78,27 @@ impl ToCellChangesetString for DateCellChangeset {
...
 }
 
 impl FromCellString for DateCellData {
   fn from_cell_str(s: &str) -> FlowyResult<Self>
   where
     Self: Sized,
   {
-    let num = s.parse::<i64>().ok();
-    Ok(DateCellData(num))
+    let result: DateCellData = serde_json::from_str(s).unwrap_or_default();
```
appflowy, line 92: It will break the changes. Because the data in disk is just a timestamp. So It can't be deserialized using the default deserializer provided by serde. We need to write our version deserializer.

frontend/rust-lib/flowy-database/src/services/field/type_options/date_type_option/date_type_option.rs
```diff
@@ -23,9 +23,6 @@ pub struct DateTypeOptionPB {
 
   #[pb(index = 2)]
   pub time_format: TimeFormat,
-
-  #[pb(index = 3)]
-  pub include_time: bool,
```
appflowy, line 28: This property can't be removed even though the property we don't use it anymore. Because the protobuf will use these properties declaration to parser the data. 

If the protobuf struct doesn't need to parser from the disk data, it's ok to remove its properties. For example, the protobuf passthrough the Rust and Dart via the FFI.

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/date_cell/date_editor.dart
```diff
@@ -111,12 +111,14 @@ class _CellCalendarWidgetState extends State<_CellCalendarWidget> {
       child: BlocBuilder<DateCellCalendarBloc, DateCellCalendarState>(
         buildWhen: (p, c) => p != c,
         builder: (context, state) {
+          bool includeTime = state.dateCellData
```
appflowy, line 114: Style suggestion:
place a comma after dateData.includeTime

richardshiue: > place a comma after dateData.includeTime

Sure thing

> Would it be better if creating a property in the state called includeTime.

Should we consider just unpacking the entire `DateCellData` into the state and remove `dateCellData`? `time` is already there as well.

image

PR #1915, opened 2023-03-03
using controllers in react hooks WIP

Ascarbek:


frontend/appflowy_tauri/src/appflowy_app/components/_shared/database-hooks/useCell.ts
```diff
@@ -0,0 +1,40 @@
...
+          const v = value.unwrap() as DateCellDataPB;
+          setData([v.date]);
+        } else {
+          const v = value.unwrap() as string;
+          setData([v]);
+        }
+      },
+    });
+
+    cellController.getCellData();
```
appflowy, line 33: image
Do you see this warning in your IDE?

frontend/appflowy_tauri/src/appflowy_app/components/_shared/database-hooks/useCell.ts
```diff
@@ -0,0 +1,40 @@
...
+import { DateCellDataPB, FieldType, SelectOptionCellDataPB } from '../../../../services/backend';
+import { useState } from 'react';
+
+export const useCell = (cellIdentifier: CellIdentifier, cellCache: CellCache, fieldController: FieldController) => {
+  const [data, setData] = useState<string[]>([]);
+
+  const loadCell = async () => {
+    const builder = new CellControllerBuilder(cellIdentifier, cellCache, fieldController);
+    const cellController = builder.build();
+    cellController.subscribeChanged({
```
appflowy, line 14: I think this cell controller should be used in the corresponding field-type cell components. Each component uses useEffect to subscribe to the change and returns an unsubscribe action to the effect when the current component gets unmounted.

frontend/appflowy_tauri/src/appflowy_app/components/_shared/database-hooks/loadField.ts
```diff
@@ -0,0 +1,99 @@
...
+  makeDateTypeOptionContext,
+  makeMultiSelectTypeOptionContext,
+  makeNumberTypeOptionContext,
+  makeSingleSelectTypeOptionContext,
+} from '../../../stores/effects/database/field/type_option/type_option_context';
+import { boardActions } from '../../../stores/reducers/board/slice';
+import { FieldInfo } from '../../../stores/effects/database/field/field_controller';
+import { AppDispatch } from '../../../stores/store';
+
+export default async function (viewId: string, fieldInfo: FieldInfo, dispatch?: AppDispatch): Promise<IDatabaseField> {
```
appflowy, line 24: The IDatabaseField contains IFieldOptions which is a collection of all field-type options(This brings too much cognitive load). So it must use the current field type to determine which option can be used. Would it be better to separate concerns from the IFieldOptions?

From my point of view, I will create multiple cell components. Each component uses its own type option.

frontend/appflowy_tauri/src/appflowy_app/components/_shared/database-hooks/useCell.ts
Ascarbek: fixed

frontend/appflowy_tauri/src/appflowy_app/components/_shared/database-hooks/loadField.ts
Ascarbek: I agree, I've made changes accordingly

frontend/appflowy_tauri/src/appflowy_app/components/_shared/database-hooks/useCell.ts
Ascarbek: Agreed. I've created component for each type and moved useCell inside of them

Issue #1912, opened 2023-03-03
[Bug] Dev Scripts cause flutter version reset


AmanNegi:
## Issue

When we setup our environment using the `install_dev_env` files, it causes the Flutter Version to reset to the latest stable one. This behavior should be improved as we are using the specific version `v3.3.10` and any other version would cause error in the building process. 

## Caused By
I was able to narrow down the cause of error as the following lines in the `install_dev_env` files:
``` sh
# Enable the flutter stable channel
printMessage "Setting up Flutter"
flutter channel stable
```

## Solutions
>  Simply, comment these lines and prompt the user to have the specific version at end. As anyone following the [gitbook](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/environment-setup) would have the right version installed. Also we are assuming here that the, `flutter` is added to path and is thus installed, so it makes sense to remove this line.

### How to Reproduce

On your system install `flutter v3.3.10`, and then run the scripts provided under `install_dev_env` and then check your flutter version.

### Expected Behavior

Do not modify pre-existing flutter version.

### Operating System

Ubuntu 20.04.5 LTS

### AppFlowy Version(s)

1.0.0.1

AmanNegi:
Let me know, if I am right on this. And I would love to comment those lines 🤣

LucasXu0:
Hey, @AmanNegi. Yes. It's right. You can submit a PR to replace them with a new command to switch to v3.3.10 version.

AmanNegi:
Hey @LucasXu0 as there are no further reports of this issue, I will be closing this issue.

PR #1923, opened 2023-03-04
fix(appflowy_flutter): Flutter Version Reset Error

AmanNegi:
- `flutter channel stable` was causing the FlutterSDK to upgrade to its latest version.
- Added code to fix this behavior.
- Tested on Windows and Linux.

Issue: #1912


LucasXu0:
Hi, @AmanNegi.

The option `-P` is invalid on macOS.

```bash
AppFlowy : Setting up Flutter
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
Unhandled exception:
```

AmanNegi:
Hey @LucasXu0, I have fixed the issue. Could you please test it now, I don't have any `MacOS` device so I can't test sorry :cry:.

LucasXu0:
Okay. Let me try it.

LucasXu0:
Hey, @AmanNegi. The script will complain as the below pic shows if the developer uses the other version of Flutter. Because the script switches to the Flutter path but doesn't switch back.

Screenshot 2023-03-08 at 14 08 38

AmanNegi:
Yep, thanks for pointing that out! I will fix that behavior.

AmanNegi:
@LucasXu0 Could you please check if the changes I made, fix the issue that we were facing?

LucasXu0:
Hi, @AmanNegi. LGTM. And I change the text format from CTLF to LF.

PR #1961, opened 2023-03-11
chore(typos): fix typos

stavares843:
Fix typos across codebase.


a-wallen:
Since this patch only contains typo fixes, I think we can merge even if codecov/patch is failing, no? @LucasXu0 @appflowy

PR #1980, opened 2023-03-12
Tauri UI issues

Ascarbek:


frontend/rust-lib/flowy-database/src/services/database_view/editor_manager.rs
```diff
@@ -66,6 +66,7 @@ impl DatabaseViews {
   pub async fn close(&self, view_id: &str) {
     if let Some(view_editor) = self.view_editors.write().await.remove(view_id) {
       view_editor.close().await;
+
```
appflowy, line 69: Shouldn't commit empty line

Issue #1671, opened 2023-01-06
[FR] Improve darkmode


annieappflowy:
### Description

This is a parent issue for making improvements on the dark mode theme. 
Related issues are listed below:
- https://github.com/AppFlowy-IO/AppFlowy/issues/433
- https://github.com/AppFlowy-IO/AppFlowy/issues/399
- https://github.com/AppFlowy-IO/AppFlowy/issues/336
- https://github.com/AppFlowy-IO/AppFlowy/issues/1807
- https://github.com/AppFlowy-IO/AppFlowy/issues/875

### Impact

Darkmode users

hyj1204:
I'd love to help with it.

PR #1987, opened 2023-03-13
[WIP]Feat(appflowy_flutter): dark mode/theme improvement(side bar theme improvement)

hyj1204:
This PR is the part of Dark mode improvement #1671 
3/14 update:
I plan to
1. Create `AFSideBarThemeExtension` to handler all the cosmetic attribute within the side bar area(the right section of the app). It has two modes to apply depends current user theme mode. 

<p float="left">
image

image
</p>

2. Use the theme extension data in all widgets'  cosmetic attributes
image

3. (TODO) To use the colors in `FlowyColorScheme` in the `AFSideBarThemeExtension`. In this way, all the colors will come form its own theme under current light/dark mode. 
4. (TODO) Organize all the theme extension in different parts of the app.


hyj1204:
This solution is not being utilized in the current project. 
Achieve this PR for future reference

PR #2031, opened 2023-03-20
feat: add icon to dropdown button in language view

Sudhanva-Nadiger:
- drop down button gives the default icon
- set auto focus true in dropdown button
- add padding to dropdown button
- change edgeIsets.only(left,right) to edgeInsets.symmetric(horizontal)


LucasXu0:
LGTM.

Sudhanva-Nadiger:
Thanks @LucasXu0 !

PR #2065, opened 2023-03-21
Feat: sync name on change

### LINKED ISSUE #2033: [FR] Auto save the change to the workspace name
### Description

image
image

Auto-save the change without needing to press Enter
For example, after changing the name from Me to Annie, I just close up the Settings by clicking on the gray area outside the Settings window. The name should be changed to Annie in this case

### Impact

Users who change their workspace name

### Additional Context

_No response_

DestinedCodes:
### Description:
[Screencast from 21-03-2023 09:15:59.webm](https://user-images.githubusercontent.com/84413505/226550394-67c0316f-8362-445d-b9e2-a7710c936ad1.webm)

This pull request fixes issue #2033, which was related to the auto-save functionality for changing the workspace name in the Settings section. The current behavior required the user to press Enter in order to save the changes. However, with this fix, the change will be automatically saved without the need to press Enter. For example, if the user changes the workspace name from "Me" to "Destiny", closing the Settings by clicking on the gray area outside the Settings window will now result in the name being changed to "Destiny".

### Types of Changes Made:

Implemented auto-save functionality for the workspace name in the Settings section. (used onChanged in place of onSubmitted)
Refactored the code to use a StatefulWidget instead of a StatelessWidget, in order to manage the lifecycle of the TextEditingController properly.
Tested the changes to ensure that the auto-save feature works as expected.
Added necessary comments to explain the changes made in the code and make it more readable.

Fixes #2033.

### DIFF
--- a/frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_user_view.dart
+++ b/frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_user_view.dart
@@ -11,6 +11,7 @@ import 'package:appflowy/generated/locale_keys.g.dart';
 import 'package:easy_localization/easy_localization.dart';
 
 import 'dart:convert';
+import 'dart:async';
 
 const defaultUserAvatar = '1F600';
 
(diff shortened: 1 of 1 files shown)


DestinedCodes:
At first, I used the onChanged method instead of onSubmitted, but during testing, I noticed that the cursor kept returning to the first line. After some investigation, I discovered that the root of the problem was a new instance of TextEditingController being created each time the username was changed.

To solve this issue, I created a new instance of TextEditingController in the initState method and passed it to the TextField as the controller. This successfully fixed the issue, and now the username input functions as expected.

a-wallen (changes requested):
Two files `pre-commit.disabled` and `go-gitlint_1.1.0_linux_x86_64.tar.gz.1` are being checked in with this PR but seem unrelated. Please restore those changes if you didn't mean to check them in.

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_user_view.dart
```diff
@@ -62,27 +62,47 @@ class SettingsUserView extends StatelessWidget {
...
   @override
   Widget build(BuildContext context) {
     return TextField(
-      controller: TextEditingController()..text = name,
+      controller: _controller,
       decoration: InputDecoration(
         labelText: LocaleKeys.settings_user_name.tr(),
       ),
-      onSubmitted: (val) {
+      onChanged: (val) {
```
a-wallen, line 93: Yes, this works, but it's expensive since we call the backend on every keystroke. Could you find a way to make that call only once? 

> https://stackoverflow.com/questions/51219330/flutter-run-function-after-showdialog-is-dismissed

LucasXu0: Or using `debounce`?

DestinedCodes: Hey @a-wallen, I noticed that there were several commits on a single file with no significant changes. Additionally, I had to revert the `pre-commit.disabled` and `go-gitlint_1.1.0_linux_x86_64.tar.gz.1` files. Adding debounce would also result in more unnecessary commits. To keep the branch clean, I restored it to its initial state and continued from where I stopped.

a-wallen:
@DestinedCodes do you still have plans for this?

DestinedCodes:
Yes I still do I'll give submit the test this weekend

appflowy:
> Yes I still do I'll give submit the test this weekend

Hey, @DestinedCodes. The status of this PR is from two weeks ago. Have you encountered any issues? Do you need any help?

PR #2079, opened 2023-03-22
Clear pending commits

rileyhawk1417:
Clear old commits then update with upstream.
This is to avoid carrying over 55 old commits and yet only one new commit was made.


rileyhawk1417:
Closing PR was trying to clear the old commits from my fork

Xazin:
If you want your forks main to be equal to the AppFlowy main, you can simply do:
```
git fetch upstream
git checkout main
git reset --hard upstream/main
```
Assuming you added the remote as upstream using `git remote upstream https://github.com/AppFlowy-IO/AppFlowy/`

If you do the above, remember to push it, `git push --force`

Issue #2016, opened 2023-03-17
[Bug] Tooltip text color is not coherent with dark theme, reduces readability. 


DarkFalc0n:
### Bug Description

Tooltip text color stays white even in the dark theme, this results in reduced readability as the tooltip background is also white(ish).

### How to Reproduce

Turn the settings to: 
Settings > Appearance > Theme Mode > Dark Mode

Then hover over any action button in the UI to reveal the tooltip.
![image](https://user-images.githubusercontent.com/59203815/225959276-eab55d2b-c114-4f71-abba-b62cfbc51892.png)

### Expected Behavior

The tooltip text should change to a darker color upon switching themes.

### Operating System

Windows

### AppFlowy Version(s)

0.1.0

### Screenshots

![image](https://user-images.githubusercontent.com/59203815/225959571-9cde3300-6567-4c81-9b7a-314e2dadba9c.png)

DestinedCodes:
I have been able to reproduce this issue and would like to work on it. Can you assign it to me? Thank you!!!

a-wallen:
@DestinedCodes, you got it!

hyj1204:
Hi guys, 
I have been working on the dark mode improvement recently and this tooltip text should be changed after it. 
@DestinedCodes  Is it okay to unsign you for this PR? 
Related to issue #1671 and PR #1987

hyj1204:
This issue has been solved. Please check the linked PR for detail.

PR #2099, opened 2023-03-23
Feat(appflowy_flutter): dark mode improvement for generic area/widgets

hyj1204:
This PR is the first part of the dark mode improvement #1671 . It update the UI in the generic area/widgets, like Side bar,  Top bar, question bubble and tool tip #2016 . The page area where plugins is haven't been updated. It will be updated in the next PR.

**The main change**
Due to the icon color on the button needs to change when it is hovered, I created a new widget `FlowySvg` to replace `svg Widget`(where has hover effect) in order to get the icon color when it is hovered from current theme data. 
image

And inside `FlowyHoverContainer`, the icon color in the ThemeData will be updated to the new icon color(when it is hovered). 
image
To clarify, the icon's color will be based on the current ThemeData when it is not being hovered over. However, when the user hovers over the icon, its color will temporarily change to a new color while still being based on the ThemeData. This allows the icon's color to be dynamically changed.

The following UI has been updated
1. All the elements in the Side bar
image

2. Hover effect in multiple locations
image
image
image

3. Font size switch menu
image

4. All the Tooltip
image

5. Question bubble
image

6. Setting Menu and DropdownMenuItem style
image


frontend/appflowy_flutter/packages/flowy_infra_ui/lib/style_widget/hover.dart
```diff
@@ -131,13 +131,27 @@ class FlowyHoverContainer extends StatelessWidget {
...
+                        .textTheme
+                        .bodyMedium
+                        ?.copyWith(
+                            color: Theme.of(context).colorScheme.onSurface),
+                  ),
+              iconTheme: Theme.of(context)
+                  .iconTheme
+                  .copyWith(color: Theme.of(context).colorScheme.onSurface)),
+          child: child!,
+        ));
```
LucasXu0, line 155: add a comma inside `))`.

LucasXu0:
LGTM.

PR #2100, opened 2023-03-23
fix: open-ai replace does not work in certain use-cases

### LINKED ISSUE #2092: [Bug] OpenAi replace does not work in certain use-cases
### Bug Description

As the title suggests, replace option when using OpenAi to generate a summary or fix-spelling does not work as expected in few use-cases, if response generated is multiline, last line overwrites the previous lines. This can cause much problem and confusion if user is working on large selections of text for correction or summary generation.

### How to Reproduce

1. Select some text on editor (likely a query to generate multiple lines from open ai)
2. select summarize and then replace the selected text.

### Expected Behavior

Selected text should be correctly replaced by the response generated by open-ai.

### Operating System

Linux

### AppFlowy Version(s)

Version 0.1.1 (1)

### Screenshots

https://user-images.githubusercontent.com/84044317/227204772-9077f453-4994-41da-9618-5ae431cac55e.mp4

### Additional Context

_No response_

squidrye:
Replacing lesser number of available multi-line text nodes with higher number of incoming multi-line nodes lead to inconsistencies such as overwriting and unexpected line skips, this PR fixes the issue.

fixes #2092

### DIFF
--- a/frontend/appflowy_flutter/packages/appflowy_editor/lib/src/core/transform/transaction.dart
+++ b/frontend/appflowy_flutter/packages/appflowy_editor/lib/src/core/transform/transaction.dart
@@ -373,12 +373,6 @@ extension TextTransaction on Transaction {
       final length = texts.length;
       var path = textNodes.first.path;
 
-      //FIND OFFSET CHARACTERS AFTER SELECTION
-      String offSetChar =
-      (document.nodeAtPath(selection.end.path) as TextNode)
-          .toPlainText()
-          .substring(selection.end.offset);
-
       for (var i = 0; i < texts.length; i++) {
         final text = texts[i];
         if (i == 0) {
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/packages/appflowy_editor/lib/src/core/transform/transaction.dart
```diff
@@ -396,15 +404,23 @@ extension TextTransaction on Transaction {
...
+                ),
+              ]);
+            } else {
+              document.insert(path, [
+                TextNode(
+                  delta: Delta()..insert(text),
+                ),
+              ]);
             }
-            insertNode(path, TextNode(delta: Delta()..insert(text)));
```
squidrye, line 405: This was more complicated than I thought, existing insertNode function transformed paths, which in case of raise issue caused unexpected line skips and loss of data after selection in the same path (line), so I ended up using document.insert() which seemed to work fine.

LucasXu0: Actually, you should not use `document.insert` to update the document content. The only way to update it is through operations, like insertNode, deleteNode, and so on.

squidrye: Alright, I'll try to work it out.

squidrye: I got it to work thanks for review @LucasXu0

LucasXu0:
Hey, @squidrye. I reimplement some parts of your code and formatted the code. Please take a look. https://github.com/AppFlowy-IO/AppFlowy/pull/2100/commits/6b13358cfc9d741ebdac9f71c965b138d2fe900f

Issue #2085, opened 2023-03-23
[FR] Load an AppFlowy workspace on startup during integration testing.


a-wallen:
### Description

Integration tests load the default AppFlowy workspace. The default workspace contains a demo document containing lots of text. It does not contain any boards, grids, or calendars that might be useful for an integration test. Any integration test that needs a calendar must programmatically create one through button clicks, making it tedious to write integration tests.

Proposal: contributors can configure workspaces and load those during an integration test. For example, I can create an AppFlowy workspace with a empty document and a Kanban board if I'd like to test Kanban functionality.

### Impact

Medium/high - improve test coverage and developer velocity.

a-wallen:
cc @LucasXu0

a-wallen:
Closing as pr#2083 is landed

PR #2106, opened 2023-03-25
[feat] Load workspace during integration test from a file.

a-wallen:
Uses a workspace uploaded from a .zip file during an integration test. Before any tests are run, the `TestWorkspaceService` unzips the workspace folder into `.ephemeral/$name`. This folder is ignored by source. After the workspace is unzipped, the `TestWorkspaceService` tells the application to read the workspace from `.ephemeral/$name`

If contributors need to write tests for Kanban board, they can create a new AppFlowy workspace with just a Kanban board, compress the folder, and add it to `test_workspaces` under `integration_test`.


a-wallen:
fixes https://github.com/AppFlowy-IO/AppFlowy/issues/2085

frontend/appflowy_flutter/integration_test/util/data.dart
```diff
@@ -0,0 +1,62 @@
...
+  const TestWorkspace(this._name);
+
+  final String _name;
+
+  Future<File> get zip async {
+    final Directory parent = await TestWorkspace._parent;
+    final File out = File(p.join(parent.path, '$_name.zip'));
+    if (await out.exists()) return out;
+    await out.create();
+    final ByteData data = await rootBundle.load(_asset);
```
a-wallen, line 22: Ran into issues reading files directly from the file system. There can be discrepancies between the local machine and the ones that run these tests on GitHub Actions. See [this issue](https://github.com/flutter/flutter/issues/94197).

.github/workflows/integration_test.yml
```diff
@@ -110,11 +110,11 @@ jobs:
         working-directory: frontend/appflowy_flutter
         run: |
           if [ "$RUNNER_OS" == "Linux" ]; then
-            flutter test integration_test -d Linux --coverage
+            flutter test integration_test/runner.dart -d Linux --coverage
```
a-wallen, line 113: This PR adds a new integration test file. Although running `flutter test integration test [-args]` should work here, it appears that integration tests are unhappy when split between files. We should keep integration tests separated for readability and we can run them all from one test file `runner.dart`. See [this issue for context](https://github.com/flutter/flutter/issues/101031).

frontend/appflowy_flutter/pubspec.yaml
```diff
@@ -165,6 +166,7 @@ flutter:
     - assets/images/common/
     - assets/images/grid/setting/
     - assets/translations/
+    - assets/test/workspaces/
```
LucasXu0, line 169: If you put the test resource here, won't it be included in the release product?

frontend/appflowy_flutter/integration_test/util/data.dart
```diff
@@ -0,0 +1,62 @@
...
+    await out.writeAsBytes(data.buffer.asUint8List());
+    return out;
+  }
+
+  Future<Directory> get root async {
+    return Directory(p.join(await TestWorkspace._parent.then((value) => value.path), name));
```
LucasXu0, line 28: A minor suggestion

```
  Future<Directory> get root async {
    final Directory parent = await TestWorkspace._parent;
    return Directory(p.join(parent.path, name));
  }
```

frontend/appflowy_flutter/pubspec.yaml
a-wallen: That's correct. Tomorrow, I'll try to see whether I can exclude those assets from the release build.

a-wallen: IIUC, files outside the `lib` directory won't be included in the build unless referenced in the `pubspec.yaml`. Hence, there needs to be some mechanism that removes this line from the pubspec.yaml for the release build.

Xazin: Could be an idea to have integration tests in a separate package (integration test package).

Otherwise we can write a script to amend the pubspec yaml, in that case the CI script should output it and await approval before releases, which is not that nice..

a-wallen: I think that having the integration tests as a separate package is a nice idea. It's better than writing a script to modify the `pubspec.yaml` I think that'll be confusing for someone that doesn't have context. I've never had to pull out an integration test suite into a separate package, so I'll let you know how that goes @Xazin.

LucasXu0: So, we're going to separate the integration_test package now?

a-wallen: I tried, but the integration tests can't run without a flutter project so it's not a good option anymore. 

Along the same lines, we can create a separate repo/branch that excludes the test assets with a .gitignore file. Another option is to use a build script to remove the files and updates the pubspec.yaml before running flutter build.

a-wallen:
Per @appflowy's request, the PR will use a build script to ignore the test assets in release. Assets between the `BEGIN: EXCLUDE_IN_RELEASE` and `END_EXCLUDE_IN_RELEASE` tags will be "removed" before building. To build the application, either run the dart script in `frontend/scripts/flutter_release_build/build_flowy.dart` or run the task `AF: flutter build release (host_platform)`.

frontend/scripts/flutter_release_build/tool.dart
```diff
@@ -0,0 +1,115 @@
...
+    // Read the contents of the file into a list
+    var lines = await file.readAsLines();
+
+    // Find the lines between BEGIN: EXCLUDE_IN_RELEASE and END: EXCLUDE_IN_RELEASE
+    var scanMode = _ScanMode.ignore;
+    for (var i = 0; i < lines.length; i++) {
+      var line = lines[i];
+      if (line.contains(excludeTagBegin)) {
+        scanMode = _ScanMode.target;
+      } else if (line.contains(excludeTagBegin)) {
```
squidrye, line 74: Should be this right?
```suggestion
      } else if (line.contains(excludeTagEnd)) {
```

a-wallen:
AppFlowy/AppFlowy#2083 had the changes in this PR and just landed. Closing this since it would be somewhat redundant.

Issue #2056, opened 2023-03-21
[FR] Show a toast after copying the path


LucasXu0:
### Description

<img width="458" alt="Screenshot 2023-03-21 at 09 59 26" src="https://user-images.githubusercontent.com/11863087/226502094-0d96952d-339c-4a42-8280-6fea6bd7199d.png">

Show a toast after double tapping to let the users know the path has been copied successfully.

### Impact

Users who copy the data path.

### Additional Context

Please refer to:

* `settings_file_customize_location_view.dart`

Sudhanva-Nadiger:
And tapping doesn't seem appropriate word in desktop mode, isn't it ! 
I suppose, it can be double click!

LucasXu0:
@Sudhanva-Nadiger. What's the difference between double tap and double click in desktop mode?

Sudhanva-Nadiger:
Tapping is in Mobile noe ?

BijayKumarPun:
@Sudhanva-Nadiger @LucasXu0, I'm wondering if a single click would suffice?

LucasXu0:
@BijayKumarPun. I think a single click would cause a misoperation.

Sudhanva-Nadiger:
https://user-images.githubusercontent.com/93595710/227791085-65ebf769-92fb-4b6f-b075-696f1eda18fa.mp4

HI @a-wallen 

### Why this approach is used?
- since the setting UI is a dialog we can't show a snack bar or a toast above it as its already on top of every other widget
- we can show the snack bar on the above level of dialog but before that, it had to be wrapped with a scaffold and make the background transparent which does not allow the closing of dialog on clicking outside of it. more info in this [blog](https://medium.com/@aakashpp/flutter-snack-bar-on-dialog-ea70a3b5e028#:~:text=showSnackBar()%3B%20inside%20the%20alert,behind%20the%20alert%20dialog%20screen.)

###

### how the above implementation is done?
- Wrapped the ListTile with a stateful builder so that the whole widget tree won't rebuild when we set the state
- whenever double tap we show the copied message for a second or so!

if any changes required or can I make a pr for this ?

Xazin:
> HI @a-wallen
> 
> ### Why this approach is used?
> * since the setting UI is a dialog we can't show a snack bar or a toast above it as its already on top of every other widget
[...]

PR #2114, opened 2023-03-26
fix: settings dialog files path copy snackbar

Xazin:
Relates: #2056 

<details>
  <summary>Screen recording of how it looks</summary>

Screen Recording 2023-03-26 at 22 09 50

</details>

### DIFF
--- a/frontend/appflowy_flutter/packages/flowy_infra_ui/lib/src/flowy_overlay/flowy_dialog.dart
+++ b/frontend/appflowy_flutter/packages/flowy_infra_ui/lib/src/flowy_overlay/flowy_dialog.dart
@@ -1,7 +1,7 @@
 import 'package:flutter/material.dart';
 import 'dart:math';
 
-const _overlayContainerPadding = EdgeInsets.all(12);
+const _overlayContainerPadding = EdgeInsets.symmetric(vertical: 12);
 const overlayContainerMaxWidth = 760.0;
 const overlayContainerMinWidth = 320.0;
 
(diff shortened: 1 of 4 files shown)


Xazin:
Not sure what is going on with the Snackbar theme, but that should be fixed in another issue.

a-wallen (approved):
LGTM

a-wallen:
@Xazin can you add an issue to fix the dialog theme?

Xazin:
> @Xazin can you add an issue to fix the dialog theme and mention this issue in it?

Will do.

a-wallen:
Also, please add a test at your convenience 😁

frontend/appflowy_flutter/packages/flowy_infra_ui/lib/src/flowy_overlay/flowy_dialog.dart
```diff
@@ -25,14 +25,14 @@ class FlowyDialog extends StatelessWidget {
...
     return SimpleDialog(
+        contentPadding: EdgeInsets.zero,
         title: title,
         shape: shape ??
             RoundedRectangleBorder(borderRadius: BorderRadius.circular(8)),
         children: [
           Material(
             type: MaterialType.transparency,
             child: Container(
-              padding: padding,
```
LucasXu0, line 35: Why do you delete this line?

Xazin: To not have the Snackbar have horizontal padding, but to fill the length of the dialog.

I've substituted it further down in the tree.

Xazin: I can also make it have the horizontal padding, and make the snackbar rounded, we need a design for snackbar anyway, so we can change it then. I wasn't able to find it at least.

PR #2122, opened 2023-03-28
fix: bold tests failed

LucasXu0:


frontend/appflowy_flutter/packages/appflowy_editor/lib/src/service/shortcut_event/built_in_shortcut_events.dart
Xazin, line 1: Line 287 also has `backquote` command which could just use the backtick character `

PR #2154, opened 2023-03-31
feat: Add more translations in Chinese(Traditional)

rhybs:
ref. #2153


LucasXu0:
@rhybs Thanks for contributing.

PR #2159, opened 2023-03-31
Cache rust target builds in CI

rileyhawk1417:
This PR aims to reduce the build time for the rust backend by caching the builds.
By enabling caching it reduces the time needed to download binaries and focuses on compiling instead


ankit-gautam23 (approved):
LGTM

PR #2162, opened 2023-03-31
Fix the app's Dockerfile and comment it

resolritter:
# Description

While trying to run the app through Docker (following https://appflowy.gitbook.io/docs/essential-documentation/install-appflowy/installation-methods/installing-with-docker) I ran into the following issues:

- Some mirrors were unavailable or timing out for me, e.g.
  `failed retrieving file 'libtiff-4.5.0-3-x86_64.pkg.tar.zst.sig' from geo.mirror.pkgbuild.com : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds`.

  Solved by mounting my own `/etc/pacman.d/mirrorlist` into the container.

- [The app is only compatible with Flutter 3.3.10](https://github.com/AppFlowy-IO/AppFlowy/issues/1741), but the Flutter version is not pinned in the Docker image.

  Solved by downloading the precise version according to https://docs.flutter.dev/get-started/install/linux#install-flutter-manually.

- I needed to run `xhost local:docker` in my host prior to starting the app, otherwise an error will occur

  Solved by running the aforementioned command and adding a note about it.

- The entrypoint `["./appflowy_flutter"]` apparently is incorrect

  Solved by figuring out where the executable is located within the generated bundle with `find . -executable -type f`

While going through those issues and solving them, I tinkered with the image and added some comments along the way.

# Trying it out

```console
$ cd frontend/scripts/docker-buildfiles
$ docker-compose rm -fvs
$ docker image rm --force appflowy/appflowy:latest
$ docker-compose up
```

### DIFF
--- a/frontend/scripts/docker-buildfiles/Dockerfile
+++ b/frontend/scripts/docker-buildfiles/Dockerfile
@@ -1,63 +1,80 @@
+#================
+# BUILDER
+#================
+
 FROM archlinux/archlinux:base-devel as builder
 
-RUN pacman -Syy
+# Upgrade the system
+RUN pacman -Syyu --noconfirm
 
-RUN pacman -Syu --needed --noconfirm git xdg-user-dirs
-
-# makepkg user and workdir
+# Set up makepkg user and workdir
 ARG user=makepkg
-ENV PATH="/home/$user/.pub-cache/bin:/home/$user/.local/flutter/bin:/home/$user/.local/flutter/bin/cache/dart-sdk/bin:${PATH}"
-RUN useradd --system --create-home $user \
-  && echo "$user ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers.d/$user
+RUN pacman -S --needed --noconfirm sudo
+RUN useradd --system --create-home $user && \
+    echo "$user ALL=(ALL:ALL) NOPASSWD:ALL" >> /etc/sudoers
+ENV PATH="/home/$user/.pub-cache/bin:/home/$user/flutter/bin:/home/$user/flutter/bin/cache/dart-sdk/bin:${PATH}"
 USER $user
 WORKDIR /home/$user
 
 # Install yay
-RUN git clone https://aur.archlinux.org/yay.git \
-  && cd yay \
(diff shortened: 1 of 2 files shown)


resolritter:
To avoid this problem from coming up again in the future, it might be worthwhile to test this Docker setup on CI. Let me know if you'd be interested in that.

LucasXu0:
@resolritter Sure. Can you help with adding the Docker build check on CI?.

resolritter:
> @resolritter Sure. Can you help with adding the Docker build check on CI?.

Alright, I'll try that. Moving to Draft in the meantime.

resolritter:
@LucasXu0 I think it's ready, but please take into account the following.

The CI to test `docker-compose build` consistently takes around 30m, as corroborated by https://github.com/resolritter/AppFlowy/actions/runs/4631817142/jobs/8195135626 and other runs of that workflow. Most of that time is unsurprinsingly spent at the Rust compilation steps; [step 21](https://github.com/resolritter/AppFlowy/actions/runs/4631817142/jobs/8195135626#step:3:745) (related to [step 22](https://github.com/resolritter/AppFlowy/actions/runs/4631817142/jobs/8195135626#step:3:1190)), which installs the Rust build tools, takes around 9 minutes; [step 25](https://github.com/resolritter/AppFlowy/actions/runs/4631817142/jobs/8195135626#step:3:1200) (related to [Step 26](https://github.com/resolritter/AppFlowy/actions/runs/4631817142/jobs/8195135626#step:3:2147)), which compiles the app, takes around 19 minutes.

At first I wanted to use caching to speed up the Rust compilation, but attempting that ended up significantly increasing the code's complexity - see https://github.com/resolritter/AppFlowy/blob/docker2/.github/workflows/docker_ci.yml (which is unfinished). That's why I'm submitting the PR without any caching whatsoever. If I manage to finish and refine the caching solution, I'll submit a follow-up PR.

LucasXu0:
@resolritter Thanks, it looks good to me. However, the Docker-CI is taking a lot of time. I think we should create a new issue to track and optimize it.

LucasXu0 (approved):

Issue #2082, opened 2023-03-22
[FR] Title of referenced board does not change in document if name is edited in the workspace sidebar.


a-wallen:
### Description

https://user-images.githubusercontent.com/44445638/227062156-f679050a-faed-4acf-8a2a-f609743e3d2a.mp4

The document widget needs to be reloaded entirely for the title of a referenced board to be updated.

### Impact

Minor fidelity impact.

Xazin:
This issue persists not only for Board, but also for grid. (Any referenced Page - same logic)

PR #2168, opened 2023-04-02
fix: update view on menu app rename

Xazin:
Relates: #2082

### DIFF
--- a/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/app/header/header.dart
+++ b/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/app/header/header.dart
@@ -1,5 +1,7 @@
 import 'dart:convert';
 
+import 'package:appflowy/plugins/document/application/doc_bloc.dart';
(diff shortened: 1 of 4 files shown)


frontend/appflowy_flutter/lib/workspace/presentation/home/menu/app/header/header.dart
```diff
@@ -86,6 +88,9 @@ class MenuAppHeader extends StatelessWidget {
                 value: context.read<AppBloc>().state.app.name,
                 confirm: (newValue) {
                   context.read<AppBloc>().add(AppEvent.rename(newValue));
+                  context
+                      .read<DocumentBloc>()
```
LucasXu0, line 92: The `DocumentBloc` isn't provided in the current context. Are you sure that you can access it here?

LucasXu0:
Hey, @Xazin. As your issue stated, you want to update the referenced board or grid name in Document when renaming them. However, the code you added is for listening to the workspace name, not the view name. Additionally, I don't think it's a good idea to reload the document every time the grid or board is renamed, especially if it doesn't contain any reference item.

PR #2261, opened 2023-04-14
Split create document and open document

qinluhe:
### Feature Preview

---

#### PR Checklist

#### PR Feature
- [x] Split create document and open document
- [x] Format doc changes
- [x]  Support to update textblock and doc title

### DIFF
--- a/frontend/rust-lib/flowy-document2/src/entities.rs
+++ b/frontend/rust-lib/flowy-document2/src/entities.rs
@@ -10,6 +10,13 @@ pub struct OpenDocumentPayloadPBV2 {
 }
 
 #[derive(Default, ProtoBuf)]
+pub struct CreateDocumentPayloadPBV2 {
+  #[pb(index = 1)]
+  pub document_id: String,
+  // Support customize initial data
+}
+
+#[derive(Default, ProtoBuf)]
 pub struct CloseDocumentPayloadPBV2 {
   #[pb(index = 1)]
   pub document_id: String,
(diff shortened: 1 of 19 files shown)


frontend/rust-lib/flowy-document2/src/entities.rs
```diff
@@ -102,3 +109,21 @@ impl Default for BlockActionTypePB {
...
+  pub is_remote: bool,
+}
+
+#[derive(Default, ProtoBuf)]
+pub struct BlockEventPB {
+  #[pb(index = 1)]
+  pub path: Vec<String>,
+
+  #[pb(index = 2)]
+  pub delta: String,
```
LucasXu0, line 128: What's the meaning of the `delta`?

frontend/rust-lib/flowy-document2/src/manager.rs
```diff
@@ -32,23 +33,45 @@ impl DocumentManager {
...
   }
 
+  pub fn create_document(&self, doc_id: String, data: DocumentDataWrapper) -> FlowyResult<Arc<Document>> {
+    self.get_document(doc_id, Some(data))
+  }
+
+  fn get_document(&self, doc_id: String, data: Option<DocumentDataWrapper>) -> FlowyResult<Arc<Document>> {
+    let collab = self.get_collab_for_doc_id(&doc_id)?;
+    let document = Arc::new(Document::new(collab)?);
+    self.documents.write().insert(doc_id, document.clone());
```
LucasXu0, line 43: I think we should put this line only in the `open_document` function, because we want to keep the document alive only when we need to listen to its callback.

LucasXu0: use `if let` will be better.

if let Some(data) = data { ... }

frontend/rust-lib/flowy-document2/src/manager.rs
```diff
@@ -32,23 +33,45 @@ impl DocumentManager {
...
+    if data.is_some() {
```
LucasXu0, line 44: use if let will be better.

```rs
if let Some(data) = data { ... }
```

PR #2287, opened 2023-04-17
feat: add url validation in cell_editor

### LINKED ISSUE #2199: [FR] Handle invalid URLs when formatting text as a link
### Description

On the link edition/creation dialog you can type anything and it's still accepted, it should only accept URL (from web URL to system (I don't know how they are called) URL like mailto, call, etc.)

In the link editing menu, we should let the user know "Type a valid URL to link"

### Impact

Users who format text as links

### Additional Context

_No response_

rwt-2001:
Fixes #2199 
### Feature Preview
image

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/core/helpers/url_validator.dart
+++ b/frontend/appflowy_flutter/lib/core/helpers/url_validator.dart
@@ -1,0 +1,13 @@
+import 'package:dartz/dartz.dart';
+
+Either<Uri, FormatException> parseValidUrl(String url) {
+  try {
+    final uri = Uri.parse(url);
+    if (uri.scheme.isEmpty || uri.host.isEmpty) {
+      throw const FormatException('Invalid URL');
+    }
+    return left(uri);
+  } on FormatException catch (e) {
+    return right(e);
+  }
+}
--- a/frontend/appflowy_flutter/lib/core/helpers/url_validator.dart
+++ b/frontend/appflowy_flutter/lib/core/helpers/url_validator.dart
@@ -1,13 +1,21 @@
 import 'package:dartz/dartz.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+part 'url_validator.freezed.dart';
 
-Either<Uri, FormatException> parseValidUrl(String url) {
+Either<UriFailure, Uri> parseValidUrl(String url) {
   try {
     final uri = Uri.parse(url);
     if (uri.scheme.isEmpty || uri.host.isEmpty) {
-      throw const FormatException('Invalid URL');
+      return left(const UriFailure.invalidSchemeHost());
     }
-    return left(uri);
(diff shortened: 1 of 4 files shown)


frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/url_cell/cell_editor.dart
```diff
@@ -73,7 +74,16 @@ class _URLCellEditorState extends State<URLCellEditor> {
     if (mounted) {
       if (_cellBloc.isClosed == false &&
           _controller.text != _cellBloc.state.content) {
-        _cellBloc.add(URLCellEditorEvent.updateText(_controller.text));
+        bool validURL = Uri.parse(_controller.text).isAbsolute;
```
Xazin, line 77: Are you sure you want to do `isAbsolute`? I'm not sure it does what you expect.

A link is valid with a fragment, and `isAbsolute` checks the validity of a link without a fragment. Meaning if you have a fragment in a link like this: `https://appflowy.gitbook.io/docs/essential-documentation/roadmap#guide-to-the-roadmap` where the fragment is this `#guide-to-the-roadmap` it will return false.

If the Uri is not valid, `Uri.parse` will throw a FormatException, see https://api.flutter.dev/flutter/dart-core/Uri/parse.html - That should be enough.

If you want you could extract this url check to a util class, as a method, or just as a util function loosely, like this:
```
/// Returns the parsed [Uri] if the URL is valid otherwise it
/// handles the [FormatException) and returns null 
Uri? parseValidUrl(String url) {
  try {
    return Uri.parse(url)
  } on FormatException {
    return null;
  }
}
```
Not the best tbf, normally the above code snippet would be improved by using Either from functional programming. But it's not really necessary to overcomplicate..

frontend/appflowy_flutter/lib/core/helpers/url_validator.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:dartz/dartz.dart';
+
+Either<Uri, FormatException> parseValidUrl(String url) {
```
Xazin, line 3: Should be something like
```
Either<UriFailure, Uri>
```
Doesn't make sense to catch an exception just to return or rethrow it.

frontend/appflowy_flutter/lib/core/helpers/url_validator.dart
```diff
@@ -0,0 +1,13 @@
...
+  try {
+    final uri = Uri.parse(url);
+    if (uri.scheme.isEmpty || uri.host.isEmpty) {
+      throw const FormatException('Invalid URL');
```
Xazin, line 7: No need to throw.
```
return left(UriFailure.invalidSchemeHost())
```

Xazin:
LGTM

PR #2343, opened 2023-04-24
Need to set the cursor to the correct position, after the DOM node is updated

qinluhe:
### Feature Preview

- [x] Need to set the cursor to the correct position, after the DOM node is updated

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
+++ b/frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
@@ -1,6 +1,6 @@
 import { triggerHotkey } from '@/appflowy_app/utils/slate/hotkey';
-import { useCallback, useContext, useState } from 'react';
-import { Descendant, Range, Editor, Element, Text, Location } from 'slate';
+import { useCallback, useContext } from 'react';
+import { Range, Editor, Element, Text, Location } from 'slate';
 import { TextDelta } from '$app/interfaces/document';
 import { useTextInput } from '../_shared/TextInput.hooks';
 import { useAppDispatch } from '@/appflowy_app/stores/store';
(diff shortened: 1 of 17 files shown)


frontend/appflowy_tauri/src/appflowy_app/stores/effects/document/document_controller.ts
```diff
@@ -124,11 +128,25 @@ export class DocumentController {
...
 
+  private composeDelta = (node: Node) => {
+    const delta = node.data.delta;
+    if (!delta) {
+      return;
+    }
+    // we use yjs to compose delta, it can make sure the delta is correct
+    // for example, if we insert a text at the end of the line, the delta will be [{ insert: 'hello' }, { insert: " world" }]
+    // but if we use yjs to compose the delta, the delta will be [{ insert: 'hello world' }]
+    const ydoc = new Y.Doc();
```
LucasXu0, line 139: How about we use `delta.js` to compose it? https://github.com/quilljs/delta

frontend/appflowy_tauri/src/appflowy_app/stores/reducers/document/async_actions/update.ts
```diff
@@ -0,0 +1,37 @@
...
+    const node = state.nodes[id];
+    const updateNode = {
+      ...node,
+      id,
+      data: {
+        ...node.data,
+        delta,
+      },
+    };
+    // Don't delay the update of the block map
```
LucasXu0, line 21: It would be better to add the reason here.

frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
```diff
@@ -10,65 +10,76 @@ import {
...
+  }, [editor]);
 
-        const { anchor } = editor.selection;
-        const isCollapsed = Range.isCollapsed(editor.selection);
-        if (isCollapsed && anchor.offset === 0 && anchor.path.toString() === '0,0') {
+  const onKeyDownCapture = useCallback(
+    (event: React.KeyboardEvent<HTMLDivElement>) => {
+      switch (event.key) {
+        // It should be handled when `Enter` is pressed
+        case 'Enter': {
```
LucasXu0, line 33: Just curious, will the 'shift+enter' keystroke trigger a response in this branch as well?

qinluhe: Yes, if we want the 'shift+enter' to trigger another action, we can filter the `shift` key

frontend/appflowy_tauri/src/appflowy_app/stores/effects/document/document_controller.ts
qinluhe: It is not really composed of delta and op, it just composes delta1 and delta2
for example
node1 delta: [{ insert: 'hello' }] 
node2 delta: [{ insert: 'world' }] 
when we compose node1 and node2, we will get [{ insert: 'hello' }, { insert: " world" }]
but we want the delta will be like [{ insert: 'hello world' }]
so we just put them in a ytext and get this ytext's delta

LucasXu0:
LGTM.

Issue #2239, opened 2023-04-11
[FR] Allow showing more properties on the calendar view


annieappflowy:
### Description

I assign tags using the multi-select property to events. I'd like to see these tags on my calendar view.

### Impact

Users who need more info shown on the calendar view

richardshiue:
This tasks has been completed, closing!

Issue #2183, opened 2023-04-04
[FR] Improve open as a page, supporting note-taking, page icon, and cover image


annieappflowy:
### Description

Currently, when a row is opened as a page, I can only edit properties.
I'd like to have a rich-text area where I can jot down a few things and use the features offered by the editor.
It would be better if I can decorate the page with a cover image and a page icon.
Once I close up the page, I should be taken back to the database where I opened up a row. Now I can see the page icon I just set up shown up next to the first column value of the row.

### Impact

Users who use `open as a page` to write extra notes for a record.

PR #2351, opened 2023-04-25
chore: grid row page detail redesign

### LINKED ISSUE #1993: [FR] Allow delete a row from the popup from "open as a page"
### Description

Title says all. For calendar, there is no readily accessible button like a hover in kanban or context menu in grid to delete a row and the only way that it can be edited right now is by opening it as a page. Therefore, it would be nice to be able to delete it from the popup window itself.

### Impact

Users who open rows of database as a page and want to delete them from there.

### Additional Context

_No response_

richardshiue:
This PR introduces a new design of the UI of opening grid row/kanban board card/calendar event as page.

- Add a right sidebar to include row actions. For now, there is only the ability to delete the row.
- Removed the thick scroll bar for a more subtle version. Also increase the scrollbar margin in the main axis to account for the rounded corners (doesn't seem to cause regressions on other popovers)
- Reorganized the layout of the page to allow future implementation of #2183.

fixes #1993

### Feature Preview

https://user-images.githubusercontent.com/71320345/234302277-b7fee5d4-cda2-4360-8374-db9dd5ee716b.mp4

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
@@ -3,6 +3,7 @@ import 'package:appflowy/plugins/database_view/application/field/type_option/typ
 import 'package:appflowy/plugins/database_view/application/row/row_data_controller.dart';
 import 'package:appflowy/plugins/database_view/grid/application/row/row_detail_bloc.dart';
 import 'package:appflowy/workspace/presentation/widgets/dialogs.dart';
+import 'package:appflowy_backend/log.dart';
 import 'package:flowy_infra/theme_extension.dart';
 import 'package:flowy_infra/image.dart';
 import 'package:flowy_infra_ui/flowy_infra_ui.dart';
@@ -14,6 +15,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:appflowy_popover/appflowy_popover.dart';
 
+import '../../application/row/row_service.dart';
 import '../../grid/presentation/layout/sizes.dart';
 import 'accessory/cell_accessory.dart';
 import 'cell_builder.dart';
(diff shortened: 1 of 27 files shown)


frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
```diff
@@ -43,147 +45,100 @@ class RowDetailPage extends StatefulWidget with FlowyOverlayDelegate {
...
-              const _Header(),
-              Expanded(
-                child: _PropertyColumn(
-                  cellBuilder: widget.cellBuilder,
-                  viewId: widget.dataController.viewId,
-                ),
+        child: ListView(
+          children: [
+            // const SizedBox(height: 100),
+            // const Divider(height: 1.0),
```
a-wallen, line 59:
```suggestion
```

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
```diff
@@ -43,147 +45,100 @@ class RowDetailPage extends StatefulWidget with FlowyOverlayDelegate {
...
+            IntrinsicHeight(
```
a-wallen, line 60: Is it possible to avoid using IntrinsicHeight here?

From [docs](https://api.flutter.dev/flutter/widgets/IntrinsicHeight-class.html)
> This class is relatively expensive, because it adds a speculative layout pass before the final layout phase. Avoid using it where possible. In the worst case, this widget can result in a layout that is O(N²) in the depth of the tree.

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
```diff
@@ -43,147 +45,100 @@ class RowDetailPage extends StatefulWidget with FlowyOverlayDelegate {
...
+              child: Row(
+                crossAxisAlignment: CrossAxisAlignment.start,
+                mainAxisSize: MainAxisSize.min,
+                children: [
+                  Flexible(
+                    flex: 4,
```
a-wallen, line 66: Are we sure that the flex factor should always be `4`? How does the property column look when the app is on a small screen? An idea here would be to use a greater flex factor when the screen width is smaller. 

```dart
enum FormFactor {
  small._(0),
  medium._(800),
  large._(1200);
  
  const FormFactor._(this.width);
  
  factory FormFactor.from(Size size) {
    final width = size.width;
    if (width > FormFactor.large.width) {
      return FormFactor.large;
    }
    
    if (width > FormFactor.medium.width) {
      return FormFactor.medium;
    }
    
    return FormFactor.small;
  }
  
  final double width;
}

// this is for illustration. the code is lazy.
int get flex => <FormFactor, int> {
      FormFactor.small: 2,
      FormFactor.medium: 3,
      FormFactor.large: 4,
    } [FormFactor.from(MediaQuery.of(context).size)];
```

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
```diff
@@ -43,147 +45,100 @@ class RowDetailPage extends StatefulWidget with FlowyOverlayDelegate {
...
+                      rowId: widget.dataController.rowId,
+                    ),
+                  ),
+                ],
               ),
-            ],
-          ),
+            ),
+            const Divider(height: 1.0),
+            const SizedBox(height: 200)
```
a-wallen, line 83: This doesn't seem to be doing anything useful on behalf of the user. Should it be removed?

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
```diff
@@ -43,147 +45,100 @@ class RowDetailPage extends StatefulWidget with FlowyOverlayDelegate {
...
+          padding: const EdgeInsets.fromLTRB(50, 50, 50, 20),
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              // const FlowyText.semibold(
+              //   "Row Title",
+              //   fontSize: 20,
+              // ),
+              // const VSpace(20),
```
a-wallen, line 115: Remove comments before checking in code.
```suggestion
```

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
```diff
@@ -377,3 +322,47 @@ GridCellStyle? _customCellStyle(FieldType fieldType) {
...
+  final String rowId;
+
+  _RowOptionColumn({required String viewId, required this.rowId, Key? key})
+      : _rowBackendService = RowBackendService(viewId: viewId),
+        super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Padding(
+      padding: const EdgeInsets.symmetric(horizontal: 20, vertical: 50),
```
a-wallen, line 337: This is a lot of padding. What does it look like on a smaller screen?

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
richardshiue: Hello @a-wallen! I tried as much as possible to avoid `IntrinsicHeight` but I really couldn't figure out a more eloquent solution. If I simply remove it, the `VerticalDivider` inside the `Row` will get a height of zero making it disappear. If I do `CrossAxxiAlignment.stretch`, the `Column`s get unbounded height and throw exceptions. I've also tried experimenting with drawing left and right borders on the columns, but the heights of the columns aren't equal either. Please let me know if you have any suggestions to do this properly 🙏

a-wallen: No worries. We can always fix it if it becomes an issue in the future.

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
richardshiue: I left this part and the other one above the `IntrinsicHeight` so as to demonstrate how the layout of the page could be done (with the future banner at the top, and the rich text area at the bottom) using a `ListView`. Without it, there wouldn't be a need to use `IntrinsicHeight` at all. What do you think?

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
richardshiue: Ditto

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/row_detail.dart
a-wallen: A comment above these widgets would be nice if that's the case. However, I would still like for these widgets to be removed since they're not doing anything :)

richardshiue: Done!

appflowy:
close #2381

appflowy:
close #2239

appflowy:
close #2377

PR #2365, opened 2023-04-27
Support pressing keyboard event to move cursor

qinluhe:
### Feature Preview

- [x] Support moving the cursor to the previous line with the same offset when pressing <b>`up`</b>, and the current cursor position is the upper edge
- [x] Support moving the cursor to the next line with the same offset when pressing **`down`,** and the current cursor position is the lower edge
- [x] Support pressing the **`left`** key to move the cursor to the end of the previous line, and the current cursor position is at the beginning of the line
- [x] Support pressing the **`right`** key to move the cursor to the beginning of the next line, and the current cursor position is at the end of the line

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
+++ b/frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
@@ -16,8 +16,14 @@ import {
   canHandleBackspaceKey,
   canHandleTabKey,
   onHandleEnterKey,
+  keyBoardEventKeyMap,
+  canHandleUpKey,
+  canHandleDownKey,
+  canHandleLeftKey,
+  canHandleRightKey,
 } from '@/appflowy_app/utils/slate/hotkey';
 import { updateNodeDeltaThunk } from '$app/stores/reducers/document/async_actions/update';
+import { setCursorPreLineThunk, setCursorNextLineThunk } from '$app/stores/reducers/document/async_actions/set_cursor';
 
 export function useTextBlock(id: string) {
   const { editor, onChange, value } = useTextInput(id);
@@ -48,17 +54,11 @@ export function useTextBlock(id: string) {
   };
 }
 
-// eslint-disable-next-line no-shadow
-enum TextBlockKeyEvent {
-  Enter,
-  BackSpace,
-  Tab,
-}
-
 type TextBlockKeyEventHandlerParams = [React.KeyboardEvent<HTMLDivElement>, Editor];
 
 function useTextBlockKeyEvent(id: string, editor: Editor) {
-  const { indentAction, backSpaceAction, splitAction, wrapAction } = useActions(id);
+  const { indentAction, backSpaceAction, splitAction, wrapAction, focusPreLineAction, focusNextLineAction } =
+    useActions(id);
 
   const dispatch = useAppDispatch();
   const keepSelection = useCallback(() => {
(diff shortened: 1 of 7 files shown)


frontend/appflowy_tauri/src/appflowy_app/components/document/_shared/TextInput.hooks.ts
```diff
@@ -161,13 +161,13 @@ function setSelection(editor: ReactEditor, currentSelection: TextSelection) {
   const { path, offset } = currentSelection.focus;
   // It is possible that the current selection is out of range
   const children = getDeltaFromSlateNodes(editor.children);
-  if (children[path[1]].insert.length < offset) {
+  const index = path[1];
```
appflowy, line 164: What happen if the length of the path is zero

qinluhe: The path is not allowed to be zero or null. Because the slate node is a `Two-dimensional array`, its path must be [number, number]

frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
```diff
@@ -105,10 +105,60 @@ function useTextBlockKeyEvent(id: string, editor: Editor) {
...
+      },
+    };
+  }, [focusPreLineAction]);
+
+  const rightEvent = useMemo(() => {
+    return {
+      key: keyBoardEventKeyMap.Right,
+      canHandle: canHandleRightKey,
+      handler: (...args: TextBlockKeyEventHandlerParams) => {
+        void focusNextLineAction({
```
LucasXu0, line 150: Why does both the `arrow.down` and `arrow.right` keys trigger the same handler `focusNextLineAction`? Shouldn't the `arrow.right key` trigger the something like `focusNextCharacter` handler instead?

frontend/appflowy_tauri/src/appflowy_app/utils/slate/text.ts
```diff
@@ -30,3 +42,163 @@ export function getInsertRangeBy(editor: Editor) {
...
+}
+
+/**
+ * get the selection of the beginning of the node
+ */
+export function getNodeBeginSelection(): TextSelection {
+  const point = {
+    path: [0, 0],
+    offset: 0,
+  };
```
LucasXu0, line 73: Is this code unused? Or do you intend to assign it to a value below?

`{ anchor: point, focus: point }`.

frontend/appflowy_tauri/src/appflowy_app/utils/slate/text.ts
```diff
@@ -30,3 +42,163 @@ export function getInsertRangeBy(editor: Editor) {
...
+
+/**
+ * get the selection of the end of the node
+ * @param delta
+ */
+export function getNodeEndSelection(delta: TextDelta[]) {
+  const len = delta.length;
+  const offset = len > 0 ? delta[len - 1].insert.length : 0;
+  const cursorPoint: SelectionPoint = {
+    path: [0, len > 0 ? len - 1 : 0],
```
LucasXu0, line 95: max(len - 1, 0)

frontend/appflowy_tauri/src/appflowy_app/components/document/TextBlock/TextBlock.hooks.ts
qinluhe: we handle these events reason is used to focus on another block. we did not handle these events when we just want to focus next char

frontend/appflowy_tauri/src/appflowy_app/utils/slate/text.ts
qinluhe: It is not allowed, it will use the same obj

LucasXu0:
LGTM.

frontend/appflowy_tauri/src/appflowy_app/components/document/_shared/TextInput.hooks.ts
appflowy: Got it. it would be better if you add comments there.

PR #2390, opened 2023-04-29
feat: rewrite date logic

richardshiue:
### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/database_view/application/cell/cell_data_persistence.dart
```diff
@@ -29,8 +29,8 @@ class TextCellDataPersistence implements CellDataPersistence<String> {
 @freezed
 class DateCellData with _$DateCellData {
   const factory DateCellData({
-    required DateTime date,
-    String? time,
+    required DateTime? dateTime,
+    required String? time,
```
a-wallen, line 33: These are optional, but IIUC, at least one of them, shouldn't be null. We could consider adding an assertion to check that one is provided.

Also, consider exposing two factory constructors rather than one with two optional parameters. Here's an example.

```dart
class Example {
  factory Example.fromString({required String s}) => Example._(int.parse(s));
  factory Example.fromInt({required int i}) => Example._(i);
  // the main constructor here is private so that we don't have to have logic that deals with initializing 
  // the backing property, each constructor above does that for us.
  Example._(this.backingProperty);
  final int backingProperty;
}
```

a-wallen: Again, these are fly-by comments. You can correct me if I'm wrong :)

frontend/appflowy_flutter/lib/plugins/database_view/application/field/type_option/date_bloc.dart
```diff
@@ -54,9 +54,9 @@ class DateTypeOptionBloc
         typeOption.timeFormat = timeFormat;
       }
 
-      if (includeTime != null) {
-        typeOption.includeTime = includeTime;
-      }
+      // if (includeTime != null) {
+      //   typeOption.includeTime = includeTime;
+      // }
```
a-wallen, line 59:
```suggestion
```

remove commented code if it isn't being used.

frontend/appflowy_flutter/lib/plugins/trash/src/trash_cell.dart
```diff
@@ -62,10 +62,8 @@ class TrashCell extends StatelessWidget {
 
   String dateFormatter($fixnum.Int64 inputTimestamps) {
     final outputFormat = DateFormat('MM/dd/yyyy hh:mm a');
-    final date = DateTime.fromMillisecondsSinceEpoch(
-      inputTimestamps.toInt() * 1000,
-      isUtc: true,
-    );
+    final date =
+        DateTime.fromMillisecondsSinceEpoch(inputTimestamps.toInt() * 1000);
```
a-wallen, line 66: Consider adding an extension on DateTime called `fromSecondsSinceEpoch` that takes the inputTimeStamp and does the multiplication. That would consolidate this "magic" number.

a-wallen (commented):
cc @appflowy for review on rust portion

frontend/appflowy_flutter/lib/plugins/database_view/application/cell/cell_data_persistence.dart
richardshiue: Done! Removed required fields if nullable

frontend/appflowy_flutter/lib/plugins/database_view/application/field/type_option/date_bloc.dart
richardshiue: Deleted, resolving thread

frontend/appflowy_flutter/lib/plugins/trash/src/trash_cell.dart
Xazin: You can always make it a static method or getter in the extension and not a constructor.

richardshiue: Hey @Xazin, thanks for the suggestion. Static method works fine:

```
void main() {
  final timestamp = DateTime.now().millisecondsSinceEpoch ~/ 1000;
  final time = DateTimeExtension.fromSecondsSinceEpoch(timestamp);
  print(time);
}

extension DateTimeExtension on DateTime {
  static DateTime fromSecondsSinceEpoch(int seconds) =>
      DateTime.fromMillisecondsSinceEpoch(seconds * 1000);
}
```

But if I had to choose between the original and this one, I'd much prefer the original with comments since this introduces a DateTimeExtension out of nowhere, which doesn't improve readability much. As for the getter method, do you have an example code snippet? I don't quite get what you mean.

Xazin: Ah yes, I see. 

Leave it as is for now, maybe extract `1000` to a const `const millisToSeconds = 1000` or a function `millisToSeconds(int milliseconds) => milliseconds ~/ 1000`, or both...

I'm not sure if we will have support for the above in the near future, I doubt it though.

richardshiue: Alrighty then, I'm going to resolve this thread

Issue #2404, opened 2023-05-01
[FR] Calendar events on the same date should be displayed in a chronological order


annieappflowy:
### Description

<img width="190" alt="image" src="https://user-images.githubusercontent.com/12026239/235407416-4e485861-a30a-480e-8cba-e88ab2858a7d.png">

An 8:00 AM event should be above a 16:00 PM event on the same date although the 8:00 AM event was created after the 16:00 PM event

### Impact

Calendar users

annieappflowy:
Ticking off a checkbox shouldn't affect the order
<img width="142" alt="image" src="https://user-images.githubusercontent.com/12026239/235407686-f640fabb-6f73-4cfc-81a7-4fafd85e279b.png">

annieappflowy:
And the May 3rd's event, Pilates, is in the May 4th cell ?

PR #2412, opened 2023-05-01
fix: sort the events in each day cell by time

richardshiue:
### Feature Preview

fix: #2404

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart
@@ -48,14 +48,13 @@ class CalendarDayCard extends StatelessWidget {
     return ChangeNotifierProvider(
       create: (_) => _CardEnterNotifier(),
       builder: (context, child) {
-        List<GestureDetector> cards = _buildCards(context);
-
         Widget? multipleCards;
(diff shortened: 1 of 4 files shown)


frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart
```diff
@@ -98,6 +98,9 @@ class CalendarDayCard extends StatelessWidget {
   }
 
   List<GestureDetector> _buildCards(BuildContext context) {
+    events.sort(
```
a-wallen, line 101: This sorts the events every time the build method is called, no? If you convert this to a stateful widget, then you can sort the events in `initState()`.

a-wallen: This was just a fly-by comment. I realized that converting to a stateful widget might not be the most appropriate solution. My only concern is that sorting can get expensive if we sort the events on every build.

appflowy: I moved it to upper level that construct this widget.

appflowy:
Build the card for each event  instead of building all of them.
image

Issue #2354, opened 2023-04-25
[Bug] The Numbers column doesn't support negative numbers


annieappflowy:
### Bug Description

When I enter -20, it displays 20

### How to Reproduce

Create a table-view database
Add a Numbers column
Enter -20

### Expected Behavior

Negative numbers can be properly displayed

### Operating System

Windows 10

### AppFlowy Version(s)

v0.1.3

0xkelvin:
i think this issue related to my implementation on Rust part, so let me take it

VenomSupreme:
Here's a small vid for better understanding of the issue. 
I'm pressing ENTER key after entering the numbers.


https://user-images.githubusercontent.com/69515491/235305725-8300cff8-44f0-422a-b909-afd38f44ff02.mp4

PR #2443, opened 2023-05-03
fix: number format

appflowy:


appflowy:
close #2354

frontend/rust-lib/flowy-database/src/services/field/type_options/date_type_option/date_tests.rs
```diff
@@ -72,70 +72,70 @@ mod tests {
...
-            Some("11:23 pm".to_owned()),
-            "May 27,2022 11:23 PM",
-            true,
-            &field_rev,
-          );
-        },
-      }
-    }
-  }
+  // #[test]
```
Xazin, line 75: I see this is merged, but a few tests are commented out.

Should there be a todo to fix these, or should they be completely removed?

PR #2453, opened 2023-05-03
Feat/number list block

qinluhe:
### Feature Preview

- [x] Support Bulleted List Block via `* ` or `- ` or `+ `
- [x] Support Numbered List Block via `1. or 2. or 3. `
- [x] Calculate the current numbered index according to the type and index of the previous sibling node

image

---

#### PR Checklist


frontend/appflowy_tauri/src/appflowy_app/components/document/NumberedListBlock/NumberedListBlock.hooks.ts
```diff
@@ -0,0 +1,26 @@
...
+      return nodes[id].type !== BlockType.NumberedListBlock;
+    });
+    if (lastIndex === -1) return prevNodeIds.length;
+    return lastIndex;
+  });
+
+  return {
+    index: prevNumberedIndex + 1,
+  };
+}
```
LucasXu0, line 26:
```ts
    const { nodes, children } = useAppSelector((state) => state['document']);
    const parent = nodes[node.parent!];
    const siblings = children[parent.children];
    const index = siblings.indexOf(node.id);
    if (index === 0) return 1;
    const prevNodeIds = siblings.slice(0, index);
    // The index is distance from last block to the last non-numbered-list block
    const lastIndex = prevNodeIds.reverse().findIndex((id) => {
      return nodes[id].type !== BlockType.NumberedListBlock;
    });
    const prevNumberedIndex = lastIndex === -1 ?  prevNodeIds.length : lastIndex;
    return {
      index: prevNumberedIndex + 1,
    };
```

It seems the state only used in the first two line. Is it necessary to wrap them in a function?

qinluhe: Your code is equivalent to subscribing to all state updates, not only when the index changes

LucasXu0:
LGTM

PR #2485, opened 2023-05-09
ci: tauri build

appflowy:


Xazin:
Consider removing `yarn` from the `test:prettier` script (`"test:prettier": "prettier --list-different src",`)

appflowy:
> Consider removing `yarn` from the `test:prettier` script (`"test:prettier": "prettier --list-different src",`)

Done

PR #2493, opened 2023-05-10
fix: don't show whitespace when no options are selected

richardshiue:
Notice how there is extra whitespace when the visibility of the select option field is being toggled

### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/365c9535-1e40-491f-9bd8-d7d0821d833f

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart
@@ -162,6 +162,9 @@ class CalendarDayCard extends StatelessWidget {
     });
 
     renderHook.addSelectOptionHook((selectedOptions, cardData, _) {
+      if (selectedOptions.isEmpty) {
+        return const SizedBox.shrink();
+      }
       final children = selectedOptions.map(
         (option) {
           return SelectOptionTag.fromOption(


frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart
```diff
@@ -162,6 +162,9 @@ class CalendarDayCard extends StatelessWidget {
     });
 
     renderHook.addSelectOptionHook((selectedOptions, cardData, _) {
+      if (selectedOptions.isEmpty) {
+        return const SizedBox();
```
Xazin, line 166:
```suggestion
        return const SizedBox.shrink();
```

richardshiue: Hey @Xazin, I could be wrong this seems rather unnecessary since there is no child widget inside the `SizedBox`?

Xazin: You want it to take up the least amount if space possible. Exactly what the `.shrink()` constructor is made for.

LucasXu0: @richardshiue, I agree with @Xazin. As the flutter documentation said, using `Sizedbox.shrink` would be better/

> The simplest way to hide a child is just to not include it, or, if a child must be given (e.g. because the parent is a [StatelessWidget](https://api.flutter.dev/flutter/widgets/StatelessWidget-class.html)) then to use [SizedBox.shrink](https://api.flutter.dev/flutter/widgets/SizedBox/SizedBox.shrink.html) instead of the child that would otherwise be included.

a-wallen: +1

richardshiue: Got it thanks guys for the advice! I've update the PR

PR #2516, opened 2023-05-12
fix: tauri tests

appflowy:


frontend/appflowy_tauri/src/appflowy_app/utils/change_notifier.ts
```diff
@@ -1,17 +1,29 @@
...
 
-  get observer() {
+  get observer(): Observable<T> | null {
+    if (this.isUnsubscribe) {
+      return null;
+    }
     return this.subject.asObservable();
   }
 
+  // Unsubscribe the subject might cause [UnsubscribedError] error if there is
```
qinluhe, line 18: Maybe we don't need to add judgment here, just need to ensure the execution of the `subscription.unsubscribe();`

Issue #2476, opened 2023-05-08
[Bug] Filter UI overflow


annieappflowy:
### Bug Description

<img width="426" alt="image" src="https://user-images.githubusercontent.com/12026239/236753711-6937b9d7-aabe-42c7-8520-cef8f12e1c4c.png">

### How to Reproduce

Add two filters to a grid
The second filter filters on a Text column with a long property name

### Expected Behavior

No UI overflow message

### Operating System

MacOS

### AppFlowy Version(s)

v0.1.4

Samiksha-Garg:
Hello! Can I work on this issue?

Xazin:
<img width="682" alt="Screenshot 2023-05-08 at 20 51 23" src="https://user-images.githubusercontent.com/42929161/236920072-529f9836-384f-4dda-9692-dfc64f143f3c.png">

This should already have been resolved, this is on `main`, as seen in the screenshot, actions have been moved above the current applied filters.

Samiksha-Garg:
Hey, @Xazin! I don't think this issue is about filter panel, I feel this is about editing a particular filter with long name. This is the screenshot of editor panel for a long column name on main branch : 
<img width="1117" alt="Screenshot 2023-05-09 at 1 46 11 AM" src="https://user-images.githubusercontent.com/79906086/236925554-9f2c9929-bd26-4a02-a5bc-abfdb785601e.png">

PR #2554, opened 2023-05-17
fix: Filter UI popover pixels overflow with dynamic width and height

gitstart:
fixes [#2476](https://github.com/AppFlowy-IO/AppFlowy/issues/2476)
### PR Description
This PR fixes pixels overflow issue using by `Expandable` and text overflow rule `overflow: TextOverflow.ellipsis`.

### Issue Recording
https://www.loom.com/share/007bae93eabb4d819fc169d0239a15ff

###  Fix Recording
https://www.loom.com/share/0246a6cdc88f42558716c7364279b4aa

---
### PR Checklist
- [x]  I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.
---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/filter/choicechip/text.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/filter/choicechip/text.dart
@@ -117,18 +117,25 @@ class _TextFilterEditorState extends State<TextFilterEditor> {
       height: 20,
       child: Row(
         children: [
-          FlowyText(state.filterInfo.fieldInfo.name),
-          const HSpace(4),
-          TextFilterConditionPBList(
-            filterInfo: state.filterInfo,
-            popoverMutex: popoverMutex,
-            onCondition: (condition) {
-              context
-                  .read<TextFilterEditorBloc>()
-                  .add(TextFilterEditorEvent.updateCondition(condition));
-            },
+          Expanded(
+            child: FlowyText(
+              state.filterInfo.fieldInfo.name,
+              overflow: TextOverflow.ellipsis,
+            ),
           ),
-          const Spacer(),
+          const HSpace(4),
+          Expanded(
+            child: TextFilterConditionPBList(
+              filterInfo: state.filterInfo,
+              popoverMutex: popoverMutex,
+              onCondition: (condition) {
+                context
+                    .read<TextFilterEditorBloc>()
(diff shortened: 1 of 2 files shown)


a-wallen:
This PR needs a test in order to land.

Regardless, I'm not sure about this solution. `IntrinsicHeight` squares the complexity of the layout pass for it's subtree and `TextFilterEditor` already has an `IntrinsicHeight` widget. IIUC, the subtree would have a O((N²)²) layout pass. 
> This class is relatively expensive, because it adds a speculative layout pass before the final layout phase. Avoid using it where possible. In the worst case, this widget can result in a layout that is O(N²) in the depth of the tree.

I wonder if defining overflow behavior here (in `text.dart`) would solve this problem without squaring the layout pass for layer subtree.
https://github.com/AppFlowy-IO/AppFlowy/blob/f8f0599462ff0cca8b08dabd47f450147f3beecd/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/filter/choicechip/text.dart#L120

Xazin:
Couldn't we just eg. layout the title of the property different, eg. flexible in a row with some overflow behavior (ellipsis or clip).

That will probably more or less resolve the overflow, I'm not saying it will look the best though. But dynamically sized popovers could be a bit of a challenge.

a-wallen:
> Couldn't we just eg. layout the title of the property different, eg. flexible in a row with some overflow behavior (ellipsis or clip).
> 
> That will probably more or less resolve the overflow, I'm not saying it will look the best though. But dynamically sized popovers could be a bit of a challenge.

Agreed. @gitstart, thanks for the contribution, could we try with a different approach? This is a solution to the problem, but a different approach might be more appropriate. I apologize for any confusion. Please message me on discord a-wallen#0603 if you need any help!

gitstart:
> > Couldn't we just eg. layout the title of the property different, eg. flexible in a row with some overflow behavior (ellipsis or clip).
> > That will probably more or less resolve the overflow, I'm not saying it will look the best though. But dynamically sized popovers could be a bit of a challenge.
> 
> Agreed. @gitstart, thanks for the contribution, could we try with a different approach? This is a solution to the problem, but a different approach might be more appropriate. I apologize for any confusion. Please message me on discord a-wallen#0603 if you need any help!

Hello @a-wallen,
Requested changes are pushed using different approach. 
Thank you!.

a-wallen (approved):

Xazin (approved):

PR #2478, opened 2023-05-08
feat: created at and updated at field type

zoli:
### Feature Preview

https://user-images.githubusercontent.com/3286303/236849765-ac1db61e-b8e9-4f36-a007-ef79ca277798.mp4

Resolves #1668, #1669

---

#### PR Checklist

- [x] My code adheres to the [AppFlowy Style Guide](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/submitting-code/style-guides)
- [x] I've listed at least one issue that this PR fixes in the description above.
- [x] I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.
- [x] All existing tests are passing.

Issue #1669, opened 2023-01-05
[FR] Support the Last Edited property


annieappflowy:
### Description

Automatically update the last edited time of a row

### Impact

Database users

zoli:
Now that develop is merged to main branch, I think this issue can get closed.

PR #2572, opened 2023-05-20
feat: created at and updated at field type

### LINKED ISSUE #1668: [FR] Support the Date Created property
### Description

Allow users to use an automatically updated field to indicate when a row was created

### Impact

Database users 

### Additional Context

_No response_

zoli:
This is continued from https://github.com/AppFlowy-IO/AppFlowy/pull/2478

### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/3286303/9707c500-c72a-4fea-a3d0-429b5b5ba3c0

Resolves #1668, #1669

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/database_view/application/filter/filter_service.dart
```diff
@@ -84,10 +84,19 @@ class FilterBackendService {
...
     int? end,
     int? timestamp,
   }) {
+    assert(
+      [
+        FieldType.DateTime,
+        FieldType.UpdatedAt,
+        FieldType.CreatedAt,
+      ].contains(fieldType),
+    );
```
LucasXu0, line 98: In addition to the assert, I suggest returning it and adding an assert inside. This is because in release mode, the assert won't have any effect, and we should filter out incompatible parameters.

```dart
if (![...].contains(fileType)) {
  assert('some message').
  return;
}
```

frontend/rust-lib/flowy-database2/src/services/cell/cell_operation.rs
```diff
@@ -364,6 +366,19 @@ impl CellBuilder {
...
+      if !cell_by_field_id.contains_key(&field.id) {
+        let field_type = FieldType::from(field.field_type);
+        if field_type == FieldType::UpdatedAt || field_type == FieldType::CreatedAt {
+          cells.insert(
+            field.id.clone(),
+            insert_date_cell(timestamp(), Some(true), field),
+          );
+        }
+      }
+    }
```
LucasXu0, line 381: An option.

```suggestion
    let auto_fill_files = fields
      .iter()
      .filter(|field| !cell_by_field_id.contains_key(&field.id))
      .filter(|field| {
        let field_type = FieldType::from(field.field_type);
        field_type == FieldType::UpdatedAt || field_type == FieldType::CreatedAt
      })
      .map(|field| {
        (
          field.id.clone(),
          insert_date_cell(timestamp(), Some(true), field),
        )
      })
      .collect::<HashMap<String, Cell>>();
    cells.extend(auto_fill_files);
```

frontend/appflowy_flutter/lib/plugins/database_view/application/filter/filter_service.dart
zoli: Maybe throw an exception? Because the return can't be empty.

appflowy:
@zoli I am going to make some changes on this PR. If you have any suggestions, please let me know.

appflowy:
image

Add FieldType to the DateTypeOption. So multiple field type can share the same type option.

zoli:
> @zoli I am going to make some changes on this PR. If you have any suggestions, please let me know.

Thanks, Nothing came to my mind about the recent changes.

richardshiue:
Just a suggestion, and I don't know if everyone will agree, but maybe we could use "Last modified" instead of "Updated at"?

appflowy:
> Just a suggestion, and I don't know if everyone will agree, but maybe we could use "Last modified" instead of "Updated at"?

Yep. Agreed. `last edited time` and `created time`.

Issue #2346, opened 2023-04-25
[Bug] Integration tests often fail


richardshiue:
### Bug Description

The empty document and typing `/grid` to create a new grid integration test often fails during GitHub pull CI checks, even though the pr didn't have such changes and the error cannot be reproduced manually outside of testing.

```
Tests on a workspace with only an empty document /grid shortcut creates a new grid and view of the grid (failed)

  The following TestFailure was thrown running a test:
  Expected: at least 2 matching nodes in the widget tree
    Actual: _WidgetTypeFinder:<zero widgets with type "SelectionMenuItemWidget" (ignoring offstage
  widgets)>
     Which: means none were found but some were expected
```

### How to Reproduce

Make a flutter-related pr and observe that the integration test fails. 

### Expected Behavior

The test shouldn't fail so often.

### Operating System

CI ubuntu-latest

### AppFlowy Version(s)

0.1.3

a-wallen:
Thanks for opening this issue. We are still recovering from the upgrade to flutter 3.7.5, which disabled OpenGL support and caused our integration tests on macOS to fail. We are now running the integration tests on Linux, but I think some race condition causes this issue to happen while all the tests in CI are running.

Aside, `macos (build)` also flakes. That needs to be addressed after this is addressed too.

a-wallen:
https://github.com/AppFlowy-IO/AppFlowy/pull/2622 fixes hard failure in CI for `open_ai_smart_menu_test`

a-wallen:
It looks like https://github.com/AppFlowy-IO/AppFlowy/pull/2621 is solving the broken tests in `Flutter-CI` for Windows. The next step is solving the errors with `pumpAndSettle` awaiting for an infinite amount of time because of the circular progress indicator.

PR #2011, opened 2023-03-17
fix: bug causing permanent CircularProgressIndicator when referenced board is deleted fixes #2001

DestinedCodes:
If you add a referenced board or grid to a document and delete the board, the document will have a permanent circular progress indicator.

This pull request fixes a bug where deleting a referenced board results in a permanent circular progress indicator on the page. The issue occurs because the app is not handling the case where the referenced board is deleted.

To fix this issue, I added a check for whether the referenced board is null. If it is null, I returned `SizedBox.shrink()` to remove the board from the page. Additionally, I added a check for if the `FutureBuilder` has an error, and returned `SizedBox.shrink()` in that case as well.

I tested this fix by creating a new board, referencing it on a page, deleting the board, closing and reopening the app, and checking that the circular progress indicator was no longer present.

Please review and merge this pull request.

Thank you!

PR #2622, opened 2023-05-26
Revert "Support regenerating ai results #2012" (fix broken integration tests).

a-wallen:
`integration test` has been flaky for a while, and I'm fixing it. I noticed that integration tests for `open_ai_smart_menu_test` have been failing consistently lately. I ran a `git bisect` and tracked it to this PR.

**Issue** https://github.com/AppFlowy-IO/AppFlowy/issues/2346

Link to the workflow to show that tests failed when this PR was introduced
> https://github.com/AppFlowy-IO/AppFlowy/pull/2528/checks

PR that introduced `open_ai_smart_menu_test` (passed when introduced).
> https://github.com/AppFlowy-IO/AppFlowy/actions/runs/4869490828/jobs/8684057865

PR #2635, opened 2023-05-26
[Fix] Integration test flakes.

a-wallen:
**Issue** https://github.com/AppFlowy-IO/AppFlowy/issues/2346

Tests on a workspace with an empty document flakes. It [succeeds in certain runs](https://github.com/AppFlowy-IO/AppFlowy/actions/runs/5086697721/jobs/9141396746#step:11:3083), but [fails in others](https://github.com/AppFlowy-IO/AppFlowy/actions/runs/5059606445/jobs/9081384873#step:11:3069).

Here are the two exceptions that I caught for the flaky test.
```
1. Tests on a workspace with only an empty document /grid shortcut creates a new grid and view of the grid (failed)

  The following TestFailure was thrown running a test:
  Expected: at least 2 matching nodes in the widget tree
    Actual: _WidgetTypeFinder:<zero widgets with type "SelectionMenuItemWidget" (ignoring offstage
  widgets)>
     Which: means none were found but some were expected

-- and --

2. The following assertion was thrown while running async test code:
pumpAndSettle timed out

When the exception was thrown, this was the stack:
#0      WidgetTester.pumpAndSettle.<anonymous closure> (package:flutter_test/src/widget_tester.dart:651:11)
<asynchronous suspension>
<asynchronous suspension>
```

Different errors happen in CI, but they all have the exact root cause.
1. The first test run on this workspace creates a new view of a board. https://github.com/AppFlowy-IO/AppFlowy/blob/18672b197cf8e4ef79b8bfd7cbeae90bbc7f9da8/frontend/appflowy_flutter/integration_test/empty_document_test.dart#L34
2. Since the previous test didn't clean the workspace, the editor looks for the board created in the previous test. 
3. The workspace can't find the board, so it returns a `CircularProgressIndicator.` (actively being worked on in https://github.com/AppFlowy-IO/AppFlowy/pull/2011/files). 
4. The second test runs and calls `pumpAndSettle` after it requests focus from the editor.
https://github.com/AppFlowy-IO/AppFlowy/blob/18672b197cf8e4ef79b8bfd7cbeae90bbc7f9da8/frontend/appflowy_flutter/integration_test/empty_document_test.dart#LL93C8-L93C8
5. The call never settles because of the infinite animation for `CircularProgressIndicator.` 

**Solution**: Remove the unzipped workspace before running the next test. 

Since the editor reads from the file system, all documents must be cleaned between tests!

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/integration_test/board_test.dart
+++ b/frontend/appflowy_flutter/integration_test/board_test.dart
@@ -34,6 +34,8 @@ void main() {
   group('board', () {
     setUpAll(() async => await service.setUpAll());
     setUp(() async => await service.setUp());
+    tearDown(() async => await service.tearDown());
+    tearDownAll(() async => await service.tearDownAll());
 
     testWidgets(
         'integration test unzips the proper workspace and loads it correctly.',
--- a/frontend/appflowy_flutter/integration_test/empty_document_test.dart
+++ b/frontend/appflowy_flutter/integration_test/empty_document_test.dart
@@ -30,6 +30,8 @@ void main() {
   group('Tests on a workspace with only an empty document', () {
     setUpAll(() async => await service.setUpAll());
     setUp(() async => await service.setUp());
+    tearDown(() async => await service.tearDown());
+    tearDownAll(() async => await service.tearDownAll());
 
     testWidgets('/board shortcut creates a new board and view of the board',
         (tester) async {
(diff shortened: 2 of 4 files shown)


a-wallen:
This should land after https://github.com/AppFlowy-IO/AppFlowy/pull/2622. `integration test` has hard failures due to a PR that landed although the PR failed CI. I'd like to resolve that first and re-run these checks against that pipeline.

Issue #1070, opened 2022-09-16
[FR] Create a new theme for AppFlowy Editor


annieappflowy:
Add a new theme to the theme library.

Reference: https://github.com/AppFlowy-IO/AppFlowy/blob/main/frontend/app_flowy/packages/appflowy_editor/documentation/customizing.md#customizing-a-theme-new-feature-in-005-alpha

ChiragKr04:
@annieappflowy Can I work on this? And do we need to add only the theme is given in reference?

annieappflowy:
@ChiragKr04 , I assigned this issue to you. Can you elaborate on your question?

MayurSMahajan:
I know the issue is already assigned but I had a small doubt. You want new default themes, This will enable users to directly select from a range of newly created custom themes. 

So basically what I am trying to say is that the person solving this issue will submit a set of new themes that the user doesn't need to customize himself. The user can just simply should be able to write `editorStyle: EditorStyle.monokai()`, here monokai being one of the themes that the person writing a PR will create?

Is that correct? And if yes then can I work on this issue simultaneously with the person who is assigned. 
I am experienced in UI/UX development and Flutter development and I feel confident about contributing on this issue.

annieappflowy:
@MayurSMahajan , yes, we’d like to offer users with a diverse set of themes. Multiple people can work on this issue together. You can just create a separate issue for the theme you’d like to create. I’ll then label it accordingly. If you need help with this task, please feel free to reach out to @LucasXu0 on Discord.

annieappflowy:
@ChiragKr04 , can you create a separate issue for the theme you’re building? I’ll then label it accordingly!. Thanks

ChiragKr04:
> @ChiragKr04 , can you create a separate issue for the theme you’re building? I’ll then label it accordingly!. Thanks

Sure
Should I pick on my own or you got any suggestions? 😃

ChiragKr04:
@MayurSMahajan You can pick any theme and If you pick one please let me know. 😄

MayurSMahajan:
Okay so I will think of themes and design them in Figma first. I'll tell their names to you so you can create issues. And soon I will submit PR with the code containing the Themes

MayurSMahajan:
> @ChiragKr04 , can you create a separate issue for the theme you’re building? I’ll then label it accordingly!. Thanks

[...]

PR #2670, opened 2023-05-31
[feat] Dynamically Load Themes in AppFlowy

a-wallen:
### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/44445638/2ce856ab-b5a8-4e67-8830-0c6c633446dd

Loads themes from a JSON file at runtime. 

TODO
- [x] Run build_runner in flowy_infra to pass CI
- [x] Fix `InvalidType` build_runner errors.
- [X] Provide default values or throw an error if `FlowyColorScheme` API changes and the old theme is incompatible with the new API.
- [x] Load the themes from a different directory (for example, `path_provider` `getApplicationDocumentsDirectory()`).
- [x] Allow users to drag and drop themes via the GUI (in the appearance settings menu).

https://github.com/AppFlowy-IO/AppFlowy/assets/44445638/f6d32036-73b3-4cdc-96ec-cef34b587331

- [x] Delete uploaded themes.

https://github.com/AppFlowy-IO/AppFlowy/assets/44445638/e4b050eb-b07a-421a-88d8-fea7978877e2

- [ ] ~Better prompt messages~
- [ ] ~Better error messages~
- [x] Fix all merge conflicts from the recent merge of `develop` into `main`
- [x] Refactor (code style)
- [ ] Widget/integration test for previous use cases.
- [ ] Users should be able to browse available JSON themes on a website.

Issues https://github.com/AppFlowy-IO/AppFlowy/issues/1070. It does not fix but makes it easier to experiment with new themes.

---

#### PR Checklist


frontend/appflowy_flutter/packages/flowy_infra/lib/utils/color_converter.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+import 'package:json_annotation/json_annotation.dart';
+
+class ColorConverter implements JsonConverter<Color, String> {
+  const ColorConverter();
+
+  static const Color fallback = Colors.transparent;
```
a-wallen, line 9: If the `FlowyColorScheme` API changes. The parser will provide `transparent` as the default value. 

Another way to fail silently is to pass over dynamic themes that can't be parsed. I.e., they won't be included if the app is updated since the theme is invalid. I thought this would be the best way to fail silently.

frontend/appflowy_flutter/packages/flowy_infra/lib/plugins/plugin.dart
```diff
@@ -0,0 +1,118 @@
...
+import 'package:flowy_infra/colorscheme/colorscheme.dart';
+import 'package:flowy_infra/theme.dart';
+import 'package:path/path.dart' as p;
+
+/// A class that encapsulates dynamically loaded plugins for AppFlowy.
+///
+/// This class can be modified to support loading node widget builders and other
+/// plugins that are dynamically loaded at runtime for the editor. For now,
+/// it only supports loading app themes.
+class FlowyPlugin {
```
appflowy, line 13: How about FlowyDynamicPlugin?

a-wallen:
Note: after adding `json_annotation`, `json_serializable`, and `build_runner` from appflowy's pubspec to flowy_infra's pubspec causes 3 different versions of the dart analyzer to download to the pub cache. Since the `InvalidType` error suggests that the `Locale` type is unresolved, the issue is probably coming from a broken analyzer. The current main `@71d118904871d5526c13479ca2e672d4dd81b6ea` downloads analyzer version `5.2.0` and `_fe_analyzer_shared-50.0`. Adding the same packages to flowy_infra downloads analyzer `5.13` and `_fe_analyzer_shared-61.0.0`.

a-wallen:
Working on testing some of the plugin infrastructure but `FlowyDynamicPlugin.decode` is making it a bit hard for me. I will resume tomorrow.

frontend/appflowy_flutter/lib/workspace/application/appearance.dart
```diff
@@ -181,7 +181,7 @@ class AppearanceSettingsState with _$AppearanceSettingsState {
     double menuOffset,
   ) {
     return AppearanceSettingsState(
-      appTheme: AppTheme.fromName(themeName),
+      appTheme: AppTheme.fallback,
```
Xazin, line 184: Is this correct? I would assume the `fromName`would use the fallback if not found.

Or do we not have the themeName in the initial state?

a-wallen: Initializing the theme is now async because it has to read files from disk. The initial state for the `AppearanceSettingsBloc` shouldn't be async, though. That's why I used the fallback here.

```
  static Future<AppTheme> fromName(String themeName) async {
    for (final theme in await themes) {
      if (theme.themeName == themeName) {
        return theme;
      }
    }
    throw ArgumentError('The theme $themeName does not exist.');
```

The full function to get the theme `fromName` also throws an error if the theme name isn't found. The theme should always be resolved, if it isn't it should be considered unexpected behavior. The user should never be presented with an option to select a theme that doesn't exist.

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/theme_upload/theme_upload_decoration.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:dotted_border/dotted_border.dart';
+import 'package:flutter/material.dart';
+
+import 'theme_upload_view.dart';
+
+class ThemeUploadDecoration extends StatelessWidget {
```
appflowy, line 6: Would it be better if we moved this widget to flowy_infra_ui and made it a configurable widget? Then it could be used for theme upload and file upload in the database.

frontend/appflowy_flutter/packages/flowy_infra/lib/plugins/bloc/dynamic_plugin_bloc.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:bloc/bloc.dart';
+import 'package:flowy_infra/plugins/service/plugin_service.dart';
+
+import '../../file_picker/file_picker_impl.dart';
+import 'dynamic_plugin_event.dart';
+import 'dynamic_plugin_state.dart';
+
+class DynamicPluginBloc extends Bloc<DynamicPluginEvent, DynamicPluginState> {
```
appflowy, line 8: I prefer placing the event and state in the Bloc file. However, it's also acceptable to separate them.

frontend/appflowy_flutter/packages/flowy_infra/lib/plugins/service/location_service.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:io';
+
+import 'package:path_provider/path_provider.dart';
+
+class PluginLocationService {
+  static Future<Directory> get fallback async =>
+      await getApplicationDocumentsDirectory();
```
appflowy, line 7: I think we should place the theme files in appflowy data folder. check out the `appFlowyDocumentDirectory`

```dart
Future<Directory> appFlowyDocumentDirectory() async {
  switch (integrationEnv()) {
    case IntegrationMode.develop:
      final Directory documentsDir = await getApplicationSupportDirectory()
        ..create();
      return Directory(path.join(documentsDir.path, 'data_dev')).create();
    case IntegrationMode.release:
      final Directory documentsDir = await getApplicationSupportDirectory();
      return Directory(path.join(documentsDir.path, 'data')).create();
    case IntegrationMode.test:
    case IntegrationMode.integrationTest:
      return Directory(path.join(Directory.current.path, '.sandbox'));
  }
}
```

appflowy (approved):

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/theme_upload/theme_upload_decoration.dart
a-wallen: I will hold off on this because I'd like to keep all of the theme upload widgets together. Some of the theme upload widgets use LocaleKeys which requires AppFlowy.

annieappflowy:
Can we aim to release the MVP by the end of next week?

a-wallen:
Sure 👍

a-wallen:
https://github.com/AppFlowy-IO/AppFlowy/assets/44445638/2b299db0-87b3-405b-926f-dcb07eb6dde6

frontend/appflowy_flutter/packages/flowy_infra/lib/plugins/service/models/flowy_dynamic_plugin.dart
```diff
@@ -0,0 +1,137 @@
...
+        .listSync()
+        .where((event) =>
+            event is File && p.basename(event.path).contains(lightExtension))
+        .first as File;
+
+    final dark = src
+        .listSync()
+        .where((event) =>
+            event is File && p.basename(event.path).contains(darkExtension))
+        .first as File;
```
Xazin, line 120: I'm looking at this and I can't help but think if there is a better way to do it.

The only way I could think of introduced ~15 more lines of code though, not sure it's worth it.

<details>
  <summary>~25 lines of code version</summary>

``` 
    final srcAsList = src.listSync();

    File? light;
    File? dark;

    for (int i = 0; i < srcAsList.length; i++) {
      if (light != null && dark != null) {
        break;
      }

      final event = srcAsList[i];
      if (event is File) {
        if (p.basename(event.path).contains(lightExtension)) {
          light = event;
        } else if (p.basename(event.path).contains(darkExtension)) {
          dark = event;
        }
      }
    }

    if (light == null || dark == null) {
      throw PluginCompilationException(
        'The theme plugin does not contain both a light and dark version: `theme.light.json` and `theme.dark.json`.',
      );
    }
``` 

</details>

a-wallen: At least this way, we're only listing the directory once. This is a good review comment. Let me try to expand on this.

a-wallen: Haha @Xazin, if we're code golfing it then you can try to beat this :)
```dart
    final files = src.listSync().whereType<File>().toList();

    File? light, dark;

    for (int i = 0; i < files.length || (light == null && dark == null); ++i) {
      final file = files[i];
      if (p.basename(file.path).contains(lightExtension)) {
        light = file;
      } else if (p.basename(file.path).contains(darkExtension)) {
        dark = file;
      }
    }
```

a-wallen: Either way we still get the benefit of only listing the directory once.

Xazin: Yeah that is indeed better. Semantically it's the same, but more readable.

PR #2692, opened 2023-06-03
refactor: updated at and created at

### LINKED ISSUE #2643: [Bug] auto-fill created time for existing rows
### Bug Description

After adding a "Created At" property, it only fills when a new row is added
image

### How to Reproduce

Add a Created At property to an existing database

### Expected Behavior

Pre-fill values for existing rows

### Operating System

Windows 11

### AppFlowy Version(s)

develop

### Screenshots

_No response_

### Additional Context

_No response_

appflowy:
refactor:
The updatedAt and createdAt  are reading the timestamp from the row instead of inserting a cell data. 

fix #2643 

#### PR Checklist

### DIFF
--- a/frontend/appflowy_tauri/src-tauri/Cargo.toml
+++ b/frontend/appflowy_tauri/src-tauri/Cargo.toml
@@ -34,12 +34,12 @@ default = ["custom-protocol"]
 custom-protocol = ["tauri/custom-protocol"]
 
 [patch.crates-io]
-collab = { git = "https://github.com/AppFlowy-IO/AppFlowy-Collab", rev = "b06be9" }
-collab-folder = { git = "https://github.com/AppFlowy-IO/AppFlowy-Collab", rev = "b06be9" }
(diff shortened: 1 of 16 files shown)


appflowy:
@zoli FYI, I've refactored the logic of updatedAt and createdAt

appflowy:
fix #2643

zoli:
@appflowy, About updated at and created at not being editable I thought we can enable freeze by default when the feature is completed but also let the user disable the freeze and edit them manually. Also, I wasn't sure prefilling the rows created before is a good idea. Maybe its better to use the collab row `created_at` field?

appflowy:
> @appflowy, About updated at and created at not being editable I thought we can enable freeze by default when the feature is completed but also let the user disable the freeze and edit them manually. Also, I wasn't sure prefilling the rows created before is a good idea. Maybe its better to use the collab row `created_at` field.

These two fields are designed to prevent manual editing. Yep, I use the row's created_at and updated_at define in collab crate.

Issue #2690, opened 2023-06-03
[Bug] Please remove Https:// prefix if already added.


ChiragPaneliya:
### Bug Description

When we add New task with url porperty in board document and clicked on url icon which opens url to browser it's adding https:// prefix to url even if there is already https:// prefix available.

https://github.com/AppFlowy-IO/AppFlowy/assets/51508126/0715388f-2625-4df8-a415-be92ce604589

### How to Reproduce

1. Create New Board Document.
2. Create Task and Add It's property as url and add your url with https:// prefix.
3. Click on Url icon which is available right side of url property.

### Expected Behavior

if url already contains https:// prefix then don't add (https://) instad of just open url.

### Operating System

Windows 10

### AppFlowy Version(s)

develop

annieappflowy:
I've assigned this issue to you, @ChiragPaneliya

PR #2719, opened 2023-06-06
fix:http:// prefix issue fix

ChiragPaneliya:
---

#2690 

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/url_cell/url_cell.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/url_cell/url_cell.dart
@@ -299,7 +299,7 @@ class _VisitURLAccessoryState extends State<_VisitURLAccessory>
   void onTap() {
     var content =
         widget.cellContext.getCellData(loadIfNotExist: false)?.content ?? "";
-    if (!content.contains('http://')) {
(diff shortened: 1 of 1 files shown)


frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/url_cell/url_cell.dart
```diff
@@ -299,7 +299,7 @@ class _VisitURLAccessoryState extends State<_VisitURLAccessory>
   void onTap() {
     var content =
         widget.cellContext.getCellData(loadIfNotExist: false)?.content ?? "";
-    if (!content.contains('http://')) {
+    if (!content.startsWith('http://') && !content.startsWith('https://')) {
```
a-wallen, line 302: It would be better for performance if you could do a single pass to check that the string is `http` and not `https`. You could do that with a regular expression or your function. Right now, this is O(2n). Not that it makes a huge difference.

a-wallen (approved):
Approving, but I think that the check could be written differently. @ChiragPaneliya, let me know if you want to try a different approach. Willing to merge as is.

a-wallen:
@ChiragPaneliya I think that the `Commit messages lint` is failing because you're missing a space after the semicolon in your commit message. `git commit --amend` should allow you to rewrite, then you'll have to force push `git push --force`

Issue #2633, opened 2023-05-26
[Bug] New grid row button is out of place


richardshiue:
### Bug Description

The new row button looks out of place. There is a weird amount of blank space between the last row of the grid and the new row button.

### How to Reproduce

Open up a grid and scroll

### Expected Behavior

It should either be attached to the bottom of the row list, or just above the "Count" bar.

### Operating System

Fedora 38

### AppFlowy Version(s)

latest develop

### Screenshots

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/5dbbc749-d8ed-4c94-8316-0785a0909ad2

appflowy:
This is caused by using the SliverFillRemaining. 

> SliverFillRemaining: It sizes its child to fill the remaining space in the viewport. Typically, this is used for a flexible space where you don't have a specific size in mind. It can be quite useful for content that should take up the remaining available space on the screen.

It works fine if we use the SliverAnimatedList. But it does not support moving the row.

Xazin:
I'll take a look at this, we don't need to use SliverFillRemaining, so if that's the primary issue we can refactor that part.

PR #2723, opened 2023-06-06
fix: add new row issue

Xazin:
Closes: #2633 

### Feature Preview

---

#### PR Checklist

CC @LucasXu0 @appflowy

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
@@ -2,7 +2,6 @@ import 'package:appflowy/generated/locale_keys.g.dart';
 import 'package:appflowy/plugins/database_view/widgets/row/cell_builder.dart';
 import 'package:easy_localization/easy_localization.dart';
 import 'package:flowy_infra_ui/flowy_infra_ui_web.dart';
-import 'package:flowy_infra_ui/style_widget/scrolling/styled_list.dart';
 import 'package:flowy_infra_ui/style_widget/scrolling/styled_scroll_bar.dart';
 import 'package:flowy_infra_ui/style_widget/scrolling/styled_scrollview.dart';
 import 'package:flowy_infra_ui/style_widget/text.dart';
@@ -122,10 +121,9 @@ class _FlowyGridState extends State<FlowyGrid> {
         final contentWidth = GridLayout.headerWidth(state.fields.value);
         final child = _wrapScrollView(
           contentWidth,
-          [
-            const _GridRows(),
-            const _GridFooter(),
-          ],
+          _GridRows(
+            verticalScrollController: _scrollController.verticalController,
+          ),
         );
 
         return Column(
@@ -144,20 +142,11 @@ class _FlowyGridState extends State<FlowyGrid> {
 
   Widget _wrapScrollView(
     double contentWidth,
-    List<Widget> slivers,
+    Widget child,
   ) {
-    final verticalScrollView = ScrollConfiguration(
-      behavior: const ScrollBehavior().copyWith(scrollbars: false),
-      child: CustomScrollView(
-        physics: StyledScrollPhysics(),
-        controller: _scrollController.verticalController,
-        slivers: slivers,
-      ),
-    );
-
     final sizedVerticalScrollView = SizedBox(
       width: contentWidth,
-      child: verticalScrollView,
+      child: child,
     );
 
     final horizontalScrollView = StyledSingleChildScrollView(
@@ -170,6 +159,7 @@ class _FlowyGridState extends State<FlowyGrid> {
       axis: Axis.vertical,
       controller: _scrollController.verticalController,
       barSize: GridSize.scrollBarSize,
+      autoHideScrollbar: false,
       child: horizontalScrollView,
     );
   }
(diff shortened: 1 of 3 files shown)


Xazin:
> 

I'll take a look.

As for reordering, the list isn't shrinkwrapped if that is what you mean, render-wise.

Xazin:
@appflowy Forgot to remove the dispose since it would be handled from further up after my changes. The error should be gone.

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
```diff
@@ -221,28 +213,29 @@ class _GridRowsState extends State<_GridRows> {
...
             final rowInfos = context.watch<GridBloc>().state.rowInfos;
 
-            return SliverFillRemaining(
+            return ScrollConfiguration(
+              behavior:
+                  ScrollConfiguration.of(context).copyWith(scrollbars: false),
               child: ReorderableListView.builder(
+                /// This is a workaround related to
+                /// https://github.com/flutter/flutter/issues/25652
+                cacheExtent: 5000,
```
a-wallen, line 232: Why'd you choose 5000?

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
```diff
@@ -256,16 +249,21 @@ class _GridRowsState extends State<_GridRows> {
...
-                    index: index,
-                    isSortEnabled: sortState.sortInfos.isNotEmpty,
-                    isFilterEnabled: filterState.filters.isNotEmpty,
-                  );
+                itemCount: rowInfos.length + 1,
+                itemBuilder: (context, index) {
+                  if (index < rowInfos.length) {
+                    final rowInfo = rowInfos[index];
+
+                    return _renderRow(
```
a-wallen, line 257: If this returns a widget, shouldn't we make another widget instead of a function that returns a widget?

a-wallen (approved):

a-wallen:
Code LGTM, but Nathan has valid concerns that should be addressed.

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
Xazin: That is also my preference. I dislike returning widgets inside methods and functions.

But I didn't want to change it in already working code, since in this use-case it doesn't really have a demerit other than preference..

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
Xazin: Honestly I didn't even want to use cacheExtent, I think it's a really bad solution, but I couldn't find any other good ways to eliminate the inconsistent scrollbar behavior without spending too much time.

The number 5000 is half arbitrary, but it's approximately an empty cell * 110 (+ a bit).

I should probably add a todo here, so we can look into fixing it at a later time. The inconsistency occurs because a ListView isn't rendered all at once, so the Scrollbar needs to know the extent of each cell, and believes them to be consistently the same. That means once a cell with a different extent (height) renders, the scrollbar is confused and thus changes position (correctly) and becomes smaller (correctly).

I have some solutions in mind, but let's discuss those at a later time.

Xazin:
> Using the reorder list, IIRC, all the rows need to be built when opening the grid. This might pose a problem as the number of rows in the grid increases.
> 
> BTW, this error show up when I close the database. image

The ReorderableListView.builder only builds the items when needed, also what I meant by not shrinkwrapping.

> See also: [ReorderableListView.builder], which allows you to build a reorderable list where the items are built as needed when scrolling the list.

I'm not using the `reorderables` package here. 
https://api.flutter.dev/flutter/material/ReorderableListView/ReorderableListView.builder.html

appflowy:
> > Using the reorder list, IIRC, all the rows need to be built when opening the grid. This might pose a problem as the number of rows in the grid increases.
> > BTW, this error show up when I close the database. image
> 
> The ReorderableListView.builder only builds the items when needed, also what I meant by not shrinkwrapping.
> 
> > See also: [ReorderableListView.builder], which allows you to build a reorderable list where the items are built as needed when scrolling the list.
> 
> I'm not using the `reorderables` package here. https://api.flutter.dev/flutter/material/ReorderableListView/ReorderableListView.builder.html

Ah, they look the same. The ListView will load rows within the cacheExtent region. Having it set to 5000 causes a large number of rows to load.

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/grid_page.dart
```diff
@@ -221,28 +213,30 @@ class _GridRowsState extends State<_GridRows> {
...
-            return SliverFillRemaining(
+            return ScrollConfiguration(
+              behavior:
+                  ScrollConfiguration.of(context).copyWith(scrollbars: false),
               child: ReorderableListView.builder(
+                /// TODO(Xazin): Resolve inconsistent scrollbar behavior
+                ///  This is a workaround related to
+                ///  https://github.com/flutter/flutter/issues/25652
+                cacheExtent: 5000,
                 key: _key,
```
LucasXu0, line 234: This key is used for animation and it should be associated with the `SliverAnimatedListState`.

You changed `SliverFillRemaining` to `ReorderableListView`, so the animation will no longer take effect.

LucasXu0: I suppose the logic of animation can be removed now.

LucasXu0:
@Xazin FYI, I removed the unused animation key and fixed the `setState` error bug.

PR #2802, opened 2023-06-15
IDE i18n workflow improvement / i18n CLI

NilsJacobsen:
> Let me know if this PR needs changes

## Feature Preview

Appflowy already uses `inlang` to get a [web editor](https://inlang.com/editor/github.com/AppFlowy-IO/AppFlowy) to contribute translations. I updated the plugin links in the config, cause the old ones where depricated (inlang moved plugins to their monorepo). This allows you to use the [IDE extension](https://inlang.com/documentation/apps/ide-extension) for `vscode` with no additional configuration. It's a simple helper to extract messages and have inline annotations. 

You also get ootb the [inlangCLI](https://inlang.com/documentation/apps/inlang-cli) to lint & machine translate the i18n strings. 

I also added the extension to the workspace recommendations so everybody can use it right away and benefit from this shortcut.

#### Extract Messages (translations)

Extract Messages (translations) via the `Inlang: Extract Message` code action.

Extract Messages

#### Inline Annotations

See translations directly in your code. No more back-and-forth looking into the translation files themselves.

Inline Annotations

### DIFF
--- a/frontend/appflowy_flutter/lib/startup/tasks/supabase_task.dart
+++ b/frontend/appflowy_flutter/lib/startup/tasks/supabase_task.dart
@@ -1,9 +1,14 @@
 import 'package:appflowy/env/env.dart';
+import 'package:appflowy/workspace/application/settings/application_data_storage.dart';
+import 'package:flutter/foundation.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
+import 'package:hive_flutter/hive_flutter.dart';
+import 'package:path/path.dart' as p;
 
 import '../startup.dart';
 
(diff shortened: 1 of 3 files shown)


qinluhe:
> Ignore the last 3 commits, I was just trying out a different syntax.

Hi, I understand that the purpose of this PR is to add type hints for i18n, which is already available in the main branch, and the command in tauri is pnpm sync:i18n

NilsJacobsen:
@qinluhe  I cleaned up the PR and removed the CLI. Updated the inlang config to the newest versions in a follow-up PR.

PR #2900, opened 2023-06-25
zh-CN update

LainHE:
I just updated the zh-CN.json based on en.json. After replace the original file, the app has able to run succeefully.

It seems still a lot of entries<sub>(Is that word correct?)</sub> are not translated[^1]. Is that even en.json are not coverd all entries?

No matter what, it‘s still increased about 200+ entries in Simplified Chinese translation. Hopes that would helpful.
[^1]:


LucasXu0:
@LainHE Yes, some keys are not covered. I will add them soon.

PR #2907, opened 2023-06-28
feat: the application remembers being resized

forecaster-cyber:
Hi, hope this is what you meant...


frontend/appflowy_flutter/lib/startup/tasks/windows.dart
```diff
@@ -18,6 +19,7 @@ class InitAppWindowTask extends LaunchTask with WindowListener {
 
   @override
   Future<void> initialize(LaunchContext context) async {
+    final SharedPreferences prefs = await SharedPreferences.getInstance();
```
LucasXu0, line 22: Please don't use `SharedPreferences` directly.

There's a already existing window manager. Use it instead.

forecaster-cyber: I use it because I create a variable using it that goes true if the window is maximized...
What do you mean using the window manager?

LucasXu0: I pushed the code, please take a look.

frontend/appflowy_flutter/lib/startup/tasks/app_window_size_manager.dart
```diff
@@ -13,7 +13,10 @@ class WindowSizeManager {
   static const width = 'width';
   static const height = 'height';
 
-  Future<void> saveSize(Size size) async {
+  static const dx = 'dx';
+  static const dy = 'dy';
```
Xazin, line 17:
```suggestion
  static const String dx = 'dx';
  static const String dy = 'dy';
```

richardshiue:
From my end, the maximized state still isn't being saved.

LucasXu0:
@richardshiue Can you help me check if the function `onWindowMaximize` is called correctly on Linux? It works fine on my macOS.

richardshiue:
> @richardshiue Can you help me check if the function `onWindowMaximize` is called correctly on Linux? It works fine on my macOS.

Of course. The function is called when I click the maximize (breakpoint). Though, when I restart the program in this state, the size of the window is the same but the window isn't maximized. The second button is supposed to be a restore icon and I should be able to resize the window by placing the cursor at the edge of the window.

image

LucasXu0:
@richardshiue Previously, I planed to save the maximize status instead of the window size. 

```dart
```

richardshiue:
> @richardshiue Previously, I planned to save the maximized status instead of the window size. However, it appears that the windowsManager only supports the maximize function on Windows.
> 
> ```dart
>   /// Maximizes the window. `vertically` simulates aero snap, only works on Windows
>   Future<void> maximize({bool vertically = false}) async {
>     final Map<String, dynamic> arguments = {
>       'vertically': vertically,
>     };
>     await _channel.invokeMethod('maximize', arguments);
>   }
> ```

Ahh got it. LGTM then.

Xazin:
What's the plan here @LucasXu0 @richardshiue, is anything missing or can we rebase and merge?

richardshiue:
I don't mind getting this in since saving window position is an improvement. The original issue shouldn't be closed though, since the application still can't remember being maximized.

That said, I will play around with this when I have the time. If I'm not mistaken, Kwin (not sure about other WMs and OSs) remembers window position and size.

LucasXu0 (approved):

PR #2915, opened 2023-06-30
fix: trigger option selection on tap down

### LINKED ISSUE #2832: [Bug] click twice to select an option
### Bug Description

When I select an entry within the Kanban board and then the priority property (select), I have to click twice on an option until it is applied (only for the first time). (As if the first click activates the options window and then the second click selects the option.) 

### How to Reproduce

Add a select property to a Kanban board
Add an option
Add another card
Select the option in the select property

### Expected Behavior

Click once to select an option

### Operating System

Windows 11

### AppFlowy Version(s)

v0.2.2

### Screenshots

_No response_

### Additional Context

_No response_

vedant-pandey:
Closes #2832

This appears to be related to an [issue in Flutter](https://github.com/flutter/flutter/issues/53797) wherein onTapUp is not triggered correctly leading to onTap not firing. As a possible fix, I have moved this to use onTapDown.

### Feature Preview

---

#### PR Checklist

- [x] This PR only contains semantic changes. 

https://github.com/AppFlowy-IO/AppFlowy/assets/25723331/c052ea36-ae4b-406e-9da6-76139f9ea458


richardshiue:
Hello @vedant-pandey, thanks for the contribution. It seems to me the linked issue is describing a different scenario from the one here. Is there perhaps an alternative solution? Listening to `onTapDown` instead of `onTap` isn't a user interaction that is usually expected.

vedant-pandey:
Hi @richardshiue, thank you. The linked flutter issue is slightly different in what seems to be happening, but it mentions [Inkwell.onTap not working](https://github.com/flutter/flutter/issues/53797#issuecomment-1373206884) which is what seems to be happening here too. Although, I do agree onTapDown isn't accurate to be used here, but found this to be most suitable workaround. Let me know if you think we should stick to `onTap`

annieappflowy:
@richardshiue , is this PR ready to merge? What is blocking it from closing?

richardshiue:
It fixes the bug but with side effects. I don't mind getting this is in, but it would be better if there is a TODO added saying this is a hack.

annieappflowy:
Hi @richardshiue , please help expediting this issue

PR #2921, opened 2023-07-01
feat: add duplicate in context menu

### LINKED ISSUE #2833: [FR] Add Duplicate to the card action menu
### Description

To duplicate a canbanboard item it would be handy if I didn't have to open it first, but could select it from the context menu (:), where I can also delete it so far.
image

### Impact

Kanban users

### Additional Context

_No response_

vedant-pandey:
### Feature Preview

Fixes #2833

---

#### PR Checklist

- [x] this PR only contains semantic changes.


vedant-pandey:
One important note about this PR- on duplication the group does not get updated, and it appears as if the duplication hasn't happened. This happens due to the code [here in view_editor](https://github.com/AppFlowy-IO/AppFlowy/blob/main/frontend/rust-lib/flowy-database2/src/services/database_view/view_editor.rs#L221-L228) and because RowActions is stateless widget and does not have a group id to forward. 
Correct me if I'm wrong but my guess is this should be handled from the rust side, since upon row creation, the group should always be notified?
If someone from the team can provide a pointer as to how this change should be handled then I'll add it in the PR

richardshiue:
Hello @vedant-pandey, thanks for tackling this issue. The `RowService` on the Flutter side does, in fact allow duplicating a row while also passing a `groupId` along. So you could potentially add an optional `groupId` to the constructors of `RowActionSheetBloc` and also `RowActions`, and when telling the row service to send the event in the bloc, include the `groupId`. Let's see if that works.

vedant-pandey:
@richardshiue updated with suggested changes

richardshiue:
Great progress, any possibility of not entering edit mode upon row duplication?

vedant-pandey:
@richardshiue not sure about it, since duplicate from popup menu behaves similarly. Seems to be possible if we can differentiate event as `DidUpdateViewRows` + `duplicate` + duplicate happening from context menu and not popup menu. On the contrary, disabling edit mode everywhere might be simpler but I believe not desired

richardshiue:
Okay sure, we can open up a separate issue to work on that. Things look good from my end. I'm curious about your opinions on this @Xazin @a-wallen

frontend/appflowy_flutter/lib/plugins/database_view/board/presentation/board_page.dart
```diff
@@ -295,6 +298,7 @@ class _BoardContentState extends State<BoardContent> {
         rowCache: rowCache,
         cardData: groupData.group.groupId,
         groupingFieldId: groupItem.fieldInfo.id,
+        groupId: groupData.group.groupId,
```
Xazin, line 301: This looks a bit weird to me:
```
cardData: groupData.group.groupId,
groupId: groupData.group.groupId,
```

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/row/action.dart
```diff
@@ -110,7 +116,7 @@ extension _RowActionExtension on _RowAction {
   bool enable() {
     switch (this) {
       case _RowAction.duplicate:
-        return false;
+        return true;
       case _RowAction.delete:
         return true;
```
Xazin, line 121:
```suggestion
      case _RowAction.delete:
        return true;
```

Xazin: This didn't render that nicely, basically combines the two cases.

frontend/appflowy_flutter/lib/plugins/database_view/board/presentation/board_page.dart
richardshiue: Great catch, thanks a bunch. Let's just use the `cardData` since it's there. That field was originally already meant to carry custom information.

vedant-pandey: Hi @Xazin @richardshiue this was intentional as the cardData field seems to be used for different purpose [here in calendar_day](https://github.com/AppFlowy-IO/AppFlowy/blob/main/frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_day.dart#L309) to avoid complications I created a new field. Please let me know if my understanding isn't correct

Xazin: For now, we can leave it as is.

LucasXu0:
@vedant-pandey This PR contains not only semantic changes. Please include a test to cover it.

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/row/action.dart
```diff
@@ -16,16 +16,22 @@ import '../../layout/sizes.dart';
 class RowActions extends StatelessWidget {
   final String viewId;
   final RowId rowId;
+  final String? groupId;
   const RowActions({
     required this.viewId,
     required this.rowId,
+    this.groupId,
```
a-wallen, line 23: Wondering if this should be required. If it isn't required, does that mean the card can't be duplicated?

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/row/action.dart
```diff
@@ -16,16 +16,22 @@ import '../../layout/sizes.dart';
...
   @override
   Widget build(BuildContext context) {
     return BlocProvider(
-      create: (context) => RowActionSheetBloc(viewId: viewId, rowId: rowId),
+      create: (context) => RowActionSheetBloc(
+        viewId: viewId,
+        rowId: rowId,
+        groupId: groupId,
+      ),
       child: BlocBuilder<RowActionSheetBloc, RowActionSheetState>(
```
a-wallen, line 35: I would recommend moving the BlocProvider to the parent widget. When both the BlocProvider and the BlocBuilder are in the same widget, it makes it harder to test since you cannot mock the BlocProvider's values.

a-wallen: Adding this will make the unit tests that you need to land this easier 👍

frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/row/action.dart
richardshiue: `RowCard` is also used in calendar which does not have groups.

a-wallen: Makes sense to me. It might be worth it to keep the `RowActions` API this way and implement two factory constructors for `RowCard` to prevent API misuse.

richardshiue: Got it

vedant-pandey:
Hi @LucasXu0 as discussed the tests seem to be failing for kanban board specifically, while others are working as expected. @a-wallen thank you, I will raise the new changes soon

Xazin:
Hey @vedant-pandey how is it going?

If you need help with adding the tests feel free to hit us up on Discord 👍

LucasXu0:
I have added the integration for deleting duplicates. The CI failed because of the cache.FYI, @Xazin

PR #3051, opened 2023-07-24
fix: sorting rows with cells that are empty

richardshiue:
When there are unchecked, initialized checkbox cells with empty cells, the empty cells are automatically sent to the end of the grid. Even when I add a second sort, say, by text, the empty cell will still be at the end even if it should've shifted up a bit.

There is quite a bit of extra code here, but unless we initialize every single checkbox cell, some extra logic is necessary.

Before:

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/a44f91cc-33f6-49d7-b4d5-d988118860ee

at 0:25, AB and AD are true, AE is false, and the others are empty. When I sort by checkbox, the natural order should be preserved instead of AE at the bottom. Worse still when I sort by Name, the order doesn't change

### Feature Preview

---

#### PR Checklist


frontend/rust-lib/flowy-database2/tests/database/sort_test/multi_sort_test.rs
```diff
@@ -5,46 +5,44 @@ use crate::database::sort_test::script::DatabaseSortTest;
...
-      field_id: checkbox_field.id.clone(),
-      orders: vec!["Yes", "No", "Yes", "No", "No", "Yes"],
+      orders: vec!["A", "", "C", "DA", "AE", "AE", "CB"],
     },
-  ];
-  test.run_scripts(scripts).await;
-
-  let scripts = vec![
+    // Insert checkbox sort
     InsertSort {
```
appflowy, line 22: I think it would be better to assert the order first that can help the other developers understand how the order changes by inserting a sort.

richardshiue: I combined the two scripts into one so the lines above assert the order. Or maybe I misunderstood?

richardshiue: I combined the two scripts into one so the lines above assert the order. Or maybe I misunderstood?

PR #3061, opened 2023-07-25
Support document format text color and bg color

qinluhe:
### Feature Preview

- [ ] Support document format text color and bg color
image

---

#### PR Checklist


frontend/appflowy_tauri/src/appflowy_app/components/document/Node/index.tsx
```diff
@@ -90,7 +90,7 @@ function NodeComponent({ id, ...props }: { id: string } & React.HTMLAttributes<H
         {...props}
         ref={ref}
         data-block-id={node.id}
-        className={`pt-[0.5px] ${className}`}
+        className={`${className}`}
```
Xazin, line 93:
```suggestion
        className={className}
```

frontend/appflowy_tauri/src/appflowy_app/components/document/TextActionMenu/menu/BgColorPicker.tsx
```diff
@@ -0,0 +1,101 @@
...
+      {
+        name: t('colors.custom'),
+        key: 'custom',
+        color: 'transparent',
+      },
+      {
+        key: 'gray',
+        name: t('colors.gray'),
+        color: '#78909c',
+      },
```
Xazin, line 38: Currently, all of these colors exist multiple times.

I think it would be nice to have them defined somewhere such that any type of color picker can use them.

qinluhe: > Currently, all of these colors exist multiple times.

I don't quite understand, where exactly?

Xazin: Not really sure where the best location would be since I'm not too into the architecture and abstraction of the tauri app.

But something like this:
```
import { TFunction } from "react-i18next";

function getColorOptions(TFunction t)  {
  return [
    {
      name: t('colors.default'),
      key: 'default',
      color: 'transparent',
    },
    ...
  ];
}
```

qinluhe: In fact, this entry file is used for configuration, and the real component is ColorPicker
After using useMemo, if `t` has not changed, this object is cached

Xazin: Ah okay, it's just that they're the same, I was expecting the color options would be 1:1 even in the future. (Eg. if a user adds a custom color to their palette, it should be possible to pick that for both background and text.).

qinluhe: The background color and text color have different values now, so I separate their entry files

PR #3073, opened 2023-07-28
fix: height for calendar view in document

### LINKED ISSUE #2806: [Bug] inline cal area too small
### Bug Description

image

### How to Reproduce

Create a calendar inside a document

### Expected Behavior

enough space for viewing a month of events

### Operating System

Windows 11

### AppFlowy Version(s)

v0.2.1

### Screenshots

_No response_

### Additional Context

_No response_

richardshiue:
title says all, resolves #2806 

image

### Feature Preview

---

#### PR Checklist


Xazin:
How does the calendar reference and calendar size the height of each cell when there's multiple events?

I'm curious to know if the whole row resizes to fit width, or if we could make it so? _(Within some boundaries so that it's not infinite)._

richardshiue:
Hi @Xazin, the calendar component we're using stretches out horizontally and then based on the aspect ratio provided (0.6 in this case), calculate the height of each cell. If `useAvailableSpace` is specified, the cell aspect ratio is calculated using the constraint's maxHeight. Individual cells cannot dynamically change its height based on its contents, only by resizing the overall calendar component. Therefore, when there are multiple events in a cell, I'm forced to make them live in a listview and let them scroll.

> I'm curious to know if the whole row resizes to fit width

I'm not sure if I got your question. As I understand it, the calendar already fits the width and the cells are distributed along that x axis

Xazin:
> Hi @Xazin, the calendar component we're using stretches out horizontally and then based on the aspect ratio provided (0.6 in this case), calculate the height of each cell. If `useAvailableSpace` is specified, the cell aspect ratio is calculated using the constraint's maxHeight. Individual cells cannot dynamically change its height based on its contents, only by resizing the overall calendar component. Therefore, when there are multiple events in a cell, I'm forced to make them live in a listview and let them scroll.
> 
> 
> 
> > I'm curious to know if the whole row resizes to fit width
> 
> 
> 
> I'm not sure if I got your question. As I understand it, the calendar already fits the width and the cells are distributed along that x axis
> 
> 
> 
> 

Ah yes, meant height there.

frontend/appflowy_flutter/lib/plugins/database_view/calendar/presentation/calendar_page.dart
```diff
@@ -175,19 +168,27 @@ class _CalendarPageState extends State<CalendarPage> {
...
-    return Expanded(
-      child: Padding(
-        padding: GridSize.contentInsets,
-        child: MonthView(
+  Widget _buildCalendar(
+    BuildContext context,
+    EventController eventController,
+    int firstDayOfWeek,
+  ) {
+    final isInDocument = context.read<DocumentBloc?>() != null;
```
Xazin, line 176: I'm not sure if I like this. I think I'd prefer a field in the `CalendarPage` eg. `bool shrinkWrap` or `bool fixedHeight`.

richardshiue: Good idea, how would you do this? I still need the build context so it has to be in the build method. Or do you mean passing in the bool in the constructor? I don't think that's a bad idea since other database views could also benefit from this, but I don't want to make the code for DatabaseViewWidget complicated just for the calendar.

Xazin: I see what you mean. Let me check what other options we have.

LucasXu0: @Xazin @richardshiue Any updates? If not, I would prefer a solution that involves adding a parameter called 'shrinkWrap'.

Xazin: I think it's fine to default `shrinkWrap` to false, and then in the relevant database block builder simply set it to true.

LucasXu0: Sounds good to me. @richardshiue What do you think of this solution?

richardshiue: I'm a bit confused. Do you mean we pass shrinkwrap through constructors from database_view_block_component down to calendar_page? I wanna make sure before I commit

LucasXu0: Yes. Use shrinkWrap instead of the isInDocumrnt to identify whether the database view needs more space or not.

Xazin (approved):

Issue #3108, opened 2023-08-03
[Bug] bugs when sorting in multi-sort


richardshiue:
### Bug Description

Currently, when a field is being sorted, the following happens following a previous PR https://github.com/AppFlowy-IO/AppFlowy/pull/2953 . For the sake of explanation, we'll assume Ascending sort for now.

- If the cell is empty (as in it isn't initialized), it is immediately moved to the bottom.
- If the cell is initialized, but the content is empty (such as a being an empty string, or the timestamp of a date field is None), it is then moved to the bottom.

This works well for the most part: empty cells are moved to the bottom and not sorted. However, if you look at the order of these rows closely, it's not quite right:

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/b428d4ef-2dc5-461f-a8c0-6c8a3b6aaf52

By typing text into the cell and then clearing it, I initialize it and then clear the value. Following the checkbox, that row that was previously below another empty row, suddenly rises above it after the sort. This is unexpected because natural order should be preserved.

The following example illustrates what happens during multi-sort

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/8314b6f3-1cf2-4943-89f7-603af8b96698

### How to Reproduce

Please refer to the description for what exactly is happening.

### Expected Behavior

The sort results should be intuitive and reasonable to the user.

### Operating System

Fedora Linux 38

### AppFlowy Version(s)

0.2.8

PR #3109, opened 2023-08-03
feat: selectOption sort + fix logic in edge cases

richardshiue:
This PR fixes the sort logic in multi-sort scenarios.

The following logic is in place, hopefully it makes sense. Here `cell` means the cell of the row that is under the field being used to sort the grid and sort condition is either ascending or descending

- 2 uninitialized cells should preserve natural order
- uninitialized cell and initialized cell that is regarded as empty (empty text string, empty date timestamp, empty list of select ids, etc) should preserve natural order 
- BUT if the field type is checkbox then the uninitialized cell must be regarded as UNCHECKED and compared to the initialized cell according to the sort condition
- a pair of uninitialized and initialized cell that is not empty should move the initialized cell to the top regardless of sort condition
- 2 initialized that are both regarded as empty should preserve natural order
- 2 initialized cells that are both not empty should be compared against each other according to the field type and sort condition
- 1 initialized cell that is regarded as empty and another initialized cell that is not empty should move the non-empty cell to the top regardless of sort condition

### Feature Preview

fix: #3108

#### PR Checklist


frontend/rust-lib/flowy-database2/src/services/field/type_options/checklist_type_option/checklist.rs
```diff
@@ -203,8 +205,17 @@ impl TypeOptionCellDataCompare for ChecklistTypeOption {
     }
   }
 
-  fn exempt_from_cmp(&self, cell_data: &<Self as TypeOption>::CellData) -> bool {
-    cell_data.selected_option_ids.is_empty()
+  fn apply_cmp_with_uninitialized(
+    &self,
+    cell_data: Option<&<Self as TypeOption>::CellData>,
+    other_cell_data: Option<&<Self as TypeOption>::CellData>,
+    _sort_conditon: SortCondition,
```
LucasXu0, line 212:
```suggestion
    _sort_condition: SortCondition,
```

frontend/rust-lib/flowy-database2/src/services/field/type_options/date_type_option/date_type_option.rs
```diff
@@ -306,16 +307,32 @@ impl TypeOptionCellDataCompare for DateTypeOption {
...
+      (Some(left), Some(right)) => {
+        let order = left.cmp(&right);
+        match sort_condition {
+          SortCondition::Ascending => order,
+          SortCondition::Descending => order.reverse(),
+        }
+      },
       (Some(_), None) => Ordering::Greater,
       (None, Some(_)) => Ordering::Less,
       (None, None) => default_order(),
```
LucasXu0, line 322: There're many similar codes like this one.

```rs
match (a, b) => {
 some, some,
 some, none,
 none, some,
 none, none,
}
```

Can we use the built-in `cmp` function instead, and try to reduce the boilerplate code?

frontend/rust-lib/flowy-database2/src/services/field/type_options/checklist_type_option/checklist.rs
richardshiue: keen eye

frontend/rust-lib/flowy-database2/src/services/field/type_options/date_type_option/date_type_option.rs
richardshiue: In some cases, definitely. But most cases it's a bit difficult since the sort condition only applies to a single arm (some, some). What I can do is something like this

```rust
match (a, b) => {
  (some, some) => ...,
  _ => a.cmp(&b)
}
```

Or do you an alternative solution?

LucasXu0: I don't have a good solution for the case you mentioned. If that's the case, just leave it as is.

richardshiue: Sounds good, resolving this thread for now!

frontend/rust-lib/flowy-database2/src/services/field/type_options/date_type_option/date_type_option.rs
```diff
@@ -306,16 +307,32 @@ impl TypeOptionCellDataCompare for DateTypeOption {
...
     }
   }
 
-  fn exempt_from_cmp(&self, cell_data: &<Self as TypeOption>::CellData) -> bool {
-    cell_data.timestamp.is_none()
+  fn apply_cmp_with_uninitialized(
+    &self,
+    cell_data: Option<&<Self as TypeOption>::CellData>,
+    other_cell_data: Option<&<Self as TypeOption>::CellData>,
+    _sort_conditon: SortCondition,
```
appflowy, line 330: typo here: condition

frontend/rust-lib/flowy-database2/src/services/field/type_options/selection_type_option/multi_select_type_option.rs
```diff
@@ -140,31 +141,51 @@ impl TypeOptionCellDataCompare for MultiSelectTypeOption {
...
+    match cell_data.len().cmp(&other_cell_data.len()) {
+      Ordering::Equal => {
+        for (left_id, right_id) in cell_data.iter().zip(other_cell_data.iter()) {
+          let left = self.options.iter().find(|option| &option.id == left_id);
+          let right = self.options.iter().find(|option| &option.id == right_id);
+
+          let order = match (left, right) {
+            (None, None) => Ordering::Equal,
+            (None, Some(_)) => Ordering::Greater,
+            (Some(_), None) => Ordering::Less,
```
appflowy, line 155: If the left side has a value, then the comparison would be: (Some(_), None) => Ordering::Greater

frontend/rust-lib/flowy-database2/src/services/field/type_options/selection_type_option/multi_select_type_option.rs
```diff
@@ -140,31 +141,51 @@ impl TypeOptionCellDataCompare for MultiSelectTypeOption {
...
+            (Some(left_option), Some(right_option)) => {
+              let name_order = left_option.name.cmp(&right_option.name);
+              match sort_condition {
```
appflowy, line 158: I've seen this code in many places. How about we create a functions for it?
```rust
match sort_condition {
       SortCondition::Ascending => name_order,
       SortCondition::Descending => name_order.reverse(),
}
```

frontend/rust-lib/flowy-database2/src/services/field/type_options/selection_type_option/multi_select_type_option.rs
richardshiue: Oh. I wanted to move the non-empty cells to the front and the empty ones to the back. That's why I write Ordering::Less

frontend/rust-lib/flowy-database2/src/services/field/type_options/selection_type_option/multi_select_type_option.rs
```diff
@@ -140,31 +141,51 @@ impl TypeOptionCellDataCompare for MultiSelectTypeOption {
...
+        SortCondition::Ascending => order,
+        SortCondition::Descending => order.reverse(),
+      },
     }
-    default_order()
   }
 
-  fn exempt_from_cmp(&self, cell_data: &<Self as TypeOption>::CellData) -> bool {
-    cell_data.is_empty()
+  fn apply_cmp_with_uninitialized(
```
appflowy, line 178: Using the default trait implementation reduces code repetition, and you only need to implement it when necessary

frontend/rust-lib/flowy-database2/src/services/field/type_options/selection_type_option/multi_select_type_option.rs
appflowy: Got it. It would be better to have documentation there.

PR #3127, opened 2023-08-06
fix: push to sign in screen when logout

appflowy:
#### PR Checklist


frontend/appflowy_flutter/lib/user/application/auth/supabase_auth_service.dart
```diff
@@ -229,16 +245,15 @@ Completer<Either<FlowyError, UserProfilePB>> supabaseLoginCompleter({
...
     final user = event.session?.user;
-    if (event.event != AuthChangeEvent.signedIn || user == null) {
-      completer.complete(left(AuthError.supabaseSignInWithOauthError));
-    } else {
+    if (event.event == AuthChangeEvent.signedIn && user != null) {
       final response = await onSuccess(
         user.id,
         user.email ?? user.newEmail ?? '',
       );
+      // Only cancle the subscription if the Event is signedIn.
```
LucasXu0, line 253:
```suggestion
      // Only cancel the subscription if the Event is signedIn.
```

LucasXu0: Why cancel the subscription only when it's signed in? I think we should cancel it even if the sign-in event fails.

appflowy: The subscription will receive the event upon logout, which can sometimes cause the login to malfunction. Using this function means it only focuses on the login event

Issue #3119, opened 2023-08-05
[FR] (don't) wrap column


richardshiue:
### Description

In the grid, most if not all fields wrap cell contents if they cannot fit in a single line. There should be a consistent behavior and setting to toggle this on and off, which will allow users to have consistent row height if they wish. In this screenshot, the title field wraps, the multi-select wraps, but date ellides and number has a hard clip.

![image](https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/26cf128b-5e66-4b57-8021-fc1f107be5d3)

### Impact

Users who want to have consistent row height in their grids

clevertension:
yes, the url cell should also support auto wrap

richardshiue:
Closing as completed via https://github.com/AppFlowy-IO/AppFlowy/pull/5128

PR #3128, opened 2023-08-06
fix issue 3119 to auto wrapper the date, number and url cell

clevertension:
### Feature Preview

image

fixes https://github.com/AppFlowy-IO/AppFlowy/issues/3119
---

just make the **maxLines** to null of what text cell does

#### PR Checklist


richardshiue (approved):
Thanks for the contribution! After this change, everything wraps by default. Though, I'd prefer to keep the original issue open after merging this since we'll work on allowing the user customize the wrapping behavior later.

clevertension:
yes, exactly, i will contribute the code if we start to allow the user customize the wrapping behavior

PR #3131, opened 2023-08-06
Update zh-TW translations

StarAtty:
I've updated the translations to make them more consistent and natural. New translations were added.

Preview the messages on https://inlang.com/editor/github.com/StarAtty/AppFlowy


frontend/resources/translations/en.json
```diff
@@ -657,5 +657,6 @@
       "nature": "Nature",
       "frequentlyUsed": "Frequently Used"
     }
-  }
+  },
+  "grid.settings.Properties": "Properties"
```
richardshiue, line 661: Thank you for the contribution! I think this token already exists in `en.json` L314. Could you double check it?

LucasXu0: Due to the limitations of `easy_translation`, we are unable to define the key with dot syntax(.) in the `en.json` file. 

Please remove it or redefine it as 

```
grid {
  settings: {
    Properties: ...
  }
}
```

StarAtty: I think I messed that up in inlang. I have removed the changes in L660.

LucasXu0 (approved):

PR #3134, opened 2023-08-07
cleanup PR, update `inlang.config.js`, followup PR #2802

NilsJacobsen:
Hey folks, this is a follow up PR to https://github.com/AppFlowy-IO/AppFlowy/pull/2802

I cleaned up the `inlang.config.js` and updated it to the newest version.

- Newest version of plugin-i18next
- Newsest version of standard-lint-rules


qinluhe:
Can you explain the purpose of this PR？

NilsJacobsen:
Hi @qinluhe,

> 1. 

The variable reference pattern is there for detecting where you use variables that are passed into the translation.
`"title": "Sign Up to @:appName"`-> example form `en.json`
Do you mean, that we are not allowed to use the `@` in this config file? I am wondering why it worked before, as I did not modify this line of code.

> 2.

Do you want a screenshot from the editor while using the extension with Appflowy? Sure.
image

> 3. There is no need to change anything in the i18next configuration. Inlang only use the resource files to access the translations. 

**Update:**
I cleaned up the translation files.

Old:
```json
{
   "a": {
      "b": "text",
   },
   "a.c": "test",
}
```

Cleaned:
```json
{
   "a": {
      "b": "text",
      "c": "text",
   },
}
```

This issue is due to the old version of the plugin, which I updated. People used the editor to add translations, which added the translations as flattened keys. But with the new plugin version, this should be detected.
**So I didn't change any values.**

Issue #2911, opened 2023-06-29
[FR] Import a page from the zip file provided by Notion


Mukund-Tandon:
### Description

When we export a page from Notion we get a zip file, which contains a markdown file with all the page contents and another folder having all the assets of the page like images.

<img width="266" alt="Screenshot 2023-06-29 at 5 41 09 PM" src="https://github.com/AppFlowy-IO/AppFlowy/assets/71614009/6f947cb8-7d03-4c10-9f05-2bd12bb0ffb6">

In the import panel, there should be an option for importing from Notion. Then, we should have options to select what we want to import, such as pages, boards, grids, or calendars. When the user clicks on "pages" they should be able to select the zip file and import their page from there.

### Impact

Any user who wants to import a page from Notion

Mukund-Tandon:
I would like to work on this issue

LogicFan:
i think it would be better to fetch notion raw data using notion public api instead of using exported markdown since some info will be missing with exported markdown.

Mukund-Tandon:
> i think it would be better to fetch notion raw data using notion public api instead of using exported markdown since some info will be missing with exported markdown.

I think we can also implement fetching data from the Notion API along with the zip file. Since the Notion API does not provide the complete workspace, we need to use the export workspace option, which gives us a zip file. This issue is a prerequisite for that step.

Additionally, could you please let me know what things are missing with the exported markdown so that I can handle those cases appropriately?

LogicFan:
> > i think it would be better to fetch notion raw data using notion public api instead of using exported markdown since some info will be missing with exported markdown.
> 
> I think we can also implement fetching data from the Notion API along with the zip file. Since the Notion API does not provide the complete workspace, we need to use the export workspace option, which gives us a zip file. This issue is a prerequisite for that step.
> 
> Additionally, could you please let me know what things are missing with the exported markdown so that I can handle those cases appropriately?

one example would be "Notion pages that contain parentheses or dashes in the title will have them removed by Notion while exporting your data so the file will be created without them, even though the link itself will still retain them." as quoted from an opensource implementation of notion2obsidian converter.

AkechiShiro:
Any milestone version for this feature @annieappflowy ?

appflowy:
We support importing notion as a new workspace. I will close this FR

MarcusDelvecchio:
[...]

PR #2981, opened 2023-07-11
feat: Import a page from zip file 

Mukund-Tandon:
### Feature Preview
This PR is WIP
Current progress - 

https://github.com/AppFlowy-IO/AppFlowy/assets/71614009/e506ec30-7299-4807-a2cf-cb7c445cd253

Fixes #2911 
---

#### PR Checklist

- [ ] My code adheres to the [AppFlowy Style Guide](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/submitting-code/style-guides)
- [X] I've listed at least one issue that this PR fixes in the description above.
- [ ] I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.
- [ ] All existing tests are passing.

PR #3146, opened 2023-08-08
feat: added import a page from notion feature

Mukund-Tandon:
Fixes: #2911 
### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/71614009/89eca69a-81b3-449a-b162-8652c71f18ce

This PR is related to https://github.com/AppFlowy-IO/AppFlowy/pull/2981

---

# How This feature is implemented -
### What does a page look like when exported from Notion-
We  get a zip file containing the the main page markdown and a folder containing assets of the main page and also markdown files of subpages of the main page and folders containing the assets of the subpages and the sub-sub pages markdown files and so on.

image

### How our main page markdown file look
We can see on line 5,6  we have two images inside the round brackets is the path where images are located and on line 19 ,21,23 are subpages

image

### Markdown file of subpage
Here we have to look at the image path it starts with AppFlowy Subpage 1 instead of instead of AppFlowy Test which is the parent folder. This is because it’s asset lies in AppFlowy Sub 1 folder which is at the same level where the  subpage markdown file is . This fact will be used late while importing.

image

image

### How is the Importing working
- First we iterate through the unzipfiles list and store all files and assets in there respective level in the `levels` list
- Next step will be importing all the pages present starting from the last level.we are importing from the last level to handle the case of subpages . For parent page to show `MentionBlock` of the subpage it requires the SubPage viewID which can only be generated once it is imported so we import the lowest level sub page first and then go up storing each page viewID in a map (**nameToID**) which has key as the page name and value as the page viewID.
- Next once we have all page imported we would move them under their respective parent

image
This class contains details of each page to be imported and the parent of that page which will be used for moving the page under correct parent in later steps
image
This class contains details of each level of the files. Like For example the main page will be level one , if it has a subpage it will be level 2 and if that subpage will have another subpage that would belong to level 3

image

### When we are importing a markdown page how are we dealing with images
- To deal with images we take all contents of a markdown file and pass it through `_preProcessMarkdownFile`   function which returns us a string which is the contents of the markdown file but with changes. The changes this function performs are related to images . It will iterate through each line and if it finds something like `name` this is how a image is represented in markdown . When we get this line is detected we get the path from this this path is actually the file name of image from the above list of unzip images so with the help of path we will get the image file and save it locally and change the current path to the path where the image is saved locally
- Now the problem we can see here is as discussed before the path name in a sub page starts with AppFlowy Sub page instead of Appflowy test which is the root . In the above list also you can see that all the file name are starting from AppFlowy Test so if we directly pass this list to `_preProcessMarkdownFile`  it wont be able find the image file.So this issue was solve by only passing the assets of a particular level to the _preProcessMarkdownFile function instead of all the assets and also during the process of adding files/assets  in the levels list if a file is not added in the particular level then the first part(parent) from it name would be removed  

### How does Subpages are handled while importing a page ?
image

We are imported from level 3 to level 1. This is done because to handle the subpages . We use `markdownToDocument` function of appflowy-editor package to chick we pass the markdown contents of our page and it return us a Document. We also pass a custom-parser to this function which handles our subpage. Whenever it detects a subpage it replace the part with a `MentionBlock`(It contains the pageID of the page that is being Mentioned) . This is reason  we are importing the lower level page first . We will import them and store their viewID in a map where the key is pageName and value is the pageID. When above level are being imported and there contents are passed through `_preProcessMarkdownFile` function this map(nameToId) is also passed in the function ,   whenever it finds a subpage which looks like [name](path)  to `{{AppFlowy-Subpage}}{$subpageName}{$subPageID} `, this pattern is used in custom parser to detect the  subpage part and convert it to mention block. We have stored the pageIds with name as key so we can easily get the pageID of subpage by using name from [name](page) .

#### PR Checklist

- [X] I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.


frontend/appflowy_flutter/integration_test/import_page_from_notion_test.dart
```diff
@@ -0,0 +1,166 @@
...
+
+      expect(find.widgetWithText(Card, 'From Markdown Zip'), findsNothing);
+      await tester.tapButtonWithName('Import from Notion Markdown Zip');
+      expect(find.widgetWithText(Card, 'From Markdown Zip'), findsOneWidget);
+      await tester.tapButtonWithName('From Markdown Zip');
+      expect(find.text('Import Notion From Markdown Zip'), findsOneWidget);
+      await mockPickFilePaths(
+        paths: paths,
+      );
+      await tester.tapButtonWithName('Upload zip file');
```
LucasXu0, line 52: Same here. Use i18n instead.

frontend/appflowy_flutter/integration_test/import_page_from_notion_test.dart
```diff
@@ -0,0 +1,166 @@
...
+      tester.expectToSeePageName(mainPageName);
+      await tester.openPage(mainPageName);
```
LucasXu0, line 54: The import panel should close automatically after selecting files.

Also, avoid using `openPage` to close the panel. Even if you intend to close it, you can simply tap on a blank area, like offset(0, 0).

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/app/header/import/importer/notion_importer.dart
```diff
@@ -0,0 +1,233 @@
...
+        break;
+    }
+    return;
+  }
+
+  Future<void> _importFromMarkdownZip(String path) async {
+    final zip = File(path);
+    final bytes = await zip.readAsBytes();
+    final unzipFiles = ZipDecoder().decodeBytes(bytes);
+    final List<Pair<List<Pair<String, ArchiveFile>>, List<ArchiveFile>>>
```
LucasXu0, line 45: This value is overly complex.

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/app/header/import/importer/notion_importer.dart
```diff
@@ -0,0 +1,233 @@
+import 'dart:convert';
```
LucasXu0, line 1: Please add more comments to this file.

The logic within it is not readily understandable.

LucasXu0:
@Mukund-Tandon Please include a brief description here to explain how you implemented this feature.

Mukund-Tandon:
https://github.com/AppFlowy-IO/AppFlowy/assets/71614009/d2f03841-7ce0-49aa-9e5e-70fe23f63d5d

I have implemented the importing of the subpages along with the main page @LucasXu0 , please can you take a look at it.

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/sidebar/import/import_from_notion_widget.dart
```diff
@@ -0,0 +1,166 @@
...
+          const FlowySvgData('external/notion-logo'),
+          color: Theme.of(context).colorScheme.tertiary,
+        ),
+        leftIconSize: const Size.square(20),
+        text: FlowyText.medium(
+          LocaleKeys.importPanel_importFromNotionMarkdownZip.tr(),
+          fontSize: 15,
+          overflow: TextOverflow.ellipsis,
+        ),
+        onTap: () async {
```
Xazin, line 50: Redundant `async`

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/sidebar/import/import_type.dart
```diff
@@ -80,3 +80,43 @@ enum ImportType {
...
+    switch (this) {
+      case ImportFromNotionType.markdownZip:
+        return '''
+        1. Navigate to the desired page for export.
+        2. Click on the three dots(···) located in the top right corner.
+        3. Choose "Export" from the menu options.
+        4. Select "Markdown & CSV" as the export format.
+        5. Click the "Export" button once again.
+        6. Select the file you just downloaded.
+        ''';
```
Xazin, line 119: I assume this should be localized

Mukund-Tandon:
I will fix the issues , I was not active for a while because got busy with college exams ,now I will continue working on this feature.

SimonBiggs:
@Mukund-Tandon this is some absolutely amazing work!

Thank you for all the effort you're going to here!

skydiver1980 (approved):

Killian-Aidalinfo:
Hello, excellent feature! Thanks for all your hard work. Do you know when this option will be available?
Thank you and have a nice day

Evil-Coder:
Any updates?

anklav24:
I'm also looking forward to this PR. Any updates? =)

Artemis1-0:
Looking forward to this PR. Hope to see it soon

appflowy:
We support importing notion as a new workspace. So i will close this PR

Artemis1-0:
Nice, can’t wait.

PR #3162, opened 2023-08-11
fix: arrow keys on field name text field

### LINKED ISSUE #3158: [Bug] -> doesn't work when editing a property name
### Bug Description

image

### How to Reproduce

Click on a property name
Try to use -> to move the cursor from the very beginning

### Expected Behavior

I can use -> to move the cursor

### Operating System

Windows 11

### AppFlowy Version(s)

v0.2.8

### Screenshots

_No response_

### Additional Context

_No response_

richardshiue:
resolves #3158

Also makes the cursor begin at the end of the text in a FlowyTextField if an initial value is provided.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/header/field_editor.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/header/field_editor.dart
@@ -1,7 +1,6 @@
 import 'package:appflowy/plugins/database_view/application/field/field_editor_bloc.dart';
 import 'package:appflowy/plugins/database_view/application/field/type_option/type_option_context.dart';
 import 'package:appflowy_popover/appflowy_popover.dart';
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/lib/plugins/database_view/grid/presentation/widgets/header/field_editor.dart
```diff
@@ -174,39 +175,28 @@ class _FieldNameTextFieldState extends State<FieldNameTextField> {
         focusNode.unfocus();
       }
     });
-
-    super.initState();
   }
 
   @override
   Widget build(BuildContext context) {
-    return BlocListener<FieldEditorBloc, FieldEditorState>(
```
appflowy, line 183: The name might be changed by other devices. So it needs to set the name if it's modified

Issue #3135, opened 2023-08-07
[FR] Right click on View Item to open context menu


Xazin:
### Description

Currently on a navigation item, only when hovering the popover button to show the context menu shows, and you can only open it by clicking the button (`:`).

We should refactor this so that it becomes possible to right-click anywhere on the navigation item to open the context menu. 

The context menu opened with right-click should open relative to mouse.

### Impact

UX

dotarjun:
working on this

annieappflowy:
Are you still working on this, @dotarjun ?

dotarjun:
Hey! I'm not working on this

ax920:
Hi, I'd like to take this on

Xazin:
In case this is going to be picked up at some point, here is some information on the preferred solution:
- Enabling Secondary Click as a Popover Trigger
- Implementing a boolean on whether the Offset should adhere to the pointers offset on trigger, or if it should adhere to child. Eg. `usePointerPosition` - It should default to false so it doesn't break existing Popovers.
- In the Listener used to track whether a click happens, we can make a check inside the `onPointerDown` callback, that checks whether the `event.buttons` is equal to [kSecondaryMouseButton](https://api.flutter.dev/flutter/gestures/kSecondaryMouseButton-constant.html), similarly for the left click see [kPrimaryMouseButton](https://api.flutter.dev/flutter/gestures/kPrimaryMouseButton-constant.html)
- Now for the `showOverlay`, it should take an optional `Offset?`that we can provide in case `usePointerPosition` is set to `true`.
- Also the math for deciding where to display the Popover based on the PopoverDirection, should be different for when using the Pointer Offset, otherwise the popover will show up in odd locations.

PR #3186, opened 2023-08-13
feat: context menu on right-click

aryanA101a:
### Feature Preview

---
#3135 

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/view_item.dart
+++ b/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/view_item.dart
@@ -11,6 +11,7 @@ import 'package:appflowy/workspace/presentation/home/menu/view/draggable_view_it
 import 'package:appflowy/workspace/presentation/home/menu/view/view_action_type.dart';
 import 'package:appflowy/workspace/presentation/home/menu/view/view_add_button.dart';
 import 'package:appflowy/workspace/presentation/home/menu/view/view_more_action_button.dart';
+import 'package:appflowy/workspace/presentation/home/menu/view/view_more_actions.dart';
 import 'package:appflowy/workspace/presentation/widgets/dialogs.dart';
 import 'package:easy_localization/easy_localization.dart';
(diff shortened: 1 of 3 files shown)


aryanA101a:
@Xazin I have implemented the feature except for the "relative to mouse" part.
Please check and tell me If I am heading in the right direction.

aryanA101a:
@Xazin Please provide feedback

aryanA101a:
Hi, @LucasXu0. I have made some modifications to remove the redundancy, and the functionality is as we want it to be.
But I am running into an issue with an integration test. Specifically this one: `'context menu opens up for favorites'`
I would appreciate some guidance to continue.

LucasXu0:
It's not a suitable solution for this issue because the position of the popover action list will not track the touch point; it only appears in the middle of the view item wherever you click.

the better way in my mind is to 

1. Modify `AppFlowyPopover` to allow it to appear via a `SecondaryTap`. Currently, the `triggerActions` in `AppFlowyPopover` support the following three ways.
```dart
class PopoverTriggerFlags {
  static const int none = 0x00;
  static const int click = 0x01;
  static const int hover = 0x02;
}
```

2. Wrap the view item with a new popover that supports SecondaryTap and use secondaryTap as the flag.
3. Wrap the "View More" action button with the new popover and use the click as the flag.

aryanA101a:
@LucasXu0 thats a nice approach.
When I modified the AppFlowyPopover wrapped over viewmore button, to include the click flag at the same time I commented the onPressed of the viewmore button, I found that on tapping the viewmore button: popup opens up and closes at the same time.
Then to investigate I added a break point to popover widget's deactivate method. That was the case, deactivate was being called as soon as the popup opened. I think hover is causing this issue because when I removed the FlowyHover from the viewitem, it worked.

Xazin:
Hey @aryanA101a, are you still working on this? Do you need assistance?

Feel free to hit me up in our Discord channel, prefer the `#flutter-101` channel :+1:

Xazin:
I'm going to close this due to inactivity, feel free to open a PR again with the final work.

PR #3252, opened 2023-08-22
chore: bump version 0.3.0

LucasXu0:
## Version 0.3.0 - 08/22/2023

Screenshot 2023-08-22 at 16 29 55

### New Features

- Improve paste features:
  - Paste HTML content from website.
  - Paste image from clipboard.

- Support Group by Date in Kanban Board.
- Notarize the macOS package, which is now verified by Apple.
- Add Persian language translations.

### Bug fixes

- Some UI issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Release Notes
 
+## Version 0.3.0 - 08/22/2023
+
+### New Features
+
+- Improve paste features:
+  - Paste HTML content from website.
+  - Paste image from clipboard.
+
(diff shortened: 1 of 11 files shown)


Xazin:
Can we hide the logout button in the settings menu under user, it seems to cause irreversible damage to data for anonymous users at the moment.

I haven't looked too much into it, but I just tried it out on an anonymous user, and the same data folder became reset/empty.

LucasXu0:
@Xazin Yes. I'm fixing it.

Issue #2312, opened 2023-04-21
[FR] Support different layouts to view pages


annieappflowy:
### Description

This is a master issue for multiple layouts we should support to view pages or database records

- Multi panes: eg: split one into two and write side-by-side
- Tabs: open multiple tabs in the desktop / mobile app
- Sliding panes: [Andy Matuschak's notes](https://notes.andymatuschak.org/)
- Side pane: open a record in side pane so you can work on this record while referring to the entire database
- Open a record as a full page

### Impact

All users

squidrye:
Hi @annieappflowy, I would like to work on this.

annieappflowy:
Hi @Xazin , @squidrye would like to work on another project while waiting on his PR review. I think he can be of good help for the multi-pane feature.

annieappflowy:
<img width="485" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/12026239/90d3ea41-a4b4-4a0c-a659-dddb9301252f">
<img width="478" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/12026239/30c3e9d6-2bea-4e6a-8b91-124f491ab34f">

Xazin:
> Hi @Xazin , @squidrye would like to work on another project while waiting on his PR review. I think he can be of good help for the multi-pane feature. 

Yes, we are in contact. 🙏

Issue #11, opened 2021-11-16
[Bug] Fonts doesn't load or text is garbled


cmrajan:
**Describe the bug**

**To Reproduce**
1. Install version 0.0.2 from the releases
2. Open the App
3. Garbled text
4. See Screenshot

**Expected behavior**
Show correct sample text

**Screenshots**
<img width="782" alt="image" src="https://user-images.githubusercontent.com/410065/141936407-788d0bc8-7497-49d7-8af0-bf3241a71dc4.png">

**Desktop (please complete the following information):**
 - OS: MacOS Sierra
 - Browser NA
 - Version 10.12.6

annieappflowy:
Hi cmrajan,

Thanks for checking us out! Could you please let us know your system's language?

Annie

cmrajan:
US-en

appflowy:
> Has this ever been reproduced by anyone? @cmrajan Do you still have this problem?

Annie was able to reproduce on her old MacBook

apomalyn:
I cannot reproduce the bug with MacOS Monterey and AppFlowy v0.0.4.1. Is anyone able to reproduce it now?

MikeWallaceDev:
Closing as nobody seems to reproduce this anymore

PR #3337, opened 2023-09-04
 feat: Support Panes for AppFlowy

squidrye:
### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/84044317/7c98c24b-a965-4505-afaa-a5b985a72ef1

---
#2312 This PR targets implementation of multi-panes
cc @Xazin 

#3237 Reopening this PR, had to delete my older fork because I had pushed private code due to local git misconfiguration 😢 

#### PR Checklist

#### Todos
- [x] Decide and implement data structure for panes
- [x] Map n-ary tree to bloc for state management, allow opening and closing of panes
- [x] Refactor existing tabs implementation to each pane node
- [x] Link each pane to correct plugin and view 
- [x] Resizable panes
- [x] Setup drag and drop for tabs
- [x] Setup drag and drop for panes
- [x] Allow horizontal/ vertical scroll for stacked panes based on visible viewport
- [x] Stack panes over another (Sliding panes)
- [x] Refactor other features which might break due to this (Mostly scroll related issues)
- [x] Enable drag and drop across views, tabs and panes
- [x] Handle edge cases for view -> tab & tab -> tab drops
- [x] Handle edge cases for view -> pane & tab -> pane  drops
- [ ] Refactor cross-draggable entity for better reusability
- [ ] Might need a better design spec for tab-bar, with/without tabs
- [x] Tests


LucasXu0:
@squidrye I have merged the editor optimization into the main branch. You can now rebase the latest main branch and test the performance of multiple editor's windows again.

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/base/built_in_page_widget.dart
```diff
@@ -155,11 +155,9 @@ class _BuiltInPageWidgetState extends State<BuiltInPageWidget> {
...
             switch (action.inner) {
               case _ActionType.viewDatabase:
-                getIt<TabsBloc>().add(
-                  TabsEvent.openPlugin(
-                    plugin: viewPB.plugin(),
-                    view: viewPB,
-                  ),
+                getIt<PanesCubit>().openTab(
+                  plugin: viewPB.plugin(),
+                  view: viewPB,
```
Xazin, line 160: I guess we could simplify this and just pass the `viewPB`, are there any cases where we only have the plugin?

Since you're touching this we can refactor if need be.

frontend/appflowy_flutter/lib/workspace/application/panes/panes_cubit/panes_cubit.dart
```diff
@@ -0,0 +1,113 @@
...
+    emit(
+      state.copyWith(
+        root: root,
+      ),
+    );
+    emit(
+      state.copyWith(
+        count: panesService.countNodeHandler(root),
+      ),
+    );
```
Xazin, line 61: Double emit?
```suggestion

    emit(
      state.copyWith(
        root: root,
        count: panesService.countNodeHandler(root),
      ),
    );
```

frontend/appflowy_flutter/lib/workspace/application/panes/panes_service.dart
```diff
@@ -0,0 +1,152 @@
...
+import 'package:appflowy/workspace/presentation/home/home_stack.dart';
+import 'package:appflowy_backend/protobuf/flowy-folder2/view.pb.dart';
+import 'package:flutter/material.dart';
+import 'package:nanoid/nanoid.dart';
+
+enum Direction { front, back }
+
+class PanesService {
+  PanesService();
+  PaneNode splitHandler(
```
Xazin, line 14:
```suggestion
  const PanesService();

  PaneNode splitHandler(
```

frontend/appflowy_flutter/lib/workspace/application/tabs/tabs.dart
```diff
@@ -0,0 +1,83 @@
...
+
+    /// If currentIndex is greater than the amount of allowed indices
+    /// And the current selected tab isn't the first (index 0)
+    ///   as currentIndex cannot be -1
+    /// Then decrease currentIndex by 1
+    final newIndex = currentIndex > _pageManagers.length - 1 && currentIndex > 0
+        ? currentIndex - 1
+        : currentIndex;
+
+    currentIndex = newIndex;
```
Xazin, line 42:
```suggestion
    currentIndex = currentIndex > _pageManagers.length - 1 && currentIndex > 0
        ? currentIndex - 1
        : currentIndex;
```

frontend/appflowy_flutter/lib/workspace/presentation/home/home_stack.dart
```diff
@@ -280,6 +275,18 @@ class HomeTopBar extends StatelessWidget {
...
+            BlocBuilder<PanesCubit, PanesState>(
+              builder: (context, state) {
+                return state.count > 1
+                    ? IconButton(
+                        onPressed: () {
+                          context.read<PanesCubit>().closePane(paneId);
+                        },
+                        icon: const Icon(Icons.close_sharp),
+                      )
+                    : const SizedBox.shrink();
```
Xazin, line 287: I prefer this, but it has same effect, if you prefer the other way just keep it as is :+1:
```suggestion
                if (state.count <= 0) {
                  return const SizedBox.shrink();
                }
                
                return IconButton(
                  onPressed: () => context.read<PanesCubit>().closePane(paneId),
                  icon: const Icon(Icons.close_sharp),
                );
```

frontend/appflowy_flutter/lib/workspace/presentation/home/hotkeys.dart
```diff
@@ -97,7 +96,10 @@ class HomeHotKeys extends StatelessWidget {
   }
 
   void _selectTab(BuildContext context, int change) {
-    final bloc = context.read<TabsBloc>();
-    bloc.add(TabsEvent.selectTab(bloc.state.currentIndex + change));
+    final bloc = context.read<PanesCubit>();
+    // bloc.add(TabsEvent.selectTab(bloc.state.currentIndex + change));
```
Xazin, line 100: Remove

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane.dart
```diff
@@ -0,0 +1,87 @@
...
+    return ChangeNotifierProvider<Tabs>(
+      create: (context) => node.tabs,
+      child: Consumer<Tabs>(
+        builder: (context, value, child) {
+          final horizontalController = ScrollController();
+          final verticalController = ScrollController();
+          return BlocBuilder<HomeSettingBloc, HomeSettingState>(
+            builder: (context, state) {
+              return Scrollbar(
+                controller: verticalController,
```
Xazin, line 36: Can we make sure the scrollbar fits the style of the other existing scrollbars, see eg. the vertical scrollbar in the kanban board.

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane.dart
```diff
@@ -0,0 +1,87 @@
...
+  const FlowyPane({
+    super.key,
+    required this.node,
+    required this.layout,
+    required this.delegate,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final pageController = PageController();
```
Xazin, line 26:
```suggestion
    final pageController = PageController();
    final size = MediaQuery.sizeOf(context);
```

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane.dart
```diff
@@ -0,0 +1,87 @@
...
+                                  .toList(),
+                            ),
+                          ),
+                        ],
+                      ).constrained(
+                        width: MediaQuery.of(context).size.width -
+                            (state.isMenuCollapsed
+                                ? 0
+                                : (state.resizeOffset + Sizes.sideBarWidth)),
+                        height: MediaQuery.of(context).size.height,
```
Xazin, line 75:
```suggestion
                        width: size.width -
                            (state.isMenuCollapsed
                                ? 0
                                : (state.resizeOffset + Sizes.sideBarWidth)),
                        height: size.height,
```

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane_group.dart
```diff
@@ -0,0 +1,141 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    if (node.children.isEmpty) {
+      return FlowyPane(
+        node: node,
+        delegate: delegate,
+        layout: layout,
+      );
+    } else {
```
Xazin, line 36: redundant else, since you return inside the if.

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane_group.dart
```diff
@@ -0,0 +1,141 @@
...
+        child: LayoutBuilder(
+          builder: (context, constraints) => Stack(
+            key: ValueKey(node.paneId),
+            children: [
+              ...node.children.indexed
+                  .map(
+                    (indexNode) => GestureDetector(
+                      child: ChangeNotifierProvider<PaneSizeController>(
+                        create: (context) => node.sizeController,
+                        builder: (context, widget) =>
```
Xazin, line 47: Is `builder` required? Couldn't you just use `child` here? _(I'm not too into change notifier providers, just curious)_

Xazin:
This looks really good Mihir! I'm stoked to try this out and have it in AppFlowy! Great work 🎉 👏

frontend/appflowy_flutter/lib/workspace/application/panes/panes_cubit/panes_cubit.dart
squidrye: Oh it regressed, will fix it

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane.dart
squidrye: Sure thing

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane.dart
squidrye: This is an older commit, I have refactored all calculations to panes_layout.dart

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane_group.dart
squidrye: Same here, not too into change notifiers either, I am hoping we will be able to refactor all these to blocs if we persist some data, will look into it though.

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/base/built_in_page_widget.dart
squidrye: I likely moved all of existing tabs logic without looking too much into it, I think we should refactor it, I suppose for opening trash we are using plugin.

Xazin: We only need the trash view/plugin once globally I believe? Maybe it's worth looking into making it persist and be available across an app lifecycle..

I know it's not really part of your task, but if you do this let me know, otherwise I'll put it on my list of things to check if we should do.

squidrye: Hmm, I will likely do something about it since I will refactor parts of it to implement drag and drop for tabs and panes.

squidrye:
https://github.com/AppFlowy-IO/AppFlowy/assets/84044317/38e4b7c9-ba87-4b9d-a0ee-02ccce87c157


I have covered most extensive parts of this feature, I will restructure tabs, enable reordering, fit style to existing themes and write some tests, besides that feature is built, to polish things out before release we would need to work on a few things, which I am not too sure how to implement:

1. GestureDetector doesn't receive taps on editor likely because it is stacked dynamically, thus taps on editor do not trigger an event to set active pane in bloc, though taps on tabbar / header do trigger that event.

2. Need to find a way to get current cursor location and a way to map it to scrollController, to automatically scroll horizontally/ vertically, since without it user will have to manually scroll if content goes out of small viewports in panes.

3. Optimize resizablity of panes, right now a lot of events are added during drag, which leads to flex calculations, so resizing panes doesn't feel crisp, any ideas on how we might improve this @Xazin ?

annieappflowy:
Can this feature get merged by this week?

squidrye:
Sure thing @annieappflowy

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/draggable_pane_target.dart
```diff
@@ -0,0 +1,238 @@
...
+          ),
+        ),
+      FlowyDraggableHoverPosition.tab => Positioned(
+          top: 0,
+          child: Container(
+            height: top,
+            width: widget.size.width,
+            color: Theme.of(context).hoverColor.withOpacity(0.5),
+          ),
+        ),
```
Xazin, line 124: Can we rewrite this to only calculate `top, left, right, bottom`, so that we only need one Widget.

Maybe something like this?
```
(
  double? left,
  double? top,
  double? right,
  double? bottom,
) _getPositioned(FlowyDraggableHoverPosition position) {
  double? left, top, right, bottom;

  switch (position) {
    ...
  }

  return (left, top, right, bottom);
}
```

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane_group.dart
```diff
@@ -0,0 +1,143 @@
...
+                    paneLayout.childPaneWidth,
+                    indexNode.$1,
+                    details.delta.dx,
+                  ),
+          onVerticalDragUpdate: (details) =>
+              context.read<PaneNodeCubit>().resize(
+                    paneLayout.childPaneHeight,
+                    indexNode.$1,
+                    details.delta.dy,
+                  ),
```
Xazin, line 105: I'm not sure what's a good way to improve this, I'll need to look at it in-depth.

Could we store the width of one of the panes at drag start, and then get the difference on drag to only change the width of the pane like that, or is this not viable?

Currently the drag behavior feels quite out of place since it doesn't match the cursor.

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/panes_layout.dart
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    childPaneLPosition =
+        parentPane.axis == Axis.vertical ? accumulatedWidth : null;
+
+    childPaneTPosition =
+        parentPane.axis == Axis.horizontal ? accumulatedHeight : null;
+
+    resizerWidth =
+        parentPane.axis == Axis.vertical ? 3 : parentPaneConstraints.maxWidth;
```
Xazin, line 48: Let's extract the magic number `3` into a const variable so it is more readable to know what it does.

frontend/appflowy_flutter/lib/workspace/presentation/home/tabs/draggable_tab_item.dart
```diff
@@ -0,0 +1,147 @@
...
+      final data = draggable.draggable as TabNode;
+      if (data.pageManager == widget.pageManager) {
+        return TabDraggableHoverPosition.none;
+      }
+      final threshold = size.width / 2;
+      if (offset.dx < threshold) {
+        return TabDraggableHoverPosition.left;
+      } else {
+        return TabDraggableHoverPosition.right;
+      }
```
Xazin, line 143:
```suggestion

        return TabDraggableHoverPosition.right;
```

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane_group.dart
squidrye: Yeah that would be a bit difficult because right now there is a List<double> in all the branch nodes, which equates flex for each consecutive child node, since having a double value for width in each leaf node(that is being rendered) meant I would have to do a whole tree traversal to recalculate all the other widths which would have been more expensive.

So storing initial offset for each pane would mean usage of another array and then modifying that, major struggle of this entire problem is finding the right way to convert the offset change to change in width / flex

I can give it a shot though to see if that might bring improvements.

frontend/appflowy_flutter/lib/workspace/presentation/home/home_stack.dart
```diff
@@ -201,23 +248,31 @@ class PageManager {
...
+  Widget stackTopBar({required HomeLayout layout, required String paneId}) {
     return MultiProvider(
       providers: [
         ChangeNotifierProvider.value(value: _notifier),
       ],
       child: Selector<PageNotifier, Widget>(
         selector: (context, notifier) => notifier.titleWidget,
         builder: (context, widget, child) {
-          return MoveWindowDetector(child: HomeTopBar(layout: layout));
+          return MoveWindowDetector(
```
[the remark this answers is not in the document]
squidrye: 🤔 I do not remember adding that, I guess it has always been there.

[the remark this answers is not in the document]
Xazin: Yeah, it's old code.

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/draggable_pane_target.dart
squidrye: Oh yeah that would be better

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/flowy_pane_group.dart
squidrye: Hey Mathias I tried this out, approach didn't do much for optimization however I was able to define a better algorithm for flex adjustment, from all the things I could try out I couldn't optimize it anymore it still does feel heavy in debug builds, there are a lot of junk frames however practically it feels pretty smooth when I was inspecting expensive parts in profile mode.

What do you think, here is a vid



https://github.com/AppFlowy-IO/AppFlowy/assets/84044317/15bb3596-5aca-48e4-a366-80d8ccf668e2

Xazin: This looks really good compared to the old one, I'll try it out :pray:

frontend/appflowy_flutter/lib/workspace/presentation/home/panes/draggable_pane_target.dart
```diff
@@ -0,0 +1,251 @@
...
+
+  Widget _buildChildren(
+    BuildContext context,
+    FlowyDraggableHoverPosition position,
+  ) {
+    final (left, top, height, width) = _getHoverWidgetPosition(position);
+
+    if (position == FlowyDraggableHoverPosition.none) {
+      return const SizedBox.shrink();
+    }
```
Xazin, line 83:
```suggestion
    if (position == FlowyDraggableHoverPosition.none) {
      return const SizedBox.shrink();
    }

    final (left, top, height, width) = _getHoverWidgetPosition(position);
```

Xazin:
Hey Squidrye, the team gathered some feedback, I'll list them below.

### [Must fix] Cannot scroll in Read-only view

### [Must fix] When having two vertical panes, the top bar of one can disappear

<details>
  <summary>Screenshot</summary>

image

</details>

### [Must fix] There are two scroll configurations for a view

<details>
  <summary>Recording</summary>

https://github.com/AppFlowy-IO/AppFlowy/assets/42929161/2db1742e-b194-4352-afec-641316e144ab

</details>

### [Nice to have] When having the same view open in two panes, and closing the one that is editable, the other one is still read-only

The fix I see right now is updating the read-only view if this happens so that it's synched with the latest changes.

### [Nice to have] Horizontal scrollbar disappears when contents cannot fit (it’s at the very bottom of the document)

### [Nice to have] The divider between panes seems too thick. When there’s only one pane, the divider between the sidebar and the window is only one pixel

---

Let me know which ones you resolve, the rest we will create new issues for.

We're working on developing synchronization between the backend and the document, this will eliminate the need for read-only views, but it's not certain when this will be ready for release.

squidrye:
Thanks for the review, I'll see what I can do to enable scroll in read only view, right now I have wrapped it with AbsorbPointer and opacity, thus no events go through to AppFlowy Editor's scroll.

Couldn't reproduce the one where top bar disappears I have a guess on why it happens, likely because Tabs which are being handled by change notifiers under bloc get disposed, when nodes in tree are updated, however I don't have any means to fix it unless I can reproduce it every time.

Two scroll configurations only happen for a view, if views/ databases have their own scrolling implemented, like AppFlowy editor has its own vertical scroll implemented, I think it is a separate issue because, I'll need to expose modifiable properties of these views and databases, still I'll see if there is a quick work around.

I am correcting the logic for opening and closing read only views, it is a bit tricky, because if lets say there are three panes of same kind, I'll have to traverse through the tree and all the tabs of each node in tree to remove read only status on next similar pane.

I will position vertical scroll bar to end of pane rather than view, will do 1 px for divider size, we can change it again if it doesn't fit in.

Xazin:
> Two scroll configurations only happen for a view, if views/ databases have their own scrolling implemented, like AppFlowy editor has its own vertical scroll implemented, I think it is a separate issue because, I'll need to expose modifiable properties of these views and databases, still I'll see if there is a quick work around. Edit: Just fixed it with nested scroll view
> 

I was about to say, use a nested scroll view.

> I am correcting the logic for opening and closing read only views, it is a bit tricky, because if lets say there are three panes of same kind, I'll have to traverse through the tree and all the tabs of each node in tree to remove read only status on next similar pane.
>

Maybe you can store the position of a read-only view in a separate map, where the key is the ID. When you close a tab/view/pane, you can iterate all the views against this map and you now have all the locations.

The only issue to be aware of if you went with such a solution is, that of course it also has performance implications similar to what you describe, but in this case it's limited to when closing panes with lots of open tabs.

Another thing to take note of, is that there should be duplicate checks before re-rendering a read-only view to an editable view, because the view can be open N-times and isn't limited to being open just twice.

squidrye:
I have incorporated most of the feedback as well as fixed a lot of problems I could find myself.

Scroll related suggestions could not be incorporated due to nature of current implementation (which I think is good for now), I can create a separate issue and try to work these two changes out: 

> [Must fix] There are two scroll configurations for a view
- The above isn't really a problem it was intended so that this feature does not interfere too much with sizing of other nested features, implementing this would require refactoring all the available plugins in appflowy, which I think is a separate issue on its own. I would also like to have a horizontal scroll for each of the plugins and expose them to higher level widgets in order to automatically scroll once cursor goes out of viewport, which again is not possible right now or maybe it is and I haven't looked around enough, let me know :eyes:.

> [Nice to have] Horizontal scrollbar disappears when contents cannot fit (it’s at the very bottom of the document)
- think it would be better to have this once above mentioned feature is implemented since it would overlap with scrolls and underlying content which might be a better experience compared to what we have right now. It is much more convenient to scroll via shift + scrollwheel or trackpad swipe gesture. 

> [Must fix] When having two vertical panes, the top bar of one can disappear
- I was able to reproduce this in a lot of scenarios actually, changenotifier was randomly getting disposed under blocs I didn't really know why though, using two different state management under one another is unconventional enough I guess, however since I didn't had access rust related notifiers as we aren't persisting any data, working with blocs under recursive build was a nightmare, change notifiers were a great fit for individual nodes of tree related data-structure, adding a `ValueKey` fixed random disposals of tabs, thus fixing the above issue.

Things I have added and should work: 

- editor is not automatically in focus, if opened as a read-only variant.
- tabs can be reordered.
- closing a view with write access leads to duplicated view being editable along with synced data
- read only views are scrollable (it is a workaround with gestures and timers pretty much, it isn't really scrollable), once sync feature is bought into appflowy this should be removed I suppose, should I add a todo for future?


I think this is review ready, there is always more to do however this PR is already too big, it would be better to work in iterations moving forward. I'll keep adding tests for a few more things until this is merged.

Xazin:
Resizing panes looks really clean now, good work Mihir!

I'm going through the feature now and noting down anything I find, hopefully most the finds can be done at a later timer :+1:

- Kind of weird behavior when resizing the menu, try it out
- We can improve the look of dragging a tab, the render itself is just the title of the tab, needs some form of design
- Cannot reorder tab to the 0th position, it automatically goes to index 1 if there are only 2 tabs
- When reordering a tab to 0th position, and there are many tabs (seen on MacOS) I get assertion error (see below Assertion: 1)
- The tabbar of the top pane, where there are no tabs, there shouldn't be draggable, on MacOS I can't move the window by dragging there. 😢 
- I guess we should figure out a way to more easily show all content horizontally within smaller views. Eg. max and min width of the editor, and reduce padding depending on pane horizontal size.
- There are multiple "toggle-sidebar" actions on each headerbar, I'm not sure what a good solution is, if it's possible to check and only add it to the one at the top-left
- The icon here https://github.com/squidrye/AppFlowy//blob/0c208c8cef8a3f7ba538fbbea61cd528d89a3450/frontend/appflowy_flutter/lib/workspace/presentation/home/home_stack.dart#L299 is wrong, and should be `FlowySvg(FlowySvgs.close_s)` and should not be that big

<details>
  <summary>Assertion: 1</summary>

  ```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: 'package:appflowy_editor/src/flutter/scrollable_helpers.dart': Failed assertion: line 236 pos 7: 'globalRect.size.width >= _dragTargetRelatedToScrollOrigin.size.width &&
          globalRect.size.height >=
              _dragTargetRelatedToScrollOrigin.size.height': Drag target size is larger than scrollable size, which may cause bouncing
#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      EdgeDraggingAutoScroller._scroll (package:appflowy_editor/src/flutter/scrollable_helpers.dart:236:7)
#3      EdgeDraggingAutoScroller.startAutoScrollIfNecessary (package:appflowy_editor/src/flutter/scrollable_helpers.dart:215:5)
#4      AutoScroller.startAutoScroll (package:appflowy_editor/src/editor/editor_component/service/scroll/auto_scroller.dart:48:5)
#5      _DesktopScrollServiceState.startAutoScroll (package:appflowy_editor/src/editor/editor_component/service/scroll/desktop_scroll_service.dart:103:19)
#6      _ScrollServiceWidgetState.startAutoScroll (package:appflowy_editor/src/editor/editor_component/service/scroll_service_widget.dart:173:13)
#7      _ScrollServiceWidgetState._onSelectionChanged.<anonymous closure> (package:appflowy_editor/src/editor/editor_component/service/scroll_service_widget.dart:117:9)
#8      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1284:15)
#9      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1223:9)
#10     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1072:5)
#11     _invoke (dart:ui/hooks.dart:142:13)
#12     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:359:5)
#13     _drawFrame (dart:ui/hooks.dart:112:31)
```

</details>

**Comment by squidrye** [comment]
I changed trigger for drag to left ctrl + lmb, should fix the issue on mac.

**Review by Xazin — APPROVED** [review]
I've cleaned up a bit

**Comment by Xazin** [comment]
Screenshot 2023-11-09 at 17 44 19

Need left padding + icon on hover needs to be correct color

---

Screenshot 2023-11-09 at 17 45 39

Maybe we should double the minimum width of panes

---

Screenshot 2023-11-09 at 17 46 29

We can improve this later I guess, but I believe this should be not inside the content of the view, but if possible just below the Title Bar.

File: frontend/appflowy_flutter/integration_test/board/board_row_test.dart [hunk 18]
```diff
@@ -23,6 +23,7 @@ void main() {
         onHover: () async {
           final moreOption = find.byType(CardMoreOption);
           await tester.tapButton(moreOption);
+          await tester.tapButton(moreOption);
```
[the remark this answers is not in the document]
Xazin: Hmm, that is very interesting.

There must be a reasonable solution, can you point me to the Listener? (I am on phone, still want to look)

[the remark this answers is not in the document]
squidrye: Oh sure, it is in presentation/home/panes/flowy_pane_group.dart.

frontend/appflowy_flutter/integration_test/database_calendar_test.dart
```diff
@@ -166,7 +166,7 @@ void main() {
       await tester.dragDropRescheduleCalendarEvent(firstOfThisMonth);
 
       // Make sure that the event has been rescheduled to the new date
-      final sameDayNextWeek = firstOfThisMonth.add(const Duration(days: 7));
+      final sameDayNextWeek = firstOfThisMonth.add(const Duration(days: 14));
```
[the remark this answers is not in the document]
squidrye: This I am not sure either, test was failing cause drag rescheduled event to a new date which is 7 days later, not 14 days later, not really sure which part of this feature is causing this behaviour.

skydiver1980 (approved):

annieappflowy:
what blocks this PR from merging?

Xazin:
> what blocks this PR from merging?

Due to the sheer size and complexity of the work here, I requested a secondary review from @LucasXu0 , but it was in the middle of the work on mobile, so it was pushed down in priority for a while.

Xazin:
Thanks @squidrye for the merge :pray:

rileyhawk1417:
Could the code here be reused for the future or we would have to start on a clean slate? Seeing that it wasn't merged.

khorshuheng:
Yes, we do remember this PR. Though we will need to test this again, consider that a lot of the code base has changed since this PR was raised.

PR #3348, opened 2023-09-05
fixed text overflow issue of font family dropdown

### LINKED ISSUE #3319: [Bug] Settings -> Appearance -> Font Family: Font name hidden
### Bug Description

When attempting to change the application's font, if the font name is too long to fit in the drop down window it is wrapped (I assume) and the full name becomes invisible.

This makes it impossible to tell the correct font you want to apply to the application.

### How to Reproduce

Open the Settings window.

Navigate to Appearance.

Select the current font within the "Font Family" option. A drop down appears with a search box.

Search for "Noto Sans". A popular font.

Several dozen other "Noto Sans" varieties of fonts appear. But since the font names are too long to actually appear on one line in the drop-down their names are truncated. This makes it impossible to determine which is the correct font you want to apply.

### Expected Behavior

I would expect the full name to at least be partially/ be visible within the drop-down window, even if the font name is too long to fit on a single line.

Seems like the only thing that needs to happen here is word-wrap needs to be turned off for this UI element?

### Operating System

Windows 11 Pro

### AppFlowy Version(s)

0.3.0

### Screenshots

image

### Additional Context

_No response_

AslamThachapalli:
closes #3319 

## Added Tooltip for overflown font family text
Image Description

### Added the following function to check whether a text is overflowing.
```dart
 bool _isTextOverflowing(String text, TextStyle textStyle, double maxWidth) {
    final textPainter = TextPainter(
      text: TextSpan(text: text, style: textStyle),
      maxLines: 1,
      textDirection: ui.TextDirection.ltr,
    )..layout(maxWidth: maxWidth);

    return textPainter.didExceedMaxLines;
  }
```

---

#### PR Checklist


MayurSMahajan:
Great PR @AslamThachapalli! Can we add a delay of about ~300ms on the tooltip, so it shows up a bit late.

AslamThachapalli:
Thanks @MayurSMahajan :)
And I have added the delay.

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
```diff
@@ -109,35 +111,61 @@ class _ThemeFontFamilySettingState extends State<ThemeFontFamilySetting> {
...
-          parseFontFamilyName(style.fontFamily!),
-          fontFamily: style.fontFamily!,
+
+    final bool isOverFlown = _isTextOverflowing(
+      buttonFontFamily,
+      Theme.of(context).textTheme.bodyMedium!.copyWith(
+            fontWeight: FontWeight.w500,
+            fontFamily: buttonFontFamily,
+          ),
+      110,
```
hyj1204, line 133: Where does this "110" come from? Can we use a variable to replace it(like fontPreviewTextWidth?

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
```diff
@@ -109,35 +111,61 @@ class _ThemeFontFamilySettingState extends State<ThemeFontFamilySetting> {
...
+    );
+
+    return Tooltip(
+      message: isOverFlown ? buttonFontFamily : "",
+      margin: const EdgeInsets.only(right: 220),
```
hyj1204, line 138: I feel it looks better if we remove the margin(default tooltip margin instead). Closer to the original font preview. 
It did cover the next line, but if we moved the cursor around, it would disappear anyway. 

What do you think? @AslamThachapalli @MayurSMahajan @annieappflowy 
image

hyj1204 (changes requested):

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
Xazin: Also that `220` is another magic number.

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
```diff
@@ -109,35 +111,61 @@ class _ThemeFontFamilySettingState extends State<ThemeFontFamilySetting> {
...
+      waitDuration: const Duration(milliseconds: 300),
+      child: SizedBox(
+        key: UniqueKey(),
```
Xazin, line 141: Do we need a `UniqueKey` here? There is already a `Key` in the `FlowyButton`.

If we do, then try to remove line 144 `key: Key(buttonFontFamily),` and use ValueKey on line 141
```suggestion
        key: ValueKey(buttonFontFamily),
```

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
MayurSMahajan: @hyj1204 I agree.

AslamThachapalli: Ok, will correct this. 
What my thought was even though there is a tooltip the user will be able to see the below text option. 
But as you pointed out, it will anyway disappear when the cursor is moved around.

@Xazin Yeah, actually I tried with different margins and 220 looked good, so fixed it. Will revert to the default way :)

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
AslamThachapalli: Checked this.  It looks like neither of the keys is necessary here. The flow is working fine without keys. 

Should I remove both keys or should I pick this suggested change?

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
AslamThachapalli: - It looks like the `AppFlowyPopover` has a  default max and min width of 160 and 80 resp. Here in this case, it looks like it  always keeps its max width constraint. So the complete width of the popover is taken as 160. 
- Then there is default margin for popover "6 each side", thus deducted 12 from 160. 
- Then came `FlowyButton`, it also has margins, "10 horizontally", thus can deduct 20 again. 
- This brings us to 128 as the available space for text. (almost)
- But still some texts was overflowing due to its fontsize, so I tried with smaller values, and 110 looked like a decent value to cover most of the text overflows. 

That was the process. 
What you think?  Is there any better approach that I can pick to figure this width?

hyj1204: Got it. After playing around for a while, I am thinking is it better to add the tooltip on all the fonts? Because I noticed even though some font didn't overflow, it is still kind of hard to read it like the following one. 


image

What do you think?  @MayurSMahajan @Xazin @AslamThachapalli @annieappflowy

AslamThachapalli: Ok. But won't this make the ui less smooth? Like the tooltip comes in the way of scrolling.

Preview Vid:

https://github.com/AppFlowy-IO/AppFlowy/assets/113328135/4974fcca-d3fc-4b28-ad18-12fbe1c2ed02

What do you think? Should I move forward with this?

hyj1204: It looks fine to me though, I just feel the delay time of the tooltip makes it seem not to match the font that the current cursor points to. What do you think? Add a tooltip to **all the fonts** or just **the ones that overflow** is better? @Xazin @annieappflowy

skydiver1980 (approved):

AslamThachapalli:
Hi. Any update required for the pr?

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
```diff
@@ -167,37 +167,44 @@ class _FontFamilyDropDownState extends State<FontFamilyDropDown> {
...
-                  .syncFontFamily(fontFamily);
+
+    return Tooltip(
+      message: buttonFontFamily,
+      waitDuration: const Duration(milliseconds: 150),
+      child: SizedBox(
+        key: ValueKey(buttonFontFamily),
+        height: 32,
+        child: FlowyButton(
+          // key: Key(buttonFontFamily),
```
Xazin, line 178: Remove

Xazin (approved):

frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/settings_appearance/font_family_setting.dart
LucasXu0, line 178:
```suggestion
```

LucasXu0 (approved):

LucasXu0:
cc @hyj1204 

Screenshot 2023-12-18 at 20 01 48

hyj1204 (approved):

hyj1204:
> cc @hyj1204
> 
> Screenshot 2023-12-18 at 20 01 48

I just approved

PR #3418, opened 2023-09-16
chore: checklist ux flow redesign

richardshiue:
- solve the new task button vs create confusion. the new task button is no more and the new task text field with the create button is always visible, and separated from the other tasks
- the create button is now greyed out when the text is empty to indicate to users that you can't submit an empty task
- editing an existing task now updates it in place without the user having to submit by pressing enter

About the create button's padding Lucas, I figured that since I have it separated, it could be nearer to the edge. If you think it would be better being right aligned with the delete button, LMK!

[Screencast_20230916_163937.webm](https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/7b3c2fe7-7bfc-478a-81ac-eaa851a75057)

### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/checklist_cell/checklist_cell.dart
```diff
@@ -10,10 +10,30 @@ import 'checklist_cell_bloc.dart';
...
+  GridChecklistCell({
+    required this.cellControllerBuilder,
+    GridCellStyle? style,
+    super.key,
+  }) {
+    if (style != null) {
+      cellStyle = (style as ChecklistCellStyle);
+    } else {
+      cellStyle = null;
+    }
```
Xazin, line 35: Could this work?
```suggestion
    cellStyle = style as ChecklistCellStyle?;
```

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/checklist_cell/checklist_cell_editor.dart
```diff
@@ -249,13 +226,9 @@ class _ChecklistItemState extends State<ChecklistItem> {
...
-                  onSubmitted: (taskDescription) {
-                    context.read<ChecklistCellEditorBloc>().add(
-                          ChecklistCellEditorEvent.updateTaskName(
-                            widget.option.data,
-                            taskDescription,
-                          ),
-                        );
+                  onChanged: _debounceOnChangedText,
+                  onSubmitted: (description) {
+                    _submitUpdateTaskDescription(description);
```
Xazin, line 231:
```suggestion
                  onSubmitted: _submitUpdateTaskDescription,
```

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/checklist_cell/checklist_cell_editor.dart
```diff
@@ -276,6 +249,24 @@ class _ChecklistItemState extends State<ChecklistItem> {
       ),
     );
   }
+
+  void _debounceOnChangedText(String text) {
+    _debounceOnChanged?.cancel();
+    _debounceOnChanged = Timer(const Duration(milliseconds: 300), () async {
+      if (mounted) {
+        _submitUpdateTaskDescription(text);
+      }
```
Xazin, line 258: Shouldn't need a mounted check, you're not doing a `setState`, if the user leaves the dialog within 300 ms of making a change, then the last change wouldn't be saved. And although it's highly unlikely, we can eliminate it by not having the mounted check. What do you think?

Xazin:
If the last checklist item is empty, I cannot create a new one.

I can create a new one if any but not the last item, is empty.

This behavior feels kind of weird to me, I understand the concept of "Do not create a new item since you already have an empty item to use", but I'm not sure this is a good experience. When pressing create the item disappears but does not appear in the checklist.

richardshiue:
> If the last checklist item is empty, I cannot create a new one.
> 
> I can create a new one if any but not the last item, is empty.
> 
> This behavior feels kind of weird to me, I understand the concept of "Do not create a new item since you already have an empty item to use", but I'm not sure this is a good experience. When pressing create the item disappears but does not appear in the checklist.

I've added a commit that allows creation of tasks that are of the same name. But I wasn't able to reproduce the problem you're facing:

1. open the cell editor and create 2 tasks: "task 1" and "task 2"
2. Clear the text for "task 2"
3. Create a new task "task 3

Task 3 did show up: "task 1", "", "task 3". Maybe I misunderstood you?

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/checklist_cell/checklist_cell.dart
richardshiue: This is much simpler! Thanks for this.

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/checklist_cell/checklist_cell_editor.dart
richardshiue: I can't do this since there's another function call below.

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/checklist_cell/checklist_cell_editor.dart
richardshiue: Yeah that's reasonable. I added a commit to address this.

hyj1204:
I suggest discarding empty options and allowing only unique option names.

Here is the [prototype(Figma Link)](https://www.figma.com/file/YDlrA62CPHasITX3B5F4o2/Checklist-UX-Proposal?type=whiteboard&node-id=0%3A1&t=lmrfvrKVgXmC6wOw-1) I think would be good for this workflow. Basically, we need an "Add" text/icon button to show on the dialog when the user is not in typing mode(do not have a text field).  Remove "Add" text/icon button when the user is in typing mode(has a text field). 

image

richardshiue:
That could work. It's somewhat similar to the select option editor. Some concerns:

- I'd still like the tasks to be editable in place without having to press enter or click a button to submit the change. If they are automatically discarded when empty or become identical (like while editing), this will be tough.
- That said, I don't think new tasks should be empty when created. The visual indicator could be an error border around the textfield (and maybe a message at the bottom). 
- Whether the user is more interested in viewing the checklist to tick things off, or to create new ones. If it's the former, then your workflow would be better since the text field won't be a distraction. But if it's the latter, it would take one extra click to be able to start inserting tasks.

hyj1204:
- I feel it is okay to discard when it is empty since nothing is entered. For the identical one, we can check the text the user is done entering the text(like after the 'return' key is clicked), if it has the identical one, we can pop up an alert window that says 'Option already exists'  and a 'Okay' button to close the window. 
- When adding a new checklist, maybe we can add some default options to help the user understand its usage rather than show nothing there, maybe like "Task one" and "Task two"?
- From my perspective, when an option is being created, users often enter or edit its details early on. Once the checklist is stable, they only want to edit the value(not the option) when clicking that box. I also feel it is better to separate the UI for **editing the value** from **editing the checklist option property**.

richardshiue:
Ok thanks I'll take this into consideration.

richardshiue:
To move things along, I'm going to merge this and we can iterate on the UX design later.

PR #3459, opened 2023-09-19
feat: go_router refactor and bottom navigation bar in mobile 

hyj1204:
### Feature Preview
1. Refactor all the auth pages' routes by the [go_router](https://pub.dev/packages/go_router) package:

- Currently, I only did some basic refactoring on the pages related to authentication, other parts still have `Navigator.push`. I use `go_router` package mainly for navigation on mobile. The authentication workflow is still the same as the previous code, the only difference is I add `context.go` to replace the previous page(same as `Navigator.pushReplacement`), so the splash screen will be replaced by home page after the user login. On the home page, if the user clicks 'back' (android back button), the app will quit instead of back to login/skipLoginPage. 
- Remove the routeName of `SplashScreen`. Due to it is the root page of the app, I set `'/'` as its routeName. 
- All the routes(using `go_router`) are registed in `appflowy_flutter/lib/startup/tasks/generate_router.dart`

2. Add bottom navigation bar on mobile.
I utilized the [StatefulShellRoute.indexedStack](https://pub.dev/documentation/go_router/latest/go_router/StatefulShellRoute/StatefulShellRoute.indexedStack.html) to build the bottom navigation bar. This approach enables each tab to retain its own state, as demonstrated in the following video.
Currently, every tab page is just some [example code](https://github.com/flutter/packages/blob/main/packages/go_router/example/lib/stateful_shell_route.dart) and I will implement them in the future. 

https://github.com/AppFlowy-IO/AppFlowy/assets/14248245/26aef070-ca76-4f7e-9b9d-633df8163ad4

3. For the icons used on mobile, I name them starting with **m**, because some icons look different as the desktop even though their purpose/name is similar. 

**go_router note:** 
When it comes to passing the arguments between the pages. I set `extra` to a `Map<String, dynamic>` to carry all the arguments, then get it from the `state` in `pageBuilder`.

- Pass arguments: 
Set argument names as constant(start with arg) on the screen page, then use them as the key in the `Map<String, dynamic>`. 
Pass this `Map<String, dynamic>` as `extra` to GoRoute.

```
    context.go(
            DesktopHomeScreen.routeName,
            extra: {
              DesktopHomeScreen.argKey: ValueKey(userProfile.id),
              DesktopHomeScreen.argUserProfile: userProfile,
              DesktopHomeScreen.argWorkspaceSetting: workspaceSetting,
            },
          );
```
Head part and constructor of `DesktopHomeScreen`
```
class DesktopHomeScreen extends StatefulWidget {
  static const routeName = '/DesktopHomeScreen';
  // arguments names to used in GoRouter
  static const argUserProfile = 'userProfile';
  static const argWorkspaceSetting = 'workspaceSetting';
  static const argKey = 'key';

  final UserProfilePB userProfile;
  final WorkspaceSettingPB workspaceSetting;
  const DesktopHomeScreen({
    super.key,
    required this.userProfile,
    required this.workspaceSetting,
  });
```

- Receive arguments in `pageBuilder`
``` 
GoRoute(
    path: DesktopHomeScreen.routeName,
    pageBuilder: (context, state) {
      final args = state.extra as Map<String, dynamic>;
      return CustomTransitionPage(
        child: DesktopHomeScreen(
          key: args[DesktopHomeScreen.argKey],
          userProfile: args[DesktopHomeScreen.argUserProfile],
          workspaceSetting: args[DesktopHomeScreen.argWorkspaceSetting],
        ),
        transitionsBuilder: _buildFadeTransition,
        transitionDuration: _slowDuration,
      );
    },
  );
```

This is my first time using the "go_router" package in the project. Please let me know if there are any issues or a better way to implement it.

---

#### PR Checklist

- [X] I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.


Xazin:
Oh wow, it looks good! I'll review this tomorrow, but good work! 🚀

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+        ),
+      ),
+      // Routes in both desktop and mobile
+      GoRoute(
+        path: SignInScreen.routeName,
+        pageBuilder: (context, state) {
+          return CustomTransitionPage(
+            child: SignInScreen(router: getIt<AuthRouter>()),
+            transitionsBuilder: _buildFadeTransition,
+            transitionDuration: Duration(
```
LucasXu0, line 31: "Define the duration as a constant value, and then you can reuse it in the code below.

```dart
            transitionDuration: _transitionDuration,

// ...


final _transitionDuration = Duration(
  milliseconds: (RouteDurations.slow.inMilliseconds).round(),
);

```

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+        pageBuilder: (context, state) => MaterialPage(
+          child: child,
+        ),
+      ),
+      // Routes in both desktop and mobile
+      GoRoute(
+        path: SignInScreen.routeName,
+        pageBuilder: (context, state) {
+          return CustomTransitionPage(
+            child: SignInScreen(router: getIt<AuthRouter>()),
```
LucasXu0, line 29: The `AuthRouter` has been injected into `getIt`. Is it necessary to pass it to the SignInScreen, or can we use getIt<AuthRouter>() directly within the SignInScreen widget?

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+          );
+        },
+      ),
+      GoRoute(
+        path: SkipLogInScreen.routeName,
+        pageBuilder: (context, state) {
+          return CustomTransitionPage(
+            child: SkipLogInScreen(
+              authRouter: getIt<AuthRouter>(),
+              authService: getIt<AuthService>(),
```
LucasXu0, line 43: Also here. Use `getIt` in the widget directly.

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+            ),
+            transitionsBuilder: _buildFadeTransition,
+            transitionDuration: Duration(
+              milliseconds: (RouteDurations.slow.inMilliseconds).round(),
+            ),
```
LucasXu0, line 48: https://github.com/AppFlowy-IO/AppFlowy/pull/3459/files#r1333062995

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+      ),
+      GoRoute(
+        path: EncryptSecretScreen.routeName,
+        pageBuilder: (context, state) {
+          final args = state.extra as Map<String, dynamic>;
+          return CustomTransitionPage(
+            child: EncryptSecretScreen(
+              user: args['user'],
+              key: args['key'],
+            ),
```
LucasXu0, line 60: Please don't hardcode the keys. It will be better to use class or const string instead.

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+            transitionsBuilder: _buildFadeTransition,
+            transitionDuration: Duration(
+              milliseconds: (RouteDurations.slow.inMilliseconds).round(),
+            ),
```
LucasXu0, line 64: https://github.com/AppFlowy-IO/AppFlowy/pull/3459/files#r1333062995

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+      GoRoute(
+        path: WorkspaceErrorScreen.routeName,
+        pageBuilder: (context, state) {
+          final args = state.extra as Map<String, dynamic>;
+          return CustomTransitionPage(
+            child: EncryptSecretScreen(
+              user: args['userFolder'],
+              key: args['error'],
+            ),
+            transitionsBuilder: _buildFadeTransition,
```
LucasXu0, line 77: Also here.

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+            transitionDuration: Duration(
+              milliseconds: (RouteDurations.slow.inMilliseconds).round(),
+            ),
+          );
+        },
+      ),
+      // Mobile only
+      // We use StatefulShellRoute to create a StatefulNavigationShell(ScaffoldWithNavBar) to access to multiple pages, and each page retains its own state.
+      StatefulShellRoute.indexedStack(
```
LucasXu0, line 104: Add a condition 'if (mobile) ...' to exclude them on the desktop platform.

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+          // The following code comes from the example of StatefulShellRoute.indexedStack. I left there just for placeholder purpose. They will be updated in the future.
+          // The route branch for the second tab of the bottom navigation bar.
+          StatefulShellBranch(
+            // It's not necessary to provide a navigatorKey if it isn't also
+            // needed elsewhere. If not provided, a default key will be used.
+            routes: <RouteBase>[
+              GoRoute(
+                // The screen to display as the root in the second tab of the
+                // bottom navigation bar.
+                path: '/b',
```
LucasXu0, line 139: Is it just for the demonstration? If no, please use a meaningful string.

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
```diff
@@ -0,0 +1,270 @@
...
+import 'package:flowy_infra/time/duration.dart';
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+GoRouter generateRouter(Widget child) {
+  return GoRouter(
+    navigatorKey: AppGlobals.rootNavKey,
+    initialLocation: '/',
+    routes: [
+      GoRoute(
```
LucasXu0, line 16: Defining each `GoRoute` as a small function would make the code more readable.

For example,

```dart
  GoRouter(
    routes: [
      _splashRoute,
     _signInRouter
    // ...
   ])

  GoRoute _splashRoute() { return .... }
  GoRoute _signInRouter() { return .... }
```

frontend/appflowy_flutter/lib/mobile/presentation/home/mobile_home_page.dart
```diff
@@ -0,0 +1,81 @@
...
+        UserProfilePB? userProfile;
+        snapshots.data?[0].fold(
+          (workspaceSettingPB) {
+            workspaceSetting = workspaceSettingPB as WorkspaceSettingPB?;
+          },
+          (error) => null,
+        );
+        snapshots.data?[1].fold((error) => null, (userProfilePB) {
+          userProfile = userProfilePB as UserProfilePB?;
+        });
```
LucasXu0, line 37: use fold the simplify the logic.

```suggestion
        final workspaceSetting = snapshots.data?[0].fold<WorkspaceSettingPB?>(
          (workspaceSettingPB) => workspaceSettingPB as WorkspaceSettingPB?,
          (_) => null,
        );
        final userProfile = snapshots.data?[1].fold<UserProfilePB?>(
          (userProfilePB) => userProfilePB as UserProfilePB?,
          (_) => null,
        );
```

frontend/appflowy_flutter/lib/mobile/presentation/mobile_bottom_navigation_bar.dart
```diff
@@ -0,0 +1,93 @@
...
+import 'package:go_router/go_router.dart';
+
+/// Builds the "shell" for the app by building a Scaffold with a
+/// BottomNavigationBar, where [child] is placed in the body of the Scaffold.
+class MobileBottomNavigationBar extends StatelessWidget {
+  /// Constructs an [MobileBottomNavigationBar].
+  const MobileBottomNavigationBar({
+    required this.navigationShell,
+    Key? key,
+  }) : super(key: key ?? const ValueKey<String>('MobileBottomNavigationBar'));
```
LucasXu0, line 12: It's my first time seeing this code. Is it really necessary to ensure that the key is not null?

frontend/appflowy_flutter/lib/mobile/presentation/mobile_bottom_navigation_bar.dart
```diff
@@ -0,0 +1,93 @@
...
+        showSelectedLabels: false,
+        showUnselectedLabels: false,
+        // Here, the items of BottomNavigationBar are hard coded. In a real
+        // world scenario, the items would most likely be generated from the
+        // branches of the shell route, which can be fetched using
+        // `navigationShell.route.branches`.
+        type: BottomNavigationBarType.fixed,
+        items: <BottomNavigationBarItem>[
+          BottomNavigationBarItem(
+            label: 'home',
```
LucasXu0, line 33: Remember to localize the label in the future.

frontend/appflowy_flutter/lib/user/presentation/router.dart
```diff
@@ -46,39 +38,27 @@ class AuthRouter {
...
-                name: DesktopHomeScreen.routeName,
-              ),
-              RouteDurations.slow.inMilliseconds * .001,
-            ),
+          context.go(
+            DesktopHomeScreen.routeName,
+            extra: {
+              'key': ValueKey(userProfile.id),
+              'userProfile': userProfile,
+              'workspaceSetting': workspaceSetting,
```
LucasXu0, line 60: Use a const key or a class instead.

frontend/appflowy_flutter/lib/user/presentation/router.dart
```diff
@@ -90,16 +70,13 @@ class AuthRouter {
...
-        ),
-        const RouteSettings(name: EncryptSecretScreen.routeName),
-        RouteDurations.slow.inMilliseconds * .001,
-      ),
+    // After log in,push EncryptionScreen on the top SignInScreen
+    context.push(
+      EncryptSecretScreen.routeName,
+      extra: {
+        'userProfile': userProfile,
+        'key': ValueKey(userProfile.id),
```
LucasXu0, line 78: Also here.

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
hyj1204: All the routes after MobileHomeScreen are just for demonstration. I added a TODO there.

frontend/appflowy_flutter/lib/mobile/presentation/mobile_bottom_navigation_bar.dart
hyj1204: There is no text shown on the navigation bar, but an Exception will be thrown if the label is none. I added the comments to address that.

frontend/appflowy_flutter/lib/mobile/presentation/mobile_bottom_navigation_bar.dart
hyj1204: Deleted

frontend/appflowy_flutter/lib/startup/tasks/generate_router.dart
hyj1204: I changed it to const since all we need is to avoid typos in this case.

LucasXu0 (approved):

LucasXu0:
LGTM. Merge after the CI passed.

LucasXu0:
there must be something that is causing the integration tests to fail, as they are still failing even after I merged the latest main branch.

hyj1204:
I found that theme mode can only be changed once, it stops updating somehow if I want to change it again. I going to spend some time to dig what's wrong with it.

LucasXu0:
> I found that theme mode can only be changed once, it stops updating somehow if I want to change it again. I going to spend some time to dig what's wrong with it.

You can simply set a breakpoint to verify whether the widgets are being rebuilt after changing the theme. I guess it may be related to the flutter `key`.

hyj1204:
In the previous code, `context.read<AppearanceSettingsCubit>` actually accesses the `context` which is inactive, and throws the Exception("Looking up a deactivated widget's ancestor is unsafe"). 

After trying multiple methods to solve the issue, I found that using the GlobalKey for the root navigator is able to access the previously rendered context.

image


I also added a redirect function in `_rootRoute `, so every time before navigating to the splash screen, we check if user has already logged in on the desktop. If so, it will push/go to DesktopHomeScreen directly rather than showing the unnecessary splash screen again.

image




Here is the final effect that changing the theme mode by pressing 'command/ctrl + shift + L' shortcut key.

https://github.com/AppFlowy-IO/AppFlowy/assets/14248245/394d7ffc-6d0a-460c-9a89-fad303e2c1bd

richardshiue:
> Here is the final effect that changing the theme mode by pressing 'command/ctrl + shift + L' shortcut key.

It works, but there's a first initial flash that's a bit distracting

richardshiue:
Here's another thing I've been thinking about. We can refactor this later, but shouldn't the routes be the same for mobile and desktop? That way when we implement deep linking we can jump to the right page using the same URL on all platforms.

hyj1204:
> Here's another thing I've been thinking about. We can refactor this later, but shouldn't the routes be the same for mobile and desktop? That way when we implement deep linking we can jump to the right page using the same URL on all platforms.

For mobile devices, the user experience (UX) is different compared to the desktop version because of the significant differences in usage scenarios and screen sizes. On mobile, the focus is on browsing and lightweight editing, and future features such as notifications and search. Thus, the page routes(except the authentication part) are completely different on mobile and desktop, plus they have different theme data. 

Currently, the shared part between desktop and mobile is the **workflow** for the authentication part(pages prior to home pages) and localization file, the rest parts (like ThemeData, Icons/SVGs, pages UI) are independent.

hyj1204:
> > Here is the final effect that changing the theme mode by pressing 'command/ctrl + shift + L' shortcut key.
> 
> It works, but there's a first initial flash that's a bit distracting

I spent some time trying to improve the flash effect, but I still haven't found a good solution yet. It seems that the colorTween is not working well when switching between theme modes.

LucasXu0:
https://github.com/AppFlowy-IO/AppFlowy/pull/3459/commits/5b10da31565ded3946a21d48f965851693731cfc

I cached the result from `generateRouter(widget.child)` to prevent unnecessary rebuilds.

### preview
https://github.com/AppFlowy-IO/AppFlowy/assets/11863087/57da9640-1714-4bc1-97b3-5eeca0f51aee

LucasXu0:
> In the previous code, context.read<AppearanceSettingsCubit> actually accesses the context which is inactive, and throws the Exception("Looking up a deactivated widget's ancestor is unsafe").

After preventing the unnecessary rebuilds, we can now get the `AppearanceSettingsCubit` from the current context; IMO, there is no need to use `AppGlobals.rootNavKey`.

PR #3495, opened 2023-09-22
fix: some issues with the end date

richardshiue:
- color of the unselected days in darkmode. It should blend into the background now
- a day selected -> toggle end date -> click on another one . This should immediately set the range
- reorganized the editor to not rebuild everything in the top-level bloc builder. Without this, toggling include time will reset the focused day to this month. (sorry Mathias)

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/date_cell/date_cal_bloc.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/date_cell/date_cal_bloc.dart
@@ -40,6 +40,8 @@ class DateCellCalendarBloc
           didReceiveCellUpdate: (DateCellDataPB? cellData) {
             final (dateTime, endDateTime, time, endTime, includeTime, isRange) =
                 _dateDataFromCellData(cellData);
+            final newEnd =
+                isRange == state.isRange && isRange ? endDateTime : null;
             emit(
               state.copyWith(
                 dateTime: dateTime,
@@ -49,7 +51,7 @@ class DateCellCalendarBloc
                 includeTime: includeTime,
                 isRange: isRange,
                 startDay: isRange ? dateTime : null,
-                endDay: isRange ? endDateTime : null,
+                endDay: newEnd,
               ),
             );
           },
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/date_cell/date_editor.dart
```diff
@@ -100,55 +100,32 @@ class _CellCalendarWidgetState extends State<_CellCalendarWidget> {
...
-            itemBuilder: (BuildContext context, int index) => children[index],
-            padding: const EdgeInsets.only(top: 18.0, bottom: 12.0),
-          );
-        },
+      child: ListView.builder(
+        shrinkWrap: true,
+        controller: ScrollController(),
+        itemCount: children.length,
+        itemBuilder: (BuildContext context, int index) => children[index],
+        padding: const EdgeInsets.only(top: 18.0, bottom: 12.0),
```
LucasXu0, line 128: What are there differences between using the Column widget and using ListView.builder when shrinkWrap is set to true here?

richardshiue: Performance wise, nothing. they both need to evaluate everything. When I wrote this a long time ago, I think the min height on the window was a bit lower, so I had to make sure it wouldn't overflow. But now it doesn't matter since it won't overflow at all. Should I change to Column?

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/date_cell/date_cal_bloc.dart
```diff
@@ -40,6 +40,8 @@ class DateCellCalendarBloc
           didReceiveCellUpdate: (DateCellDataPB? cellData) {
             final (dateTime, endDateTime, time, endTime, includeTime, isRange) =
                 _dateDataFromCellData(cellData);
+            final newEnd =
```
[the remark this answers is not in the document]
richardshiue: Much better, thanks

frontend/appflowy_flutter/lib/plugins/database_view/widgets/row/cells/date_cell/date_editor.dart
LucasXu0: yes. it would be better to use the column here since the shrinkwrap is set to true and the widget here isn't scrollable.

Issue #3543, opened 2023-09-26
[Bug] `Change cover` button does not show correct cursor


Xazin:
### Bug Description

When you add a Cover to a document, and hover on the cover, the actions are shown in the bottom right of the Cover.

When moving mouse to hover over `Change cover`, the mouse should be of type `SystemCursors.click`, but isn't.

### How to Reproduce

1. Navigate to/create a document
2. Add a Cover
3. Hover on Change Cover button

### Expected Behavior

1. Navigate to/create a document
2. Add a Cover
3. Hover on Change Cover button
4. Cursor changes to a Click

### Operating System

All

### AppFlowy Version(s)

0.3.3

PR #3548, opened 2023-09-27
fix: on hover change_cover button correct mouse cursor

1akhanBaheti:
### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/94619783/c0fda096-aa81-4bba-979b-680df48dbab3

fixes https://github.com/AppFlowy-IO/AppFlowy/issues/3543

#### PR Checklist


1akhanBaheti:
when on_pressed was passed as null, hover effect & cursor was not changing.

richardshiue (changes requested):
Hey there @1akhanBaheti, thanks for the contribution. This fixes the issue, but I think there is a better solution. The onPressed parameter to RawMaterialButton should be nullable. I think it's better to investigate why the argument passed into the changecover button is null. It opens a popup right after all.

richardshiue:
Oh great. Yes exactly, you can change the popover trigger action to None and then trigger a PopoverController in the onPressed callback.

Xazin (approved):

richardshiue (approved):
LGTM

PR #3553, opened 2023-09-28
refactor: ui improvement in appearance

MayurSMahajan:
The PR achieves the following:
 - Adds dividers to every option in Appearance Settings
 - Moves the Tooltip closer to the ThemeMode button
 - Tooltip text now says Change Theme Mode, which is a descriptive action rather than plain Theme Mode
 - Removes unused context parameter from the tooltipMethod.

### Feature Preview

image

---

#### PR Checklist


LucasXu0:
@MayurSMahajan Actually, the divider I used is for grouping the cells, so there's no need to include additional dividers.

PR #3574, opened 2023-10-02
fix:Updating UI

SUNIL-RAGHU:
Added vertical padding because the text was appearing too close to the top.

Before:
Before Screenshot

After:
After Screenshot

In this commit, I adjusted the content padding to include both horizontal and vertical padding in order to improve the visual appearance. The text was previously too close to the top, which could lead to readability issues. This change enhances the overall user experience by providing better spacing.

- Modified contentPadding in the relevant widget.
- Included 'horizontal' and 'vertical' parameters with appropriate values.

Screenshots have been included for visual comparison between the original and modified layouts.

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/openai/widgets/auto_completion_node_widget.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/openai/widgets/auto_completion_node_widget.dart
@@ -168,7 +168,6 @@ class _AutoCompletionBlockComponentState
     return FlowyTextField(
       hintText: LocaleKeys.document_plugins_autoGeneratorHintText.tr(),
       controller: controller,
-      maxLines: 3,
       focusNode: textFieldFocusNode,
       autoFocus: false,
     );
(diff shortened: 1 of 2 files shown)


Xazin:
To me it seems it's still preferential to the top, can we make it centered?

SUNIL-RAGHU:
yes! will change it then

richardshiue:
Adding vertical padding, that I can get behind. But why alter the horizontal padding as well?

Looking at the code, it seems like the max line argument is 3 here. I suggest instead that you also remove the constraints altogether in FlowyTextField and set isDense to true.

SUNIL-RAGHU:
Before Modification:
Screenshot 2023-10-03 at 11 03 10 PM

After Modification:
Screenshot 2023-10-03 at 11 12 51 PM

Changes Made:

Updated the TextField widget behavior to allow for the input of multiple lines of text.
Implemented dynamic expansion of the widget to ensure clear visibility of entered text, even when more than one line is used.

Benefits:

Enhances user experience by accommodating longer or multiline inputs without sacrificing visibility or usability.
Provides a more flexible and intuitive text input experience for users.

richardshiue:
I don't know where the 13 symmetrical vertical padding comes from but looks good to me

SUNIL-RAGHU:
> Added vertical padding because the text was appearing too close to the top.

> Before: Before Screenshot

skydiver1980 (approved):

PR #3637, opened 2023-10-06
fix: image block issues

LucasXu0:
### Feature Preview

- [x] remove the tooltip of the theme mode, and add a description under the theme mode text.
- [x] Add padding between unsplash results and search field 
- [x] cannot use ESC to close the image picker dialog - only when navigating to embed link or unsplash tab
- [x] No longer able to trigger the menu after selecting image
- [x] it shouldn’t require Enter to search an image - (Add a search button)
- [x] Selecting an image from unsplash makes the dialog flicker as it disappear
- [x] add a loading status when querying from the image service
- [x] Selecting the image block is strange 

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/packages/appflowy_popover/lib/src/popover.dart
+++ b/frontend/appflowy_flutter/packages/appflowy_popover/lib/src/popover.dart
@@ -48,6 +48,11 @@ enum PopoverDirection {
   custom,
 }
 
+enum PopoverClickHandler {
+  listener,
+  gestureDetector,
+}
+
 class Popover extends StatefulWidget {
   final PopoverController? controller;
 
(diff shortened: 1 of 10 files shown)


frontend/appflowy_flutter/packages/appflowy_popover/lib/src/popover.dart
```diff
@@ -78,11 +83,18 @@ class Popover extends StatefulWidget {
 
   final bool asBarrier;
 
+  /// The widget that will be used to trigger the popover.
+  ///
+  /// Why do we need this?
+  /// Because if the parent widget of the popover is GestureDetector,
+  ///  the conflict won't be resolve by using Listener, we want these two gestures exclusive.
+  final PopoverClickHandler clickHandler;
```
LucasXu0, line 91: cc @richardshiue

richardshiue: Yeah this is an issue, our popover component needs more work. onTapdown isn't a very good UX.

The problem with using the gesture detector is that when you use it, the child button cannot specify an onTap or onPressed handler, not even `() {}` because it will take precedence. But setting it to null will disable the button's hover effects. (I was testing the `FlowyTextButton` in `SettingsLanguageView`) I think this can cause some confusion 

What if we change it to `onTapUp`?

LucasXu0: Why do we need to change it to `onTapUp`?

```dart
  Widget _buildClickHandler(Widget child, VoidCallback handler) {
    switch (widget.clickHandler) {
      case PopoverClickHandler.listener:
        return Listener(
          onPointerDown: (_) => handler(),
          child: child,
        );
      case PopoverClickHandler.gestureDetector:
        return GestureDetector(
          onTap: handler,
          child: child,
        );
    }
  }
```

PR #3550, opened 2023-09-27
feat: dandelion theme improvement

hyj1204:
### Feature Preview
To close #3268  and improve the following UI. 
1. Add a tooltip and set the hover color in the upload theme button as same as other icon buttons around.
Pasted Graphic 3

2. Set new page icon color to primary color
image
image
image
image
image
image

---

3. I modified the following part from the draft design under the issue discussion. 
For light mode:

- I set all the hover colors to yellow style rather than green or yellow in different parts. 
- Only the sidebar background has a green color, the rest part uses a yellow style color. 
- I didn't change the border color(divider color) because it is also linked to the divider in the settings page, which looks weird if we add a colorful color there. 

https://github.com/AppFlowy-IO/AppFlowy/assets/14248245/e41a4dd2-ac5f-4f6e-8d28-b91c093c4950

There are two options for the sidebar color I feel it would be good too. Just let me know which one you think is the best: 
**Option A: A light green color (f3fae3)**
image
**Option B: Use a light yellow instead of green color(faf0c8)**
image

For Dark mode:

- I used only yellow color there to keep it simple.
- I didn't change the border color for the same reason as the light mode. 

https://github.com/AppFlowy-IO/AppFlowy/assets/14248245/a2208b8c-588b-4f51-923c-31f7897a62b9

---

#### PR Checklist

- [x] My code adheres to the [AppFlowy Style Guide](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/submitting-code/style-guides)
- [x] I've listed at least one issue that this PR fixes in the description above.
- [x] I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.
- [x] All existing tests are passing.

PR #3666, opened 2023-10-10
feat: add lemonade theme color

hyj1204:
I made option B in #3550 become a new color theme called `Lemonade`. 

I changed the sidebar color to a deep green color in the dandelion theme to distinguish the dark mode for dandelion and lemonade. It is the only difference between these two dark modes so far. 

### Feature Preview
Lemonade theme in light mode:

image

Lemonade theme in dark mode:
image

Dandelion theme in light mode(no change in this PR):
image

Dandelion theme in dark mode(only sidebar color changed)
image

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/packages/flowy_infra/lib/colorscheme/colorscheme.dart
+++ b/frontend/appflowy_flutter/packages/flowy_infra/lib/colorscheme/colorscheme.dart
@@ -6,6 +6,7 @@ import 'package:freezed_annotation/freezed_annotation.dart';
 import 'default_colorscheme.dart';
 import 'dandelion.dart';
 import 'lavender.dart';
+import 'lemonade.dart';
 
 part 'colorscheme.g.dart';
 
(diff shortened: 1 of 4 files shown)


MayurSMahajan:
The new theme looks amazing! But Dandelion as a name for a green theme does not sound right. Maybe we should rename it to something whose natural color is green. Now that we have a individual yellow theme called lemonade, we must remove Dandelion or rename it to something more descriptive.

hyj1204:
> The new theme looks amazing! But Dandelion as a name for a green theme does not sound right. Maybe we should rename it to something whose natural color is green. Now that we have a individual yellow theme called lemonade, we must remove Dandelion or rename it to something more descriptive.

The main color of the Dandelion is still yellow like the Lemonade one. The green color is a secondary color in Dandelion, so I put it in the sidebar. It was inspired by some images like the following images showing dandelion flowers on a green lawn. 
That's why I feel it makes sense, but I'm open to better suggestions for the name or color combination of Dandelion.
image

MayurSMahajan:
Oh, now I get it. Sorry I did not think like that, dandelion on a green lawn, makes sense. I think we should not change the name of it.

PR #3673, opened 2023-10-11
fix: fix integration tests

richardshiue:
### Feature Preview

---

#### PR Checklist


Xazin:
The tests are failing because we don't add the event to check the initial screen size and thus had it collapsed by default.

This just has it expanded by default no?

Xazin:
We can consider mocking the `window_manager` channel in `initializeAppFlowy` (tests). And have the creation of the HomeSettingBloc be something like:
```
return HomeSettingBloc(
  ...
)..add(
    HomeSettingEvent.checkScreenSize(
      (await windowManager.getSize()).width,
    ),
  );
```

Xazin:
I believe tests will still fail unless we mock the window_manager.

richardshiue:
> This just has it expanded by default no?

Yes. This was a bad fix

richardshiue:
> This just has it expanded by default no?

Yes this was a bad fix.

Xazin (approved):

Issue #3630, opened 2023-10-05
[Bug] The appearance of slash menu didn't change along with theme mode changes


hyj1204:
### Bug Description

When we changed the theme or theme mode, the slash menu's appearance didn't change.

### How to Reproduce

1. Open the slash menu
2. Change theme or theme mode
3. Open the slash menu again, it shows the same appearance as the previous theme. 

### Expected Behavior

The slash menu shows the updated/current theme and theme mode color. 

### Operating System

macOS

### AppFlowy Version(s)

0.3.4

### Screenshots

https://github.com/AppFlowy-IO/AppFlowy/assets/14248245/42b836de-e90d-4ec4-a7e5-6822b873831f

lielfr:
Hi, I'd like to work on this one if possible 😄

MayurSMahajan:
@hyj1204 can assign you, if she is not already working on it.

hyj1204:
> Hi, I'd like to work on this one if possible 😄

Here you go

lielfr:
hi again, I think I managed to fix the bug. However, I'm having some trouble running the integration tests, and would like some help with that (it's probably a good idea to have at least one new test for this)

PR #3706, opened 2023-10-15
fix: The appearance of slash menu didn't change along with theme mode changes (#3630)

lielfr:
What (I think) happened was that the theme was not being updated based on the context, so I moved the `showSlashMenu` function to accept the current context and call the builder from within. Not sure if it is the best way to do it, but seems to work:

https://github.com/AppFlowy-IO/AppFlowy/assets/360928/4a5f1bdb-95c3-482d-bd41-c3183648d290

### Feature Preview

fixes https://github.com/AppFlowy-IO/AppFlowy/issues/3630

---

#### PR Checklist


hyj1204 (changes requested):

hyj1204 (approved):
LGTM

MayurSMahajan (approved):

PR #3730, opened 2023-10-19
feat: save card orders

richardshiue:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/src/services/database/database_editor.rs
+++ b/frontend/rust-lib/flowy-database2/src/services/database/database_editor.rs
@@ -942,13 +942,26 @@ impl DatabaseEditor {
         let mut row_changeset = RowChangeset::new(row_detail.row.id.clone());
         let view = self.database_views.get_view_editor(view_id).await?;
         view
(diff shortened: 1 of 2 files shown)


frontend/rust-lib/flowy-database2/src/services/database/database_editor.rs
```diff
@@ -942,14 +942,27 @@ impl DatabaseEditor {
...
+          let row_details = self.get_rows(view_id).await?;
+          row_details
+            .last()
+            .map(|row_detail| row_detail.row.id.clone())
+        };
+
+        if let Some(row_id) = to_row {
+          self.move_row(view_id, from_row, row_id).await;
+        }
+
```
Xazin, line 965: How about using `unwrap_or_else` here?
```suggestion
        let row_id = to_row.unwrap_or_else(|| {
          let row_details = self.get_rows(view_id).await?;
          row_details
            .last()
            .map(|row_detail| row_detail.row.id.clone())
        });

        self.move_row(view_id, from_row, row_id).await;

```
I guess there could be a disconnect because of `row_details.last()` though.

Issue #3703, opened 2023-10-15
[FR] Enable users to directly rename a stack by clicking on the stack name and editing the name


annieappflowy:
### Description

As described in the title

### Impact

Kanban users

annieappflowy:
related #1145

annieappflowy:
<img width="311" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/12026239/1bd706a8-d8ff-4f50-936e-43fce0dfa9d7">

supported: select, multi-select

PR #3781, opened 2023-10-25
Feat: rename stack inline

Xazin:
Enables renaming a Column/Stack inline, however in its current version only for Single- and Multi-select groups.

Relates: #3703 

#### PR Checklist


richardshiue (approved):

appflowy:
@richardshiue @Xazin It would be better if we could add the test for `DatabaseEvent::UpdateGroup` in the event-integrate crate.

PR #3790, opened 2023-10-26
test: rename group test

appflowy:


Xazin:
We can delete this test in turn then `frontend/rust-lib/event-integration/tests/database/local_test/test.rs:739`

richardshiue (approved):

PR #3813, opened 2023-10-27
fix: board unit test fix

richardshiue:
the ungrouped-item-stack is now positioned last, so tests need to be adjusted

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/test/bloc_test/board_test/create_card_test.dart
+++ b/frontend/appflowy_flutter/test/bloc_test/board_test/create_card_test.dart
@@ -20,9 +20,9 @@ void main() {
     )..add(const BoardEvent.initial());
     await boardResponseFuture();
 
-    final groupId = boardBloc.state.groupIds.first;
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/test/bloc_test/board_test/create_card_test.dart
```diff
@@ -20,16 +20,16 @@ void main() {
     )..add(const BoardEvent.initial());
     await boardResponseFuture();
 
-    final groupId = boardBloc.state.groupIds.first;
+    final groupId = boardBloc.state.groupIds.last;
```
Xazin, line 23: Instead of a fixed index, I feel it more appropriate to take the group which `isDefault`, since that will work no matter if you change the ordering. What do you think?

richardshiue: Good point, though I think part of the test is to make sure that that stack is the last group as well

Issue #3010, opened 2023-07-15
[Bug] changing group, field value view not updating


zoli:
### Bug Description

On changing card group sometimes the card data (field) doesn't get updated immediately.

### How to Reproduce

Change a card group then open it see if the shown value is correct and equal to the current card group.

### Expected Behavior

The shown field value should get updated and be equal to card group.

### Operating System

Linux

### AppFlowy Version(s)

0.2.6

### Additional Context

This problem first got mentioned in https://github.com/AppFlowy-IO/AppFlowy/pull/2641.

zoli:
I would line to work and resolve this.

zoli:
This is similar to #2926, `move_group_row` doesn't fire `DidUpdateViewRows` it only fires `DidUpdateGroupRow`. This time I'm not sure, should the frontend listen to `DidUpdateGroupRow` (`updated_rows`) aside `DidUpdateViewRows` or the backend should fire both `DidUpdateGroupRow` and `DidUpdateViewRows`?

Also, IMHO the current flowy-database2 is bug-prone to this type of not sending event bugs. It's not enough narrowed to for example only update cell through one channel. 
Again for example I'm not sure but it seems possible to use `update_cell` for moving a row between groups.

@appflowy

zoli:
Just wanted to note this issue still exists.

zoli:
This is almost (I will add some notes later) fixed by https://github.com/AppFlowy-IO/AppFlowy/pull/3730.

PR #3833, opened 2023-10-30
fix: update auto-update fields on change group

zoli:
auto-update fields should get updated when moving rows from one group to another. They were getting updated only on changing row properties directly.
Also added integration test for changing row group related to #3010.

I was working on a fix for #3010, @richardshiue PR #3730 fixed part of it. This will add integration-test for #3010 and also fix auto-update fields not getting updated when changing row group.

#### PR Checklist


zoli:
I also wanted to add some notes on the issue continuing from [my comment](https://github.com/AppFlowy-IO/AppFlowy/issues/3010#issuecomment-1637196128) on the issue.

IMO the root of problem is not calling one of these two events `DidUpdateViewRows` and `DidUpdateGroupRow`. I first discussed this with @appflowy and tried to merge these two events into single event (using `GroupRowsNotificationPB` for both) but then I figured out its not possible because updating a row has different meaning for a row view and a group. For a row view means for example a row property is updated and `updated_rows` gets filled in `RowsChangePB` and for a group changing row group (which it means updating a row property) means deleting row from one group and inserting it to another `inserted_rows` and `deleted_rows` gets filled in `GroupRowsNotificationPB`.
My next attempt was funneling so that `move_group_row` calls `update_cell`, So for both changing row group and updating row `update_cell` gets called. I applied this and it seems to work fine. But the `move_row` method of `GroupCustomize` trait should change so that it only fills a `RowChangeset` no need to `Vec<GroupRowsNotificationPB>` anymore. `update_cell` will create `Vec<GroupRowsNotificationPB>` from the `RowChangeset` with `did_update_group_row` in `v_did_update_row`.
I didn't apply the last attempt changes in this PR.

richardshiue:
Hi @zoli you raise an interesting point. I think though, maybe a row's last modified timestamp shouldn't change if it's just being reordering within the same group, much like how reordering rows in a grid doesn't update it either.

zoli:
> I think though, maybe a row's last modified timestamp shouldn't change if it's just being reordering within the same group, much like how reordering rows in a grid doesn't update it either.

Hi @richardshiue, I agree with you. What I have changed is updating last modified timestamp when a row's group has changed not reordering. From what I understand about the events. `MoveGroupRow` event is fired when the row's group has changed and `MoveRow` is fired when only the order has changed. This PR adds updating last modified on `MoveGroupRow` not `MoveRow`.

richardshiue:
I believe that the `MoveRow` event is used in the grid view. In the board view, the `MoveGroupRow` event is used for both in-group reordering and cross-group reordering. It's just dependent on whether `to_group_id` is the same as `from_group_id`.

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/dd5aa6e3-9445-418f-955e-a0f8c7d152d5

(seems like there's also a bug with the last modified not immediately updating in the card as well)

zoli:
> I believe that the `MoveRow` event is used in the grid view.

Thanks, I will fix it according what you said.

> seems like there's also a bug with the last modified not immediately updating in the card as well.

I will take a look at this.

zoli:
@richardshiue  I think both points are resolved now.

zoli:
The tests are failing. I will take a look.

zoli:
I think everything is good now.

skydiver1980 (approved):

zoli:
@richardshiue Just a gentle reminder to review this.

frontend/rust-lib/flowy-database2/src/services/database_view/view_editor.rs
```diff
@@ -332,6 +320,24 @@ impl DatabaseViewEditor {
     }
   }
 
+  pub async fn v_get_row_group(&self, row_id: RowId) -> Option<String> {
```
richardshiue, line 323: This might not return the correct value if FieldType is MultiSelect because the row might be in multiple groups. This is an edge case though and it might be wise to disable dragging on multiselect. Alternatively, pass in the from_group_id from the frontend so that this function is no longer necessary.

frontend/rust-lib/flowy-database2/src/services/database/database_editor.rs
```diff
@@ -944,17 +944,25 @@ impl DatabaseEditor {
             .last()
             .map(|row_detail| row_detail.row.id.clone())
         };
+        if let Some(row_id) = to_row.clone() {
+          self.move_row(view_id, from_row.clone(), row_id).await;
+        }
 
-        if let Some(row_id) = to_row {
-          self.move_row(view_id, from_row, row_id).await;
+        if from_group.is_some() && from_group.clone().unwrap() == to_group.to_string() {
```
richardshiue, line 951: from_group.map(|group| group == *to_group).unwrap_or(false)

zoli: Passing from group id so there is no need for this anymore.

frontend/rust-lib/flowy-database2/src/entities/database_entities.rs
```diff
@@ -162,11 +162,15 @@ pub struct MoveGroupRowPayloadPB {
 
   #[pb(index = 4, one_of)]
   pub to_row_id: Option<String>,
+
+  #[pb(index = 6)]
```
richardshiue, line 166: nit: use 5

richardshiue (approved):

frontend/rust-lib/flowy-database2/src/entities/database_entities.rs
zoli: Done.

appflowy:
> I also wanted to add some notes on the problem continuing from [my comment](https://github.com/AppFlowy-IO/AppFlowy/issues/3010#issuecomment-1637196128) on the issue.
> 
> IMO the root of problem is not calling one of these two events `DidUpdateViewRows` and `DidUpdateGroupRow`. I first discussed this with @appflowy and tried to merge these two events into single event (using `GroupRowsNotificationPB` for both) but then I figured out its not a good idea because updating a row has different meaning for a row view and a group. For a row-view updating row means for example a row property is updated and `updated_rows` gets filled in `RowsChangePB` and for a group, changing row group (which it means updating a row property) means deleting row from one group and inserting it to another. `inserted_rows` and `deleted_rows` gets filled in `GroupRowsNotificationPB`. My next attempt was funneling so that `move_group_row` calls `update_cell`, So for both changing row group and updating row `update_cell` gets called. I applied this and it seems to work fine. But the `move_row` method of `GroupCustomize` trait should change so that it only fills a `RowChangeset` no need to return `Vec<GroupRowsNotificationPB>` anymore. `update_cell` will create `Vec<GroupRowsNotificationPB>` from the `RowChangeset` with `did_update_group_row` in `v_did_update_row`. I didn't apply the last attempt changes in this PR.

@zoli, thank you for providing clarity on this matter. Yes, feel free to test this approach in a separate PR (Pull Request).

appflowy:
LGTM

PR #3891, opened 2023-11-06
build: added the option to use static libraries

zoli:
### Feature Preview
Added `USET_STATIC_LIB` environment variable to build AppFlowy SDK (`sdk-dev-build` task) using openssl, zstd, rocksdb static libraries, and eliminating their compile time.
I got static libraries from "rust-lib/target" directory (e.g. "frontend/rust-lib/target/x86_64-unknown-linux-gnu/debug/build/zstd-sys-2ceef7f8e16ac49d/out/" for zstd) which were built with `cargo make --profile development-linux-x86_64 appflowy-core-dev` command. I created a compressed file from all of them. The compressed file should get generated for each platform separately.

The result on my laptop was about 10-13 minutes saving build time for `appflowy-core-dev` task after `flutter_clean` task (the download time of the static libraries was not included).

Had some related notes and discussions before here https://github.com/AppFlowy-IO/AppFlowy/pull/3519.

#### PR Checklist


zoli:
This PR is not completed yet. Currently, I've uploaded Linux platform static libraries to my server, They should get uploaded to the AppFlowy server or maybe somewhere else. 
Also, static libraries are available only for Linux platform I need help for packaging them for other platforms.

zoli:
If you want to check this out on Linux, just add `USE_STATIC_LIB=1` before running `appflowy-core-dev` or `appflowy-dev` task.

zoli:
I asked this [before](https://github.com/AppFlowy-IO/AppFlowy/pull/3519#issuecomment-1791454939) but I want to ask it again here: 
I was wondering if the flowy_codegen result (dart, protobuf, ts) also depends on user OS or STH like that? Because if it doesn't depend on specific platform maybe we can save a couple of minutes by committing proto files or at least enabling the cargo make task to not generate proto files each time.

LucasXu0:
If you choose to package the built-in libraries for OpenSSL, Zstd, and RocksDB, we have to update them every time they get updated.

zoli:
> If you choose to package the built-in libraries for OpenSSL, Zstd, and RocksDB, we have to update them every time they get updated.

Yeah. Every time the checksum changes in Cargo.lock file. It can get automated with a workflow probably.

zoli:
I packaged MacOS static libs and tested it successfully (in my fork repo with github workflow). Also packaged for Windows but I'm not familiar with windows scripting (for checking USE_STATIC_LIBS flag and setting environment variables).
The packaging is easy just run "appflowy-core-dev" task without using prebuilt static libraries then they will get built inside "rust-lib/target" and you can copy and package them from there.
The package creation can get automated with a workflow but still, there is a need for a place to upload them. They can't get uploaded as release assets because the package version of rocksdb, zstd, openssl can get changed in a non-release commit.

Where should I upload these?

zoli:
I added `USE_STATIC_LIB` to flutter_ci workflow just for the sake of testing. There is no need to use prebuilt static libraries when we rust-cache in workflow.

zoli:
Sorry but in mac os there seems a problem with ZSTD static library. I will look into it.

zoli:
Fixed Mac OS problem, Striped static libraries Linux from debug information, and now the zip file size is 27 MB (140 before strip).

LucasXu0:
> Every time the checksum changes in "Cargo.lock" file. It can get automated with a workflow probably.

Does it include in this PR?

zoli:
> Does it include in this PR?

No, first I need you to specify where to upload the packaged static libraries.

skydiver1980 (approved):

LucasXu0:
Since the benefit from this optimization may lead to more build issues, I will close it. We can seek a better solution in the future if build time becomes a significant problem. Thanks for your effort. @zoli

Issue #2734, opened 2023-06-07
[FR] Group settings


zoli:
### Description

To set `hide_empty` property and group conditions (e.g. Group by date with month condition).

### Impact

All of the users, They can hide empty groups and for future enables toggle display for each group.

zoli:
I would like to take on this.

cc: @annieappflowy

annieappflowy:
Assigned! Thanks

annieappflowy:
We can hide / show any groups on Kanban
the feature proposed here is not yet completed.

zoli:
> We can hide / show any groups on Kanban the feature proposed here is not yet completed.

Its hide / show plus condition. The first (hide / show) is done and the second will get resolved if #3949 PR. Its pending on review.

PR #3949, opened 2023-11-16
feat: enable group condition

zoli:
As part of #2734 this enables group configuration. So grouping by a field can also have more options such as group condition. For example when grouping by a date field type the user can choose to group by day, month, week, etc.

### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/3286303/32e75359-6b47-4c28-888c-265894d27935

#### PR Checklist


skydiver1980 (approved):

richardshiue:
My biggest concern with this PR is the condition inside GroupSettings. Using an int to store the condition (which is why I guess the DateCondition is removed?) leads to confusion between different field types. I think it would be better to specify the enum type in each`GroupCustomize` implementation.

frontend/appflowy_flutter/lib/plugins/database_view/application/setting/group_bloc.dart
```diff
@@ -92,8 +106,10 @@ class DatabaseGroupEvent with _$DatabaseGroupEvent {
   const factory DatabaseGroupEvent.initial() = _Initial;
   const factory DatabaseGroupEvent.setGroupByField(
     String fieldId,
-    FieldType fieldType,
-  ) = _DatabaseGroupEvent;
+    FieldType fieldType, {
+    int? condition,
+    bool? hideEmpty,
+  }) = _DatabaseGroupEvent;
```
richardshiue, line 112: You can use optional positional parameters here with `[...]`to maintain consistency

zoli: Done.

zoli:
> But I think the int should only be seen when converting to and from PBs, and when reading/saving from Collab. The `GroupSettings` shouldn't deal with a single int and the frontend looks like it's dealing with the base ProtobufEnum.

I don't get your point exactly. Do you mean to deal with it like `filters` which stores the data as `Vec<u8>`? If yes I don't know how it's different from storing an int.
Personally, I like avoiding match patterns based on field type and creating struct based on each case when its possible.

PR #3963, opened 2023-11-19
feat: update simple Chinese translation

slhmy:
🐸 Tranlated mostly by **inlang** but I found there is some mis-update there if I work completely with it !

Anyway, I filled most of the simple Chinese translation and fixed some obvious mistakes (old / new mistakes may still remain in some place although I have tried to re-check it...)

### Feature Preview

---

#### PR Checklist


slhmy:
@appflowy 👋 Hi, I think I need approval to enable the remaining workflows, thanks！

PR #3996, opened 2023-11-23
feat: custom server url in application

appflowy:
1. Add support for changing the cloud type within the application.
2. Enable modification of the server URL within the application.

This PR aims to eliminate the step of building the AppFlowy application during self-hosting.


appflowy:
image
image

PR #4045, opened 2023-11-29
fix: date group_id generation logic

richardshiue:
initially discovered in `flowy-database2::tests::database::group_test::date_group_test::group_by_date_test` in California timezone any time around 5 PM local time.

Group building and determining the group of a row should only be done w.r.t the local time zone, not the time zone of the field.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/linux/flutter/dart_ffi/binding.h
+++ b/frontend/appflowy_flutter/linux/flutter/dart_ffi/binding.h
@@ -3,7 +3,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-int64_t init_sdk(char *path);
+int64_t init_sdk(char *data);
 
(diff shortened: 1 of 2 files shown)


zoli (commented):
IMO the timezone should be counted in determining which group the cell should go to, but in determining group id and name it should use the local timezone. Here you are excluding timezone from both.

richardshiue:
Hi Zoli, could you help me understand why timezone should be taken into consideration when determining which group the cell should go to? If I have an event that's scheduled for Dec 6 in a particular timezone, but in my local timezone it would be on Dec 5, I'd argue that that event should still show up in the Today column if today is the 5th, otherwise I would miss the event.

zoli:
Hi Richard, Now I see I've got it wrong. In everywhere the time should be calculated by local timezone.

PR #4067, opened 2023-12-02
chore: optimize recent page font size

richardshiue:
| iOS | Android |
|--------|--------|
| Screenshot 2023-12-02 at 11 26 25 AM | Screenshot_20231202-112638 | 

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/mobile/presentation/home/recent_folder/mobile_recent_view.dart
+++ b/frontend/appflowy_flutter/lib/mobile/presentation/home/recent_folder/mobile_recent_view.dart
@@ -11,8 +11,8 @@ import 'package:appflowy_backend/dispatch/dispatch.dart';
 import 'package:appflowy_backend/protobuf/flowy-document2/protobuf.dart';
 import 'package:appflowy_backend/protobuf/flowy-folder2/view.pb.dart';
 import 'package:appflowy_editor/appflowy_editor.dart';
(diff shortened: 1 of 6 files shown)


frontend/appflowy_flutter/lib/mobile/presentation/home/recent_folder/mobile_recent_view.dart
```diff
@@ -125,16 +125,13 @@ class _MobileRecentViewState extends State<MobileRecentView> {
...
-                width: double.infinity,
-                padding: const EdgeInsets.only(
-                  left: 8.0,
-                  top: 14.0,
-                  right: 8.0,
-                ),
-                child: FlowyText(
+                width: widget.height,
+                padding: const EdgeInsets.fromLTRB(8, 16, 8, 2),
+                child: AutoSizeText(
```
Xazin, line 130: I used to use this package long ago, when I first started developing in Flutter back in 2021.

AutoSizeText is quite bad, it can cause performance issue with overreliance, also, for this one simple problem, can we really not find a more suitable solution than to bring in this package?

richardshiue: I mean I locked it down to only try two font sizes so I don't think it will be that bad. I've switched to another version with IntrinsicHeight.

PR #4080, opened 2023-12-04
fix: persist group order on reopenning AppFlowy

zoli:
Currently if you change groups order, on restarting AppFlowy they will go back to what they were before.
This was because merge_groups function was not respecting the group orders which were saved before as group settings.

### Feature Preview

---

#### PR Checklist


zoli:
Done.

frontend/rust-lib/flowy-database2/src/services/group/configuration.rs
```diff
@@ -474,25 +474,32 @@ fn merge_groups(
...
   // Find out the new groups
-  merge_result
-    .all_groups
-    .extend(new_group_map.values().cloned());
-  merge_result.new_groups.extend(new_group_map.into_values());
+  let new_groups = new_group_map.into_values();
+  for (_, group) in new_groups.into_iter().enumerate() {
+    merge_result.all_groups.push(group.clone());
+    merge_result.new_groups.push(group);
+  }
```
richardshiue, line 498: Oops. I think this part can be reverted? Otherwise, this looks good to me

zoli: Done.

PR #4171, opened 2023-12-19
Feat/support document title

qinluhe:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_tauri/src/appflowy_app/components/editor/command/index.ts
+++ b/frontend/appflowy_tauri/src/appflowy_app/components/editor/command/index.ts
@@ -123,7 +123,7 @@ export const CustomEditor = {
     const index = editor.children.findIndex((child) => (child as Element).blockId === node.blockId);
 
     const level = node.level ?? 1;
-    const subordinateNodes: Element[] = [];
+    const subordinateNodes: (Element & { level: number })[] = [];
 
     if (index === editor.children.length - 1) return subordinateNodes;
 
@@ -308,6 +308,24 @@ export const CustomEditor = {
     return CustomEditor.getBlockType(editor) === EditorNodeType.GridBlock;
   },
 
+  isRootBlock: (editor: ReactEditor) => {
+    const matchEntries = Editor.nodes(editor, {
+      match: (n) => !Editor.isEditor(n) && Element.isElement(n) && n.blockId !== undefined && n.type !== undefined,
+    });
+
+    for (const matchEntry of matchEntries) {
+      const node = matchEntry[0] as Element & {
+        type: EditorNodeType;
+      };
+
+      if (node.type === EditorNodeType.Page) {
+        return true;
+      }
+    }
+
+    return false;
+  },
+
   isCodeBlock: (editor: ReactEditor) => {
     return CustomEditor.getBlockType(editor) === EditorNodeType.CodeBlock;
   },
@@ -330,35 +348,5 @@ export const CustomEditor = {
     );
     ReactEditor.focus(editor);
     Transforms.move(editor);
-  },
-
-  insertLineAtStart: (editor: ReactEditor & YjsEditor, node: Element) => {
-    const blockId = generateId();
-    const parentId = editor.sharedRoot.getAttribute('blockId');
-
-    ReactEditor.focus(editor);
-    editor.insertNode(
-      {
-        ...node,
-        blockId,
-        parentId,
-        textId: generateId(),
-        level: 1,
-      },
-      {
-        at: [0],
-      }
-    );
-
-    editor.select({
-      anchor: {
-        path: [0, 0],
-        offset: 0,
-      },
-      focus: {
-        path: [0, 0],
-        offset: 0,
-      },
-    });
   },
 };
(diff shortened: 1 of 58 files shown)


frontend/appflowy_tauri/src/appflowy_app/components/editor/command/index.ts
```diff
@@ -308,6 +308,24 @@ export const CustomEditor = {
     return CustomEditor.getBlockType(editor) === EditorNodeType.GridBlock;
   },
 
+  isRootBlock: (editor: ReactEditor) => {
```
appflowy, line 311: How about this:
```ts
isRootBlock: (editor: ReactEditor) => {
  const [match] = Editor.nodes(editor, {
    match: n => 
      Element.isElement(n) && 
      n.blockId !== undefined && 
      n.type === EditorNodeType.Page,
  });

  return Boolean(match);
},

```
1.the loop exists asap
2.using Boolean

frontend/appflowy_tauri/src/appflowy_app/components/editor/components/blocks/page/Page.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { forwardRef, memo } from 'react';
+import { EditorElementProps, PageNode } from '$app/application/document/document.types';
+
+export const Page = memo(
```
appflowy, line 4: From OpenAI suggestions:

Destructuring className: Directly destructuring className from the props for clarity. This makes it explicit that you're expecting a className prop and plan to use it.

Combined Class Name: Created a combinedClassName variable to handle the concatenation of class names. This improves readability and makes it easier to manage the class names.

Removed Unused Variable: The _node variable was unused in the original component. If it's not being used, it's better to omit it to avoid confusion.

```ts
export const Page = memo(
  forwardRef<HTMLDivElement, EditorElementProps<PageNode>>(({ node, children, className, ...attributes }, ref) => {
    const combinedClassName = `mb-2 text-4xl font-bold ${className ?? ''}`;

    return (
      <div ref={ref} {...attributes} className={combinedClassName}>
        {children}
      </div>
    );
  })
);
```

frontend/appflowy_tauri/src/appflowy_app/components/editor/components/editor/CollaborativeEditor.tsx
```diff
@@ -5,12 +5,42 @@ import { EditorProps } from '$app/application/document/document.types';
...
-  }, [props.id]);
+    return new Provider(id, showTitle);
+  }, [id, showTitle]);
+
+  const root = useMemo(() => {
+    return showTitle ? (sharedType?.toDelta()[0].insert as YXmlText | null) : null;
+  }, [sharedType, showTitle]);
+
+  useEffect(() => {
+    if (!root) return;
```
appflowy, line 20:
```ts
if (!root || title === root.toString()) return;
```

frontend/appflowy_tauri/src/appflowy_app/components/editor/components/editor/CollaborativeEditor.tsx
```diff
@@ -5,12 +5,42 @@ import { EditorProps } from '$app/application/document/document.types';
...
+  }, [title, root]);
+
+  useEffect(() => {
+    if (!root) return;
+    const onChange = () => {
+      onTitleChange?.(root.toString());
+    };
+
+    root.observe(onChange);
+    return () => {
```
appflowy, line 40: Not sure which one is better, I prefer this:
```ts
return () => root.unobserve(onChange);
```

frontend/appflowy_tauri/src/appflowy_app/components/editor/plugins/withMergeNodes.ts
```diff
@@ -1,42 +1,47 @@
...
-          Transforms.setNodes(editor, { level: previousLevel + 1, parentId: previousNode.blockId }, { at: childPath });
-        });
-      }
+    if (mergedNode.blockId === nextNode.parentId) {
+      return;
     }
 
-    mergeNodes(...args);
+    // check if the old node is removed
+    const oldNodeRemoved = !editor.children.some((child) => (child as Element).blockId === nextNode.parentId);
```
appflowy, line 29:
```ts
if (oldNodeRemoved) {
  const newLevel = mergedNode.level + 1;
  editor.children.forEach((child, index) => {
    if ((child as Element).parentId === nextNode.parentId) {
      Transforms.setNodes(
        editor,
        { level: newLevel, parentId: mergedNode.blockId },
        { at: [index] } 
      );
    }
  });
}
```
Can I combine two loop into one? It can improve efficiency

frontend/appflowy_tauri/src/appflowy_app/components/editor/command/index.ts
qinluhe: In fact, this function is used to determine whether the selection contains the root (that is, title) node. For example, if the selection contains root, the toolbar will not be displayed.
Let me change the name of this function

PR #4184, opened 2023-12-21
feat: implement folder indexer

appflowy:
#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-sqlite/src/search.rs
+++ b/frontend/rust-lib/flowy-sqlite/src/search.rs
@@ -1,0 +1,171 @@
+use diesel::{sql_query, sql_types::Text, QueryResult, RunQueryDsl, SqliteConnection};
+
+const SEARCH_VIEW_TABLE: &str = "search_view";
+const CREATE_SEARCH_VIEW_TABLE: &str =
+  "CREATE VIRTUAL TABLE if not exists search_view USING fts5(id, name)";
+
+/// Runs database migrations for local search using sqlite FTS5.
+///
+/// FTS5 tables do not have indexes, which is not supported by Diesel.
+pub fn run_migrations(conn: &mut SqliteConnection) -> QueryResult<usize> {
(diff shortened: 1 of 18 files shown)


frontend/rust-lib/flowy-sqlite/src/search.rs
```diff
@@ -0,0 +1,298 @@
...
+    let database = Database::new(path, DB_NAME, pool_config).unwrap();
+    let mut conn = database.get_connection().unwrap();
+    (*conn).run_pending_migrations(MIGRATIONS).unwrap();
+    run_migrations(&mut conn).unwrap();
+
+    (tempdir, database)
+  }
+
+  #[test]
+  fn test_migration() {
```
appflowy, line 167: the database is always new. Does it necessary to test it?

frontend/rust-lib/flowy-sqlite/src/search.rs
```diff
@@ -0,0 +1,298 @@
...
+      index_type: IndexType::Document.to_string(),
+      view_id: view_id.to_owned(),
+      id: page_id.to_owned(),
+      data: text.to_owned(),
+    }
+  }
+}
+
+/// Add search data for searching.
+pub fn add(conn: &mut SqliteConnection, data: &SearchData) -> QueryResult<usize> {
```
appflowy, line 93: Can we use the tokio::task::spawn_blocking if there are too many add called

frontend/rust-lib/flowy-sqlite/src/search.rs
Jiraffe7: This test was used as part of the dev process. It's not needed per se, so I have removed it.

frontend/rust-lib/flowy-sqlite/src/search.rs
Jiraffe7: Do you mean wrap the calls to add in `spawn_blocking`? Currently, the runs in a task spawned by `af_spawn` in `folder_indexer.rs`, so the calls can be wrapped there.

PR #4296, opened 2024-01-03
Search frontend integration

Jiraffe7:
### Feature Preview

---

#### PR Checklist


frontend/rust-lib/flowy-folder/src/search/folder_indexer.rs
```diff
@@ -0,0 +1,123 @@
...
+    S: FolderIndexStorage + 'static,
+    D: DocumentIndexContentGetter + 'static,
+  {
+    Self {
+      storage: Arc::new(storage),
+      document_index_content_getter: Arc::new(doc_getter),
+    }
+  }
+
+  pub fn search(&self, s: &str, limit: Option<i64>) -> Result<Vec<SearchData>, FlowyError> {
```
appflowy, line 48: How do you handle if keep typing that will generate too much request?

frontend/rust-lib/flowy-core/src/deps_resolve/folder_deps.rs
```diff
@@ -478,3 +483,108 @@ pub fn layout_type_from_view_layout(layout: ViewLayoutPB) -> DatabaseLayoutPB {
     ViewLayoutPB::Document => DatabaseLayoutPB::Grid,
   }
 }
+
+struct FolderIndexStorageImpl(Weak<UserManager>);
+
+impl FolderIndexStorage for FolderIndexStorageImpl {
+  fn search(&self, s: &str, limit: Option<i64>) -> Result<Vec<SearchData>, FlowyError> {
+    let manager = self
```
appflowy, line 491: lot of duplicate of this code

frontend/rust-lib/flowy-folder/src/search/folder_indexer.rs
Xazin: The client will be implementing debouncing, ensuring that every new character won't trigger a search.

Were you thinking of implementing a rate limit for search?

Jiraffe7: Good point. The following are some ways to work around this:
1. Perform the search only after the user completes typing and presses enter to search. This is the easiest solution with the limitation that there is no search assistance while users are typing.
2. Debounce by waiting for a certain amount of time after the user stops typing before executing the search. This can be done in two places:
	1. Frontend. The frontend contains the state of the search input as the user is typing, and determines when to send the search request to the backend, so it makes sense to perform debouncing here. This is also a common approach for handling inputs.
	2. Backend. Since the backend is not stateful and does not have the state of the search input as the user is typing, it does not make sense to perform debouncing in the backend.
3. Limit dynamic search scope to improve performance where dynamic refers to the period when the user is entering the search term, and only performing the full search against the document content after the user presses enter. Limiting dynamic search to page titles only (as a means of search suggestion), and limiting the number of suggestions to roughly 10.

I propose that a combination of debouncing the user input in the frontend (2i) and limiting dynamic search scope to page titles (3) will be effective in dealing with the performance concerns.

Let me know what you think.

frontend/rust-lib/flowy-core/src/deps_resolve/folder_deps.rs
Jiraffe7: Yes, I'm not clear on the structure of the crate/dependency flow and how they relate to the state of the application.

When viewed from an object-oriented perspective, a folder contains documents and databases. However, there seems to be no relation between a folder and the documents/databases it contains; there does not appear to be a way for a folder to obtain the contents of its documents/databases.

My assumption is that the different crates were created with boundaries in mind, i.e. folder should not use document/database and vice versa, which limits the implementation of shared functionality in the core package that is the only place that uses all crates.

appflowy:
Refactoring:

1. Replace the current handcrafted migration process with Diesel's migration system for enhanced efficiency and maintainability.
2. Eliminate the document indexing functionality from the folder indexer.
3. Develop a new crate, `flowy-search`, dedicated to managing indexing and search logic, and subsequently remove the indexing logic from the `flowy-user` crate.

appflowy:
Close this PR and continue on: https://github.com/AppFlowy-IO/AppFlowy/pull/4458

PR #4299, opened 2024-01-04
feat: initial file upload api

speed2exe:
- File Upload API


frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -391,3 +393,80 @@ pub(crate) async fn convert_data_to_json_handler(
...
+
+  let local_file_path = params.local_file_path.to_owned();
+  Ok(AFPluginData(UploadedFile {
+    url,
+    local_file_path,
+  }))
+}
+
+pub(crate) async fn download_file_handler(
+  params: AFPluginData<UploadedFile>,
```
LucasXu0, line 426: Download?

speed2exe: This is a scenario where user signs into another device, loads a document and have to download file to his new device file system.

LucasXu0: No. I meant the type of the params. "download_file_handler" API with a "Upload" params.

speed2exe: the params type is actually different from `upload_file_handler`. 
```
#[derive(Default, ProtoBuf)]
pub struct UploadedFilePB {
  #[pb(index = 1)]
  pub url: String,

  #[pb(index = 2)]
  pub local_file_path: String,
}
```
I read it as "downloading an uploaded file". I'm ok to change the names if it's confusing

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+
+  let url = file_service
+    .get_object_url(ObjectIdentity {
+      workspace_id: params.workspace_id.to_owned(),
+      file_id: hash.to_string(),
+    })
+    .await?;
+
+  // let the upload happen in the background
+  let url_clone = url.clone();
```
appflowy, line 438: I think the `cloned_url` would be the more conventional choice in Rust  In most cases, But url_clone is also ok

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+  tokio::spawn(async move {
+    if let Err(e) = file_service
+      .put_object(
+        url_clone,
+        ObjectValue {
```
appflowy, line 443: The suggestion here is to encapsulate the logic for creating an `ObjectValue` from a file into a dedicated method (`from_file`) within the `ObjectValue` struct or associated implementation in Rust. This approach offers several benefits in terms of code organization and dependency management. Let me clarify this:

**Suggested Refactoring**:
- Create a `from_file` method within the `ObjectValue` implementation.
- This method will handle the process of opening a file, reading its contents, and determining the necessary metadata (like hash and mime type).

**New Function in `ObjectValue`**:

```rust
impl ObjectValue {
    async fn from_file(local_file_path: &str) -> Result<Self, io::Error> {
        let mut file = tokio::fs::File::open(local_file_path).await?;
        let mut content = Vec::new();
        let n = file.read_to_end(&mut content).await?;
        info!("read {} bytes from file: {}", n, local_file_path);

        let hash = fxhash::hash(&content);
        let mime = mime_guess::from_path(local_file_path).first_or_octet_stream();

        // Construct and return the ObjectValue based on the file content and metadata
        Ok(ObjectValue {
            // ... fields populated with content, hash, mime, etc.
        })
    }
}
```

**Benefits**:

1. **Code Reusability**: By centralizing the file reading and ObjectValue construction logic in `from_file`, you avoid duplicating this code wherever you need to create an `ObjectValue` from a file. This makes the codebase DRY (Don't Repeat Yourself), easier to maintain, and less error-prone.

2. **Reduced Dependency Spread**: The dependencies `fxhash`, `mime`, and `mime_guess` are only needed within the `ObjectValue` implementation. Other parts of the codebase that create `ObjectValue` instances don't need to directly depend on these libraries, reducing the spread and management of dependencies.

3. **Simplified Interface**: External code that needs to create an `ObjectValue` from a file can do so with a single, straightforward method call, without needing to understand the underlying details of file handling and metadata processing.

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+
+  let local_file_path = params.local_file_path.to_owned();
+  Ok(AFPluginData(UploadedFilePB {
+    url,
+    local_file_path,
+  }))
+}
+
+pub(crate) async fn download_file_handler(
+  params: AFPluginData<UploadedFilePB>,
```
appflowy, line 462: as same as the `UploadFileParamsPB`. You can validate the input by using the validator

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+  manager: AFPluginState<Weak<DocumentManager>>,
+) -> FlowyResult<()> {
+  let path = params.local_file_path.clone();
```
appflowy, line 465: i think you may want this:
```rust
let UploadedFilePB { url, local_file_path } = params;
```

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+
+  // if file already exist in user local disk, return
+  if tokio::fs::metadata(&path).await.is_ok() {
+    return Ok(());
```
appflowy, line 469: maybe we can warn! here

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+  });
+
+  let local_file_path = params.local_file_path.to_owned();
+  Ok(AFPluginData(UploadedFilePB {
+    url,
+    local_file_path,
+  }))
+}
+
+pub(crate) async fn download_file_handler(
```
appflowy, line 461: using #[instrument] plus the event! can make the log more easier to read.

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+    .upgrade()
+    .ok_or(FlowyError::internal().with_context("The file storage service is already dropped"))?;
+  file_service.delete_object(params.url.clone()).await?;
+
+  tokio::fs::remove_file(&params.local_file_path).await?;
+  Ok(())
+}
+
+// Wrapper around FxHasher to implement AsyncWrite
+struct HasherWriter<F: Hasher> {
```
appflowy, line 509: You can move the HasherWriter to lib-infra.

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+pub(crate) async fn delete_uploaded_file_handler(
+  params: AFPluginData<UploadedFilePB>,
+  manager: AFPluginState<Weak<DocumentManager>>,
+) -> FlowyResult<()> {
+  let manager = upgrade_document(manager)?;
+  let file_service = manager
+    .get_file_storage_service()
+    .upgrade()
+    .ok_or(FlowyError::internal().with_context("The file storage service is already dropped"))?;
+  file_service.delete_object(params.url.clone()).await?;
```
appflowy, line 502: the url is not need to clone

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+
+  tokio::fs::remove_file(&params.local_file_path).await?;
```
appflowy, line 504: I think the frontend might can't handle the error when remove_file fail. So it doesn't need to propagate the remove_file error. We can just log it. But we can have a junk list somewhere. When user try to clean up the disk space. we can try to delete these files again.

frontend/rust-lib/flowy-server/src/af_cloud/impls/file_storage.rs
```diff
@@ -16,47 +12,48 @@ impl<T> AFCloudFileStorageServiceImpl<T> {
...
-        ObjectValue::Bytes { bytes, mime } => {
-          Ok(client.put_blob(&object.workspace_id, bytes, mime).await?)
-        },
-      }
+  fn put_object(&self, url: String, file: ObjectValue) -> FutureResult<(), FlowyError> {
+    let try_get_client = self.0.try_get_client();
+    let file = file.clone();
+    FutureResult::new(async move {
+      let client = try_get_client?;
+      let url = url;
```
appflowy, line 33: can remove

frontend/rust-lib/flowy-server/src/af_cloud/server.rs
```diff
@@ -45,7 +45,12 @@ pub struct AppFlowyCloudServer {
 
 impl AppFlowyCloudServer {
   pub fn new(config: AFCloudConfiguration, enable_sync: bool, device_id: String) -> Self {
-    let api_client = AFCloudClient::new(&config.base_url, &config.ws_base_url, &config.gotrue_url);
+    let api_client = AFCloudClient::new(
+      &config.base_url,
+      &config.ws_base_url,
+      &config.gotrue_url,
+      ClientConfiguration::default(),
```
appflowy, line 52: You can merge the main branch first. The main branch of AppFlowy already resolve it.

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+
+  // create file if not exist
+  let mut file = tokio::fs::OpenOptions::new()
+    .create(true)
+    .write(true)
+    .open(&path)
+    .await?;
+
+  let n = file.write(&object_value.raw).await?;
+  info!("downloaded {} bytes to file: {}", n, path);
```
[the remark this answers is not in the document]
speed2exe: from the source code, `info!` seems just like a macro to call `event!`. isn't it the same thing?

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
 
   data_result_ok(ConvertDataToJsonResponsePB { json: result })
 }
+
+// Handler for uploading a file
+// `workspace_id` and `file_name` determines file identity
+pub(crate) async fn upload_file_handler(
+  params: AFPluginData<UploadFileParamsPB>,
```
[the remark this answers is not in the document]
speed2exe: added validation logic

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: changed

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: moved logic

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: added

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: added

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: the compiler is good when i do this instead, it should be the same?
```
  let AFPluginData(UploadedFilePB {
    url,
    local_file_path,
  }) = params;
```


else i would get this diagnostic

```
Diagnostics:
1. mismatched types
   expected struct `AFPluginData<entities::UploadedFilePB>`
      found struct `entities::UploadedFilePB` [E0308]
2. try wrapping the pattern in `lib_dispatch::prelude::AFPluginData`: `lib_dispatch::prelude::AFPluginData(`, `)` [E0308]
```

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: added

frontend/rust-lib/flowy-document/src/event_handler.rs
```diff
@@ -401,3 +406,125 @@ pub(crate) async fn convert_data_to_json_handler(
...
+  // if file already exist in user local disk, return
+  if tokio::fs::metadata(&path).await.is_ok() {
+    return Ok(());
+  }
+
+  let manager = upgrade_document(manager)?;
+  let file_service = manager
+    .get_file_storage_service()
+    .upgrade()
+    .ok_or(FlowyError::internal().with_context("The file storage service is already dropped"))?;
```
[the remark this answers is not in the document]
speed2exe: used `ok_or_else` instead

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: removed the clone

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: I added a TODO for this. I suppose the frontend needs to retry the deletion operation when it can. not particular sure when it should trigger but mechanisms can be further discussed.

frontend/rust-lib/flowy-document/src/event_handler.rs
speed2exe: removed it instead. i realised it's not needed and used. since it's simpler and probably more effective to put it in memory, as it's required to compute the hash and also send the raw data to the server.

frontend/rust-lib/flowy-server/src/af_cloud/impls/file_storage.rs
speed2exe: done

frontend/rust-lib/flowy-server/src/af_cloud/server.rs
speed2exe: merged

frontend/rust-lib/flowy-document/src/event_handler.rs
appflowy: Yep. We can do more discussion later

frontend/rust-lib/flowy-document/src/event_handler.rs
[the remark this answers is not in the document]
appflowy: oh. Not really the same but in most case it's.

PR #4317, opened 2024-01-05
fix: upload custom theme issue

hyj1204:
### Feature Preview

- Fix the path issue in encoding FlowyDynamicPlugin
- Show a detailed message when an upload error occurs: The error messages are under the `flowy_infra` package which cannot access the `LocaleKeys`, so I didn't add the localization for them.
- Show the ‘Delete’ button only if the custom theme is not the current theme. 
- Some improvement on upload theme UI

https://github.com/AppFlowy-IO/AppFlowy/assets/14248245/e17a55dc-0ce8-4fb5-a085-be33f75f44d3

---

#### PR Checklist


frontend/appflowy_flutter/lib/workspace/presentation/settings/widgets/theme_upload/theme_upload_view.dart
```diff
@@ -41,9 +41,10 @@ class _ThemeUploadWidgetState extends State<ThemeUploadWidget> {
             key: Key('upload_theme_loading_widget'),
           );
         },
-        compilationFailure: () {
-          child = const ThemeUploadFailureWidget(
-            key: Key('upload_theme_failure_widget'),
+        compilationFailure: (errorMessage) {
```
richardshiue, line 44: It'd be better to log the message and use localized version

Issue #4355, opened 2024-01-10
[FR] Variable depth of '/outline'


PeterHero:
### Description

A number picker next to `/outline` component would set depth to which the outline should be generated. That would ease to create clear navigation of the page consisting only of h1 or h1 and h2.

### Impact

Users who want to create clear and minimal overview of the page.

### Additional Context

Could be implemented also for future feature described in #4282

PR #4358, opened 2024-01-10
feat: support dynamic depth control in outline block component

Jayaprakash-dev:
This PR is related to the feature request:

> [FR] Variable depth of '/outline' #4355

**Description**:
This PR addresses the implementation of a feature that enhances the `OutlineBlockComponent` functionality by introducing dynamic depth control. The depth control allows users to customise the visibility of headings based on their selected depth.

**Changes Made:**
Added dynamic depth control to the outline block.
When a depth H1 is selected, the app now displays headings up to level 1.
For depth H2, it shows headings up to level 2, extending upto subsequent H3 headings.

https://github.com/AppFlowy-IO/AppFlowy/assets/68953739/879fa002-7e5a-48b7-9025-a866a48ef40a

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/actions/option_action.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/actions/option_action.dart
@@ -375,7 +375,7 @@ class DepthOptionAction extends PopoverActionCell {
   }
 
   OptionDepthType depth(Node node) {
-    final level = node.attributes['depth'];
+    final level = node.attributes[OutlineBlockKeys.depth];
     return OptionDepthType.fromLevel(level);
   }
 
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/actions/option_action.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/actions/option_action.dart
@@ -19,6 +19,9 @@ enum OptionAction {
   moveUp,
   moveDown,
 
+  /// controls outline block tree depth
+  depth,
+
   /// callout background color
   color,
   divider,
@@ -42,6 +45,8 @@ enum OptionAction {
         return const FlowySvgData('editor/divider');
       case OptionAction.align:
         return FlowySvgs.align_center_s;
+      case OptionAction.depth:
+        throw UnimplementedError();
     }
   }
 
@@ -63,6 +68,8 @@ enum OptionAction {
         return LocaleKeys.document_plugins_optionAction_align.tr();
       case OptionAction.divider:
         throw UnsupportedError('Divider does not have description');
+      case OptionAction.depth:
+        return LocaleKeys.document_plugins_optionAction_depth.tr();
     }
   }
 }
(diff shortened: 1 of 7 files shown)


qinluhe:
I think it would be more flexible to make the filter ( like database hide field ), but this is also good.❤️

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/actions/option_action.dart
```diff
@@ -286,6 +334,80 @@ class ColorOptionAction extends PopoverActionCell {
...
+          parentController.close();
+        });
+
+        return IntrinsicHeight(
+          child: IntrinsicWidth(
+            child: Column(
+              children: children,
+            ),
+          ),
+        );
```
LucasXu0, line 360:
```suggestion
        return SizedBox(
          width: 60,
          child: Column(
            mainAxisSize: MainAxisSize.min,
            children: children,
          ),
        );
```

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/actions/option_action.dart
```diff
@@ -286,6 +334,80 @@ class ColorOptionAction extends PopoverActionCell {
...
+
+    final node = editorState.getNodeAtPath(selection.start.path);
+    if (node == null) return;
+
+    if (depth == this.depth(node)) return;
+
+    final transaction = editorState.transaction;
+    transaction.updateNode(
+      node,
+      {'depth': depth.level},
```
LucasXu0, line 395: define the 'depth' as a const value in `OutlineBlockKeys`

```dart
class OutlineBlockKeys {
  const OutlineBlockKeys._();

  static const String type = 'outline';
  static const String backgroundColor = blockComponentBackgroundColor;

  static const String depth = 'depth';
}
```

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/outline/outline_block_component.dart
```diff
@@ -167,6 +167,17 @@ class _OutlineBlockWidgetState extends State<OutlineBlockWidget>
 
   Iterable<Node> getHeadingNodes() {
     final children = editorState.document.root.children;
+
+    if (node.attributes.containsKey('depth')) {
+      final int level = node.attributes['depth'];
+      return children.where(
+        (element) =>
+            element.type == HeadingBlockKeys.type &&
+            element.attributes['level'] <= level &&
```
LucasXu0, line 176: use `HeadingBlockKeys.level` instead

LucasXu0:
I found an issue: with popover 

1. click on the color item.
2. click on the depth item.
3. the previously displayed color list will not appear.

Screenshot 2024-01-11 at 15 34 28

Jayaprakash-dev:
> I found an issue: with popover
> 

@**LucasXu0**, I believe the issue originates from the popover menu itself rather than the depth option action menu. Additionally, the colour option action menu is also blocking the visibility of other popovers. 

The issue here is that the popover menu does not close the child menus before opening a new one; instead, it overlays on top of another. Therefore, I suggest creating a new bug issue to track this specific problem.

https://github.com/AppFlowy-IO/AppFlowy/assets/68953739/6a03c900-447d-4e2a-9f53-9c4b3bd96c2e

richardshiue:
> > I found an issue: with popover
> 
> @**LucasXu0**, I believe the issue originates from the popover menu itself rather than the depth option action menu. Additionally, the colour option action menu is also blocking the visibility of other popovers.
> 
> The issue here is that the popover menu does not close the child menus before opening a new one; instead, it overlays on top of another. Therefore, I suggest creating a new bug issue to track this specific problem.
> Screen.Recording.2024-01-11.at.1.28.17.PM.mov

You can use the PopoverMutex deal with this issue for now

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/outline/outline_block_component.dart
```diff
@@ -167,6 +168,17 @@ class _OutlineBlockWidgetState extends State<OutlineBlockWidget>
 
   Iterable<Node> getHeadingNodes() {
     final children = editorState.document.root.children;
+
+    if (node.attributes.containsKey('depth')) {
```
LucasXu0, line 172: @Jayaprakash-dev Please use the defined const key instead of 'depth'

Jayaprakash-dev:
Hey @LucasXu0, I've made the necessary changes.

PR #4488, opened 2024-01-24
chore: code cleanup according to unintroduced lints

Xazin:
This PR includes a small bunch of commits, that each clean up the code by targeting a specific lint.

The commits relate to:
- avoid_redundant_argument_values
- avoid_unused_constructor_parameters
- sort_constructors_first _(Partially / 60-70%)_
- unnecessary_await_in_return
- unnecessary_parenthesis

I think merging this would cause some conflicts in existing PRs, albeit not really hard to resolve, but we can choose to pick out any commit from here if we don't want to merge the ones that would cause the most damage at the moment. I don't think there will ever be a "prime time" to do this clean up though, as we keep creating more and more PRs.

It is on purpose that I haven't introduced the lints themselves yet.


appflowy:
I believe we can proceed with the merge if everything is ready. Should there be any conflicts, they appear to be manageable and not too difficult to resolve.

Xazin:
I will add the lints 👍

PR #4469, opened 2024-01-23
Cloud workspace api

speed2exe:
AppFlowy Cloud Support:
- Add Workspace
- Remove Workplace

PR #4560, opened 2024-01-31
feat: add blocking option for file upload

speed2exe:


speed2exe:
already included in:

https://github.com/AppFlowy-IO/AppFlowy/pull/4469

PR #4576, opened 2024-02-01
chore: inkwell mobile quick action sheets

richardshiue:
- change the height of each item to be 52px, less padding of 4px vertically to have 44px height for the inkwell widget
- change the thickness of divider to 0.5px
- used in:
  - view actions in homepage after sliding
  - view actions in the more appbar button
  - database card detail actions
  - database view actions.

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/5536ce68-84a0-4761-8c2f-21f61e7996f5

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/mobile/presentation/database/card/card_detail/mobile_card_detail_screen.dart
+++ b/frontend/appflowy_flutter/lib/mobile/presentation/database/card/card_detail/mobile_card_detail_screen.dart
@@ -136,39 +136,14 @@ class _MobileRowDetailPageState extends State<MobileRowDetailPage> {
         mainAxisSize: MainAxisSize.min,
         children: [
           MobileQuickActionButton(
-            onTap: () {
-              final rowId = _bloc.state.currentRowId;
-              if (rowId == null) {
-                return;
-              }
-              RowBackendService.duplicateRow(viewId, rowId);
-              context
-                ..pop()
-                ..pop();
-              Fluttertoast.showToast(
-                msg: LocaleKeys.board_cardDuplicated.tr(),
-                gravity: ToastGravity.BOTTOM,
-              );
-            },
+            onTap: () =>
+                _performAction(viewId, _bloc.state.currentRowId, false),
             icon: FlowySvgs.copy_s,
             text: LocaleKeys.button_duplicate.tr(),
           ),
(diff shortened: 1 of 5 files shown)


frontend/appflowy_flutter/lib/mobile/presentation/database/card/card_detail/mobile_card_detail_screen.dart
```diff
@@ -135,49 +135,44 @@ class _MobileRowDetailPageState extends State<MobileRowDetailPage> {
...
+              }
+              RowBackendService.deleteRow(viewId, rowId);
+              context
+                ..pop()
+                ..pop();
+              Fluttertoast.showToast(
+                msg: LocaleKeys.board_cardDeleted.tr(),
+                gravity: ToastGravity.BOTTOM,
+              );
+            },
```
Xazin, line 171: How about unifying the two onTaps? They do mostly the same except a different action.
```
void _performAction(String viewId, String? rowId, bool deleteRow) {
  if (rowId == null) {
    return;
  }

  // If not deletion, then it's duplication
  deleteRow
    ? RowBackendService.deleteRow(viewId, rowId)
    : RowBackendService.duplicateRow(viewId, rowId);

  context
    ..pop()
    ..pop();
  Fluttertoast.showToast(
    msg: LocaleKeys.board_cardDeleted.tr(),
    gravity: ToastGravity.BOTTOM,
  );
}
```

```suggestion
            onTap: () => 
              _performAction(viewId, _bloc.state.currentRowId, true),
```

LucasXu0:
Should we update the style of all buttons in the action sheet? Otherwise, other buttons in the action sheet are using `FlowyOptionTile.text`, and the `MobileQuickActionButton` is only used in the database action sheet.

richardshiue:
> Should we update the style of all buttons in the action sheet? Otherwise, other buttons in the action sheet are using `FlowyOptionTile.text`, and the `MobileQuickActionButton` is only used in the database action sheet.

We should discuss this. For action sheets that lets you choose an option and then clicking Done or dismissing the sheet another way, this style is looks very weird. An example is the theme mode

image

I think we should only use MobileQuickActionButton when the action sheet is immediately dismissed.

PR #4601, opened 2024-02-04
feat: sort by last modified time or created time

richardshiue:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/src/services/database_view/view_editor.rs
+++ b/frontend/rust-lib/flowy-database2/src/services/database_view/view_editor.rs
@@ -160,6 +160,9 @@ impl DatabaseViewEditor {
     send_notification(&self.view_id, DatabaseNotification::DidUpdateViewRows)
       .payload(changes)
       .send();
+    self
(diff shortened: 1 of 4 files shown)


frontend/rust-lib/flowy-database2/src/services/database_view/view_editor.rs
```diff
@@ -1052,4 +1034,29 @@ impl DatabaseViewEditor {
...
+          .read()
+          .await
+          .did_receive_row_changed(row_id)
+          .await;
+      }
+      if let Some(calculations_controller) = weak_calculations_controller.upgrade() {
+        calculations_controller
+          .did_receive_cell_changed(field_id)
+          .await;
+      }
```
Xazin, line 1059: It's redundant to recalculate on row create, as the new cells would be empty.

richardshiue: Not necessarily, as rows can be created with pre-filled cells

PR #4608, opened 2024-02-05
feat: bottom sheets with transition

richardshiue:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_mobile_bottom_sheet.dart
+++ b/frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_mobile_bottom_sheet.dart
@@ -79,7 +79,7 @@ Future<T?> showMobileBottomSheet<T>(
 
       if (showHeader) {
         children.add(
-          _Header(
+          Header(
             showCloseButton: showCloseButton,
             showBackButton: showBackButton,
             showDoneButton: showDoneButton,
@@ -152,8 +152,9 @@ Future<T?> showMobileBottomSheet<T>(
   );
 }
 
-class _Header extends StatelessWidget {
-  const _Header({
+class Header extends StatelessWidget {
+  const Header({
+    super.key,
     required this.showBackButton,
     required this.showCloseButton,
     required this.title,
--- a/frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_mobile_bottom_sheet.dart
+++ b/frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_mobile_bottom_sheet.dart
@@ -79,7 +79,7 @@ Future<T?> showMobileBottomSheet<T>(
 
       if (showHeader) {
         children.add(
-          Header(
+          BottomSheetHeader(
             showCloseButton: showCloseButton,
             showBackButton: showBackButton,
             showDoneButton: showDoneButton,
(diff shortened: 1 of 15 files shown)


richardshiue:
https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/761e0338-71e6-4aff-93e0-5493eb21766d

frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_mobile_bottom_sheet.dart
```diff
@@ -166,8 +163,9 @@ Future<T?> showMobileBottomSheet<T>(
   );
 }
 
-class _Header extends StatelessWidget {
-  const _Header({
+class Header extends StatelessWidget {
```
LucasXu0, line 166: If this widget is public, rename it. The name `Header` is too common and likely to cause conflicts.

frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_transition_bottom_sheet.dart
```diff
@@ -0,0 +1,348 @@
...
+
+import 'package:appflowy/plugins/base/drag_handler.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:sheet/route.dart';
+import 'package:sheet/sheet.dart';
+
+import 'show_mobile_bottom_sheet.dart';
+
+Future<T?> showTransitionMobileBottomSheet<T>(
```
LucasXu0, line 11: Would it be more reusable to add a bool value named `withTransition` to `showMobileBottomSheet` instead of adding a new function?

frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_transition_bottom_sheet.dart
```diff
@@ -0,0 +1,348 @@
...
+  BuildContext context, {
+  required WidgetBuilder builder,
+  bool useRootNavigator = false,
+  double? elevation,
+  BoxConstraints? constraints,
```
LucasXu0, line 16: Some properties are not being used in this function.

richardshiue:
Sorry @LucasXu0, I pushed with `git push +:` and this PR got forced pushed as well. What changes did you make?

frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_transition_bottom_sheet.dart
richardshiue: That's a really good question. What do you think? My thoughts initially is that if we use the same function signature, some parameters will not be used and we have to rely on comments to communicate which parameters are available in one case and the other.

LucasXu0:
> Sorry @LucasXu0, I pushed with `git push +:` and this PR got forced pushed as well. What changes did you make?

Never mind. I didn't make any changes.

frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_mobile_bottom_sheet.dart
richardshiue: Fixed.

frontend/appflowy_flutter/lib/mobile/presentation/bottom_sheet/show_transition_bottom_sheet.dart
richardshiue: Removed!

PR #4641, opened 2024-02-10
feat: HHmm to HH:mm conversion in time field

### LINKED ISSUE #4145: [FR] Time entry in date property | Automatic colon for 24-hour input
### Description

When entering the time in the date property, the inclusion of inputting the colon ( : ) is necessary for 24 hour formats. I would expect a behavior in which I can simply input '1600' and press enter, with the entry being auto-completed as '16:00'. Haven't used Notion for years so I'm unsure if I picked the habit up from there, but that was my muscle memory when inputting the time. Would assume this intuition/habit is shared by most.

### Impact

Anyone that is as lazy as me. Having to input the colon is a significant detriment to my workflow.

### Additional Context

_No response_

AnsahMohammad:
### Feature Preview

[Screencast from 10-02-24 02:17:46 PM IST.webm](https://github.com/AppFlowy-IO/AppFlowy/assets/85027826/0aa8d904-6aef-41dc-8d2e-045a6ce1be44)

This PR Closes #4145 
---

I believe there might be a more efficient approach for this. I tried to utilize the 'HHmm' format directly but encountered some errors. I'm open to suggestions.

#### PR Checklist


AnsahMohammad:
Replaced parse() with parseStrict() to address the issue #4644

Xazin:
Can you also add a Formatter to the TextField? So that it's visible to the user what is happening

AnsahMohammad:
> Can you also add a Formatter to the TextField? So that it's visible to the user what is happening, I'm thinking this could even be the sole solution to the problem.

I didn't understand. Are you suggesting that the text field should display an error when the user provides incorrect input? (using a formatter in text-field)

Xazin:
> > Can you also add a Formatter to the TextField? So that it's visible to the user what is happening, I'm thinking this could even be the sole solution to the problem.
> 
> I didn't understand. Are you suggesting that the text field should display an error when the user provides incorrect input? (using a formatter in text-field)

Formatter are used to enforce a certain format of the input field. Eg. for a phone number, or in this case the `HH:mm` or `HH:mm am/pm`format.

https://api.flutter.dev/flutter/material/TextField/inputFormatters.html
https://api.flutter.dev/flutter/services/TextInputFormatter-class.html

AnsahMohammad:
Current implementation : 
[Screencast from 17-02-24 11:22:45 AM IST.webm](https://github.com/AppFlowy-IO/AppFlowy/assets/85027826/b8682bc8-f25d-4d1f-a011-d706fc2a5a15)

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/mention/mention_date_block.dart
```diff
@@ -264,10 +264,10 @@ class _MentionDateBlockState extends State<MentionDateBlock> {
 
     try {
       if (timeFormat == TimeFormatPB.TwelveHour) {
-        return twelveHourFormat.parse(timeStr);
+        return twelveHourFormat.parseStrict(timeStr);
```
Xazin, line 267: Do we still need to use parseStrict?

Xazin (approved):

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/mention/mention_date_block.dart
AnsahMohammad: Without ParseStrict the Time string can have a value greater than 24. Which will be confusing.

For example :
36:50 gets parsed into 12:50
40:90 into 17:30

If it is preferred this way, we can change it back to `parse` again

PR #4661, opened 2024-02-16
fix: newly-created rows aren't being automatically sorted

### LINKED ISSUE #4600: [Bug] Newly inserted rows aren't being sorted
### Bug Description

Title says all

### How to Reproduce

1. Go to a grid
2. Add a sort
3. Create a new row from
4. See that the row isn't immediately sorted

### Expected Behavior

It should be

### Operating System

macOS

### AppFlowy Version(s)

0.4.6

### Screenshots

_No response_

### Additional Context

_No response_

richardshiue:
resolves #4600

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/tests/database/sort_test/script.rs
+++ b/frontend/rust-lib/flowy-database2/tests/database/sort_test/script.rs
@@ -36,6 +36,7 @@ pub enum SortScript {
     row_id: RowId,
     text: String,
   },
+  AddNewRow,
   AssertSortChanged {
     old_row_orders: Vec<&'static str>,
     new_row_orders: Vec<&'static str>,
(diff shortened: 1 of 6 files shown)


Xazin (approved):
Let's add a test for this specific case so we don't end up breaking it again :eyes: :pray: What do you think @richardshiue ?

richardshiue:
> Let's add a test for this specific case so we don't end up breaking it again 👀 🙏 What do you think @richardshiue ?

I'm not against writing tests, but this is way too specific. I think it's better to refactor this code so that the field controller's `FieldInfo`s get updated whenever something related happens.

Xazin (approved):

LucasXu0 (approved):

frontend/rust-lib/flowy-database2/tests/database/sort_test/script.rs
```diff
@@ -195,6 +218,7 @@ async fn assert_sort_changed(
           old_row_orders.insert(changed.new_index, old);
           assert_eq!(old_row_orders, new_row_orders);
         },
+        DatabaseViewChanged::InsertSortedRowNotification(_changed) => {},
```
Xazin, line 221: Redundant

Xazin (approved):

PR #4676, opened 2024-02-18
Update fr-FR and fr-CA translations

Arcesilas:
Completes #4664

Improve translation for Toggle Lists.


LucasXu0:
I removed the unrelated file changes and kept only the required translations.

PR #4682, opened 2024-02-19
refactor: checkbox field type

richardshiue:
Currently, the checkbox cell data is parsed to a single-String tuple struct. It works, there are no bugs, but can be more intuitive.

image

There are three states. When grouping, calculating, filtering, sorting or what have you, we need to check whether a cell is checked or not.

image

Evidently, `is_uncheck()` may cause bugs in the future. We don't run into this issue now because we negate `is_check()` instead of using `is_uncheck()`, but it would be so much simpler if we just use a bool instead. We also get to remove the dependency on these consts "Yes" and "No" on the frontend-frontend.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/src/entities/type_option_entities/checkbox_entities.rs
+++ b/frontend/rust-lib/flowy-database2/src/entities/type_option_entities/checkbox_entities.rs
@@ -1,24 +1,32 @@
 use crate::services::field::CheckboxTypeOption;
 use flowy_derive::ProtoBuf;
 
+#[derive(Default, Debug, Clone, ProtoBuf)]
+pub struct CheckboxCellDataPB {
+  #[pb(index = 1)]
+  pub is_check: bool,
+}
+
+impl CheckboxCellDataPB {
+  pub fn new(is_check: bool) -> Self {
+    Self { is_check }
+  }
+}
+
 #[derive(Debug, Clone, Default, ProtoBuf)]
 pub struct CheckboxTypeOptionPB {
   #[pb(index = 1)]
-  pub is_selected: bool,
+  pub config: bool,
 }
 
 impl From<CheckboxTypeOption> for CheckboxTypeOptionPB {
-  fn from(data: CheckboxTypeOption) -> Self {
-    Self {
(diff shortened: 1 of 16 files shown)


frontend/rust-lib/flowy-database2/src/entities/type_option_entities/checkbox_entities.rs
```diff
@@ -1,24 +1,32 @@
 use crate::services::field::CheckboxTypeOption;
 use flowy_derive::ProtoBuf;
 
+#[derive(Default, Debug, Clone, ProtoBuf)]
+pub struct CheckboxCellDataPB {
+  #[pb(index = 1)]
+  pub is_check: bool,
```
Xazin, line 7:
```suggestion
  pub is_checked: bool,
```

frontend/rust-lib/flowy-database2/src/entities/type_option_entities/checkbox_entities.rs
```diff
@@ -1,24 +1,32 @@
...
+  pub fn new(is_check: bool) -> Self {
+    Self { is_check }
+  }
+}
+
 #[derive(Debug, Clone, Default, ProtoBuf)]
 pub struct CheckboxTypeOptionPB {
   #[pb(index = 1)]
-  pub is_selected: bool,
+  pub config: bool,
```
Xazin, line 19: Is config an appropriate name here? It's a boolean...

Xazin (approved):

frontend/rust-lib/flowy-database2/src/entities/type_option_entities/checkbox_entities.rs
richardshiue: Yeah I don't know, it's a dummy field that doesn't do anything.

LucasXu0: Agreed with @Xazin. What does a bool value named "config" signify? Adding a comment to indicate it's a dummy field or renaming it would be helpful.

PR #4704, opened 2024-02-22
Rename workspace

speed2exe:
- Added Cloud API to rename workspace


frontend/rust-lib/flowy-server/src/af_cloud/impls/user/cloud_service_impl.rs
```diff
@@ -311,6 +312,29 @@ where
...
+        .map_err(|_| ErrorCode::InvalidParams)?;
+      let client = try_get_client?;
+      client
+        .patch_workspace(PatchWorkspaceParam {
+          workspace_id,
+          workspace_name: new_workspace_name,
+        })
+        .await?;
+      Ok(())
+    })
```
Xazin, line 335: No reason to wait with the error (`?`) check on `try_get_client` right?
```suggestion
    let client = self.server.try_get_client()?;
    let new_workspace_name = Some(workspace_name.to_string());
    let workspace_id_owned = workspace_id.to_owned();
    FutureResult::new(async move {
      let workspace_id: Uuid = workspace_id_owned
        .parse()
        .map_err(|_| ErrorCode::InvalidParams)?;
      client.patch_workspace(PatchWorkspaceParam {
        workspace_id,
        workspace_name: new_workspace_name,
      })
      .await?;
      Ok(())
    })
```

speed2exe: well, apparently it doesn't work. this async function is required to return a `Future`  that when polled and runs to completion, returns a `Result`, which can either be of success and failure variant.

Issue #4720, opened 2024-02-23
[Bug] Time Parser Fails to Correctly Parse 12-Hour Format


AnsahMohammad:
### Bug Description

The time parser encounters issues when the user opts for the 12-hour format. The current implementation struggles to parse time input in the 'HH:mm a' format.

### How to Reproduce

1. Select the Time format as 12 Hour format (from user settings)
2. Enter the time in the time field "08:00 PM"
Then the time gets parsed as "08:00 AM"

### Expected Behavior

When given input "08:00 PM"
The parser should parse the input to "08:00 PM"

### Operating System

Ubuntu LTS 22.04

### AppFlowy Version(s)

0.4.9

### Screenshots

[Screencast from 24-02-24 12:22:26 AM IST.webm](https://github.com/AppFlowy-IO/AppFlowy/assets/85027826/ca513458-6a29-4d59-b2d1-e8c48510f015)

### Additional Context

From the code, The twelve Hour format is not being recognized by the if condition
![image](https://github.com/AppFlowy-IO/AppFlowy/assets/85027826/4a9a867d-378c-4029-ad01-e6824e83232a)

maxweilun1989:
@richardshiue hi, do you work on this? I debug for a while and found the root cause of this may be the `timeFormat` variable passed in is not the same object of TimeFormatPB.TwelveHour. 
using 
```dart 
timeFormat.toString() == TimeFormatPB.TwelveHour.toString()
``` 
may fix the problem. But I'm not sure if it's the proper way we should do.

PR #4721, opened 2024-02-23
fix: Twelve Hour parsing issue

AnsahMohammad:
This PR closes the Issue #4720 

- [X] Fix issue 4720
- [X] Fix 12hour format bug
- [X] Auto capitalize AM/PM in the formatter
- [x] Set placeholder according to user time format
- [ ] Add tests

### Feature Preview

---

#### PR Checklist


richardshiue:
Hi @AnsahMohammad, please add some tests to cover this functionality to avoid running into the same issue as the last PR.

Xazin:
Hey @AnsahMohammad Do you need help adding a test for this?

AnsahMohammad:
> Hey @AnsahMohammad Do you need help adding a test for this?

Yes, thank you! I've been trying to find the test files related to `mention_date_block`, but I couldn't find them (I later got busy with university exams). Could you help me find the existing test files so that I can extend them?

AnsahMohammad:
@Xazin I also found few additional bugs/ improvements that we can make:

[Screencast from 28-02-24 06:11:14 PM IST.webm](https://github.com/AppFlowy-IO/AppFlowy/assets/85027826/8a7d8638-224b-4bdd-a85b-cf8dc1f8118e)

1. The parser doesn't seem to work properly when input is given as 12:30 AM or 12:30 PM. I suspect this might be due to the time format being "HH:mm a" instead of "hh:mm a" for the 12-hour format.

image

2.  Currently, the parser only works correctly if AM/PM is typed strictly in uppercase. Should we modify it to accept AM/PM in either uppercase or lowercase?
3. Despite my time format being set to the 12-hour format, the placeholder displays the time in the 24-hour format when I click on the time field. Should this be changed?


[Screencast from 28-02-24 06:17:41 PM IST.webm](https://github.com/AppFlowy-IO/AppFlowy/assets/85027826/1197c8c9-ff80-4961-9160-6ddf66228090)

Xazin:
> 1. The parser doesn't seem to work properly when input is given as 12:30 AM or 12:30 PM. I suspect this might be due to the time format being "HH:mm a" instead of "hh:mm a" for the 12-hour format.
> 
>
I would have to look into it to be able to find out the reason why, but you can try to change the format and see if it helps.

> 2. Currently, the parser only works correctly if AM/PM is typed strictly in uppercase. Should we modify it to accept AM/PM in either uppercase or lowercase?
>
>
If we add transforming the string to uppercase to the text formatter, wouldn't that solve this particular issue?

> 3. Despite my time format being set to the 12-hour format, the placeholder displays the time in the 24-hour format when I click on the time field. Should this be changed?
> 
>
Yeah that can be changed, seems like it's something I missed when adding support for User date/time format.

frontend/appflowy_flutter/lib/workspace/presentation/widgets/date_picker/widgets/time_text_field.dart
```diff
@@ -142,9 +152,20 @@ class TimeInputFormatter extends TextInputFormatter {
...
+
+      final minutes = int.tryParse(newValue.text.substring(3, 5));
+      if (minutes == null || minutes <= 0) {
+        return newValue.copyWith(text: '${prefix}00$suffix'.toUpperCase());
+      } else if (minutes > 59) {
+        return newValue.copyWith(text: '${prefix}59$suffix'.toUpperCase());
+      }
+    }
+
+    return newValue.copyWith(text: newText.toUpperCase());
```
Xazin, line 168: Use copyWith otherwise you'll bug the cursor of the field on every new value.

Xazin (approved):

PR #4738, opened 2024-02-26
fix: some mobile database bugs

richardshiue:
- no autofocus on name of field editor
- tapping icon allows changing the type of the field
- applies to both the bottom sheet popup and the full screen editor.
- reduce spacing between toolbar and grid header
- fix a bug where users can't update the name of a database view

screen recording TBD

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/mobile/presentation/database/card/card_detail/widgets/mobile_create_field_button.dart
+++ b/frontend/appflowy_flutter/lib/mobile/presentation/database/card/card_detail/widgets/mobile_create_field_button.dart
@@ -40,7 +40,7 @@ class MobileRowDetailCreateFieldButton extends StatelessWidget {
           LocaleKeys.grid_field_newProperty.tr(),
           fontSize: 15,
         ),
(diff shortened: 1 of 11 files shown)


frontend/appflowy_flutter/lib/mobile/presentation/database/card/card_detail/widgets/mobile_create_field_button.dart
```diff
@@ -40,7 +40,7 @@ class MobileRowDetailCreateFieldButton extends StatelessWidget {
           LocaleKeys.grid_field_newProperty.tr(),
           fontSize: 15,
         ),
-        onPressed: () => showCreateFieldBottomSheet(context, viewId),
+        onPressed: () => mobileCreateFieldWorkflow(context, viewId),
```
richardshiue, line 43: I have no idea what to call this. I call it workflow because it first shows the field type bottom sheet. Upon selection, it then allows you to configure its type option. The field is only created when you click save. There is no bottom sheet called create field.

That being said, mobileCreateFieldWorkflow also sounds super weird. @LucasXu0 @Xazin

PR #4744, opened 2024-02-26
chore: cleanup disk space before running mobile ci

LucasXu0:
### Feature Preview

---

#### PR Checklist


Xazin (approved):
Something I didn't test, but seems NDK R24 is default with the runner, we could possibly just double check if the right environment variables are pointing to it, instead of using the task `nttld/setup-ndk@v1`.

The action downloads and caches the version, but if it already is on disk, then we don't have to do this.

PR #4766, opened 2024-02-28
feat: add user workspace table icon

speed2exe:


frontend/rust-lib/flowy-user-pub/src/session.rs
```diff
@@ -63,6 +63,7 @@ impl<'de> Visitor<'de> for SessionVisitor {
           created_at: Utc::now(),
           // For historical reasons, the database_storage_id is constructed by the user_id.
           workspace_database_object_id: STANDARD.encode(format!("{}:user:database", user_id)),
+          icon: "".to_owned(),
```
Xazin, line 66:
```suggestion
          icon: "".to_string(),
```

Xazin (approved):

frontend/rust-lib/flowy-user-pub/src/session.rs
speed2exe: I actually have no preference and I don't mind at all, I'm curious because it's doing the same thing under the hood, is there a good reason to choose one over the other? P.S. Readability can be subjective.

Xazin: I guess it's not so meaningful anymore, there used to be differences in them. Well of course they're still different, but not when we are talking about converting `&str` into `String`.

I myself prefer `String::new()` or `String::from("abc")` over `"".to_string()` or `"".to_owned()` for readability.

I just saw you use `"".to_string()` in your PR, and made the comment for consistency.

Issue #3060, opened 2023-07-25
[Bug] Links don't open in a browser


maticmeznar:
### Bug Description

I'm a first time user of AppFlowy and I've noticed that none of the external links open in a browser. Same for clicking help and support links from the icon in the bottom right corner. Both AppFlowy and Firefox are running as a Flatpak. I have no problems with links in Obsidian, which also runs as a Flatpak.

When I run AppFlowy from the terminal I get `Jul 25 09:44:40.201 ERROR ThreadId(14) dart_ffi:{"message":"Uncaught platform error"}
{"msg":"Uncaught platform error"}` error each time I click on a link from the help and support icon, but no error when I click "Open link" on a link in a document.

I've also tried it on another computer with Ubuntu 22.04 and Firefox in a snap with the same result.

### How to Reproduce

Click on a link in the help and support menu.

### Expected Behavior

I expect the link to open in my default browser, in a new tab.

### Operating System

Fedora Workstation 38 (x86_64), Ubuntu 22.04 (x86_64)

### AppFlowy Version(s)

0.2.7 (Flatpak)

annieappflowy:
@richardshiue , can you reproduce this issue?

ttollet:
If it helps, I can reproduce this issue.

**Firefox:** system package, version "116.0+linuxmint1+victoria"
**AppFlowy:** flatpak, version 0.2.8
**Operating system:** Linux Mint 21.2 Cinnamon.

richardshiue:
I've narrowed down the problem to this issue. I'll be able to test and fix this after I clear some space on my laptop to build the flatpak

rob32:
I have the same issue here.

Flatpak (v0.4.5).
OS: Fedora 39

apollo13:
I am also running into this. Every click on the link throws:
```
  2024-02-19 13:19:18 ERROR  Uncaught platform error
    at dart-ffi/src/lib.rs:188

  2024-02-19T12:19:18.905514Z ERROR dart_ffi: Uncaught platform error
    at dart-ffi/src/lib.rs:188
```

Flatpak (v0.4.9) on Fedora 39.

apollo13:
A bit of googling suggests that dropping the use of `canLaunch` might fix this (and where needed catch the `PlatformException`)

Xazin:
This should be resolved, feel free to let us know if you are still experiencing issues with launching urls.

Xazin:
[...]

PR #4797, opened 2024-03-03
fix: launch url on all platforms

Xazin:
Closes: #3060 

#### PR Checklist

---

I haven't been able to test that this will work on Fedora / running AF in Flatpak.

Previously we used a helper function imported via appflowy_editor, this helper function should made private and only be used in the editor itself. It's `safeLaunchUrl`.

### DIFF
--- a/frontend/appflowy_flutter/lib/core/helpers/url_launcher.dart
+++ b/frontend/appflowy_flutter/lib/core/helpers/url_launcher.dart
@@ -1,0 +1,38 @@
+import 'package:flutter/services.dart';
+
+import 'package:appflowy_backend/log.dart';
+import 'package:url_launcher/url_launcher.dart' as launcher;
+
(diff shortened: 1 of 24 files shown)


frontend/appflowy_flutter/lib/core/helpers/url_launcher.dart
```diff
@@ -0,0 +1,38 @@
...
+  launcher.LaunchMode mode = launcher.LaunchMode.platformDefault,
+  String? webOnlyWindowName,
+}) async {
+  try {
+    return await launcher.launchUrl(
+      uri,
+      mode: mode,
+      webOnlyWindowName: webOnlyWindowName,
+    );
+  } on PlatformException catch (e) {
```
LucasXu0, line 20: If the `onFailure` callback is null, consider displaying a toast message to inform users of the error.

richardshiue (approved):

PR #4806, opened 2024-03-04
Workspace rename no icon

speed2exe:


frontend/rust-lib/flowy-user/src/entities/workspace.rs
```diff
@@ -116,3 +116,25 @@ pub struct CreateWorkspacePB {
...
+
+#[derive(ProtoBuf, Default, Clone, Validate)]
+pub struct ChangeWorkspaceIconPB {
+  #[pb(index = 1)]
+  #[validate(custom = "required_not_empty_str")]
+  pub workspace_id: String,
+
+  #[pb(index = 2)]
+  #[validate(custom = "required_not_empty_str")]
+  pub new_icon: String,
```
LucasXu0, line 139: the icon should be an optional, because we will display the first character of the name if it's null.

PR #4688, opened 2024-02-21
feat: advanced filters backend logic

richardshiue:
This PR contains only the backend logic. a tech design write up is in the gantt.

Requires https://github.com/AppFlowy-IO/AppFlowy-Collab/pull/159

### Feature Preview

---

#### PR Checklist

- [x] My code adheres to [AppFlowy's Conventions](https://docs.appflowy.io/docs/documentation/software-contributions/conventions)
- [ ] I've listed at least one issue that this PR fixes in the description above.
- [x] I've added a test(s) to validate changes in this PR, or this PR only contains semantic changes.
- [ ] All existing tests are passing.

PR #4807, opened 2024-03-04
refactor: create row workflow

richardshiue:
requires https://github.com/AppFlowy-IO/AppFlowy/pull/4688

- Move some fragmented create row logic from event handler to database editor
- Streamline row duplication logic
- Remove unnecessary Arc\<Field>. The filter, sort and group controllers already are Arcs, and in the instances where we use the struct field, we are just referencing it. So we probably don't need to put it in an Arc.

### Feature Preview

---

#### PR Checklist


appflowy:
LGTM

PR #4837, opened 2024-03-07
fix: relation hotfix

richardshiue:
### Feature Preview

---

#### PR Checklist


frontend/rust-lib/flowy-database2/src/services/field/type_options/relation_type_option/relation.rs
```diff
@@ -60,7 +60,7 @@ impl CellDataChangeset for RelationTypeOption {
     let cell_data: RelationCellData = cell.unwrap().as_ref().into();
     let mut row_ids = cell_data.row_ids.clone();
     for inserted in changeset.inserted_row_ids.iter() {
-      if row_ids.iter().any(|row_id| row_id == inserted) {
+      if !row_ids.iter().any(|row_id| row_id == inserted) {
```
LucasXu0, line 63: why don't we use contains here?

PR #4906, opened 2024-03-15
fix: apply formatting to pasted text only when the block is a paragraph

### LINKED ISSUE #4905: [FR] When pasting text into an empty blcok with formatting, is it possible to not replace the block style?
### Description

When pasting text into an empty blcok with formatting, is it possible to not replace the block style?

For example, in an empty H1, pasting text will be replaced with text style. I want it to maintain the style of H1. The same is true for other formats, which is more in line with most people's operating habits.

### Impact

The empty block style will not change into the format I don’t want.

### Additional Context

_No response_

q200892907:
### Feature Preview

fix #4905 

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/editor_state_paste_node_extension.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/editor_state_paste_node_extension.dart
@@ -13,8 +13,7 @@ extension PasteNodes on EditorState {
     }
     final transaction = this.transaction;
     final insertedDelta = insertedNode.delta;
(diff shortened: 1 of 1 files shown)


frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/editor_state_paste_node_extension.dart
```diff
@@ -14,7 +14,7 @@ extension PasteNodes on EditorState {
     final transaction = this.transaction;
     final insertedDelta = insertedNode.delta;
     // if the node is empty, replace it with the inserted node.
```
LucasXu0, line 16: Please update the comment too.

PR #4916, opened 2024-03-15
chore: update sv-SE.json

Pingw1nn:
Describe the changes you have conducted here

Preview the messages on https://inlang.com/github.com/Pingw1nn/AppFlowy .


LucasXu0:
@Pingw1nn The sv-SE.json file already exists in the codebase. Could you please update the translation directly in sv-SE.json instead of creating a new one?

PR #4947, opened 2024-03-20
feat: support error for limit exceed

speed2exe:
- Use FlowyError for most cloud apis
- Support for `WorkspaceLimitExceeded` and `WorkspaceMemberLimitExceeded` error


frontend/rust-lib/flowy-error/src/impl_from/cloud.rs
```diff
@@ -21,6 +21,11 @@ impl From<AppResponseError> for FlowyError {
       AppErrorCode::NotEnoughPermissions => ErrorCode::NotEnoughPermissions,
       AppErrorCode::NetworkError => ErrorCode::HttpError,
       AppErrorCode::PayloadTooLarge => ErrorCode::CloudRequestPayloadTooLarge,
+      AppErrorCode::UserUnAuthorized => match &*error.message {
```
LucasXu0, line 24: It can be a temporary solution. Refactor it and use code instead of message as the match condition in the next version.

speed2exe: Got it. Doing that in the next release.

Issue #4968, opened 2024-03-23
[FR] directly edit the page title in place in the side panel


annieappflowy:
### Description

<img width="242" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/12026239/1660e66e-538c-4821-9d5d-34d4d4cf91bd">
 Without using ':', I'd like to edit the page title directly by double clicking 

### Impact

AppFlowy users

Omegaviv:
@annieappflowy I would like to work on this issue.

Only code changes on Flutter side is required here, so should I use the same that is used within the "rename" option inside of ":" set of options right?

inson1:
Could that by optional? By setting it in settings? I dont care what the default will be.

Xazin:
We have the F2 shortcut to rename current page, I know this is not equivalent though.

I'm not a huge fan of double click shortcuts, as it means single clicks will experience a delay before being triggered, as the engine will wait for some amount of milliseconds to see if there will be a second tap.

annieappflowy:
@inson1 , I'd like to know why you'd like it to be disabled. If you don't use this way to rename, won't it affect other existing actions?

moolight-seashell:
that would be nice to ask a node name evrytime we creating a new node

annieappflowy:
@moolight-seashell , you can turn on this feature by going to Settings -> Appearance
<img width="1320" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/12026239/97032aa1-bcf0-4746-8b84-57a56b1b9ab7">

moolight-seashell:
srry, thanks for the info

PR #5000, opened 2024-03-28
feat: double tab edit side panel 

Omegaviv:
This PR is for issue #4968, to edit side panel title with double click. 

I directly used GestureDetector of SingleInnerViewItem, I could have also passed it as function call back but it will require having MultiBlocProvider.

I am new here, any feedback @annieappflowy ?

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/view_item.dart
+++ b/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/view_item.dart
@@ -355,17 +355,19 @@ class _SingleInnerViewItemState extends State<SingleInnerViewItem> {
       behavior: HitTestBehavior.translucent,
       onTap: () => widget.onSelected(widget.view),
       onTertiaryTapDown: (_) => widget.onTertiarySelected?.call(widget.view),
-      onDoubleTap: isSelected ? (){
-        NavigatorTextFieldDialog(
-          title: LocaleKeys.disclosureAction_rename.tr(),
-          autoSelectAllText: true,
-          value: widget.view.name,
-          maxLength: 256,
-          onConfirm: (newValue, _) {
-            context.read<ViewBloc>().add(ViewEvent.rename(newValue));
-          },
-        ).show(context);
-      } : null,
+      onDoubleTap: isSelected
+          ? () {
+              NavigatorTextFieldDialog(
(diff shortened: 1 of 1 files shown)


LucasXu0:
@Omegaviv  I suggest supporting renaming only through a double-tap when the view is selected. This is because there's a minimum waiting time for the double-click gesture, which could cause a delay when opening another view with a single click.

Omegaviv:
@LucasXu0 If we just let them use double click edit option only to select views, it won't be of much use as a user will have to do three clicks to edit other than the selected view. How about a long press?

LucasXu0:
@Omegaviv The long press action for renaming is not user-friendly. You can refer to the [implementation](https://github.com/AppFlowy-IO/appflowy-editor/blob/b927ec0685c870c731c5b6d9688a031d0cd31e76/lib/src/service/selection/selection_gesture.dart#L4) of the gesture conflict handler for a solution. This widget triggers the onTap event immediately, without waiting for the doubleTap timeout.

Omegaviv:
> @Omegaviv I suggest supporting renaming only through a double-tap when the view is selected. This is because there's a minimum waiting time for the double-click gesture, which could cause a delay when opening another view with a single click.

@LucasXu0 I have added a simple check

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/view_item.dart
```diff
@@ -355,6 +355,17 @@ class _SingleInnerViewItemState extends State<SingleInnerViewItem> {
...
+        NavigatorTextFieldDialog(
+          title: LocaleKeys.disclosureAction_rename.tr(),
+          autoSelectAllText: true,
+          value: widget.view.name,
+          maxLength: 256,
+          onConfirm: (newValue, _) {
+            context.read<ViewBloc>().add(ViewEvent.rename(newValue));
+          },
+        ).show(context);
+      } : null,
```
Xazin, line 368: Seems like you didn't run the formatter.

Omegaviv:
@LucasXu0 can you close this, if there are no issues?

LucasXu0:
> @LucasXu0 can you close this, if there are no issues?

Why do we close it? I think we can support temporarily editing the current view name by double-clicking it.

Issue #3577, opened 2023-10-02
[FR] Copy markdown to clipboard, as a share option.


HatWearingMann:
### Description

The share button should display two options after pressing it, one for saving the document as a file, and the other for copying the document to the clipboard, still as markdown, of course. 

### Impact

Adding an option for sharing that would simply copy the markdown version of the document, instead of only being able to save it as a file, would make it extra convenient for sharing the document on services such as GitHub Gist.

emmggi:
I think we should also add share as HTML (in file) since we have `documentToHTML()`.

StarTrooper08:
Hey! can I work on this issue?

AdiAr11:
Hi @MayurSMahajan, Can i work on this?
Thank you

MayurSMahajan:
Sure, @AdiAr11 you can work on this.

AdiAr11:
@MayurSMahajan Thank you

AdiAr11:
Hi @Xazin, yes, will try to create a pr this weekend.

HatWearingMann:
> Hi @Xazin, yes, will try to create a pr this weekend.

Heya, status update?

HatWearingMann:
> Hi, I actually upgraded Flutter version, and this project currently does not work with it. Was trying to get it working. And got busy with work. Will try to finish it this weekend.

I hope it's not rude to bug you with this :stuck_out_tongue_winking_eye:

AdiAr11:
> > Hi, I actually upgraded Flutter version, and this project currently does not work with it. Was trying to get it working. And got busy with work. Will try to finish it this weekend.
> 
> I hope it's not rude to bug you with this :stuck_out_tongue_winking_eye: 

No no, its fine, and I'm really sorry, got busy with work. Will try to complete this soon.

HatWearingMann:
> No no, its fine, and I'm really sorry, got busy with work. Will try to complete this soon.

bump

Lagrange1813:
Hi, @MayurSMahajan 
I noticed that this issue had been previously assigned to someone else, but it appears that it hasn't been resolved yet. I would like to offer my assistance in completing this issue if it is still open.
Thank you.

PR #5021, opened 2024-03-31
feat: expand share option

Lagrange1813:
### Feature Preview

https://github.com/AppFlowy-IO/AppFlowy/assets/54029053/928e0fa4-4ebb-4407-99e0-8a2464cb1796

Closes [#3577](https://github.com/AppFlowy-IO/AppFlowy/issues/3577)

---

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/document/application/document_share_bloc.dart
```diff
@@ -11,43 +11,82 @@ import 'package:freezed_annotation/freezed_annotation.dart';
...
-  DocumentShareBloc({required this.view})
-      : super(const DocumentShareState.initial()) {
-    on<ShareMarkdown>(_onShareMarkdown);
-  }
+  DocumentShareBloc({
+    required this.view,
+  }) : super(const DocumentShareState.initial()) {
+    on<DocumentShareEvent>((event, emit) async {
+      await event.when(
+        shareMarkdown: (path) async {
```
Xazin, line 19: Seems we can consolidate this into just one event, add the `DocumentExportType` as an event parameter.

Same with the `_saveMarkdownToPath` and `_saveHtmlToPath`, their only difference is the ExportType (Which you know now by the `DocumentExportType`.

LucasXu0: I removed the shareXXX event and replaced it with share(type, path?) instead.

PR #5039, opened 2024-04-02
refactor: cell data transform logic

### LINKED ISSUE #5003: [Bug] Sorts and Filters not working for transformed fields
### Bug Description

When cell data is entered into a cell such that it is no longer `None`, changing the field type of that field and then creating a new filter on that field will hide any row that should satisfy the filter requirements. 

### How to Reproduce

1. in a grid, create some empty rows
2. create a checkbox field and select/unselect some rows
3. change the type of the field to single select, observe that now the cell which were checked are now `Yes` and those that were unchecked are `No`.
4. Create a filter on this single select, and set to either checked or unchecked
5. Observe that no rows are visible.

### Expected Behavior

The filter logic should be correct

### Operating System

macOS

### AppFlowy Version(s)

0.5.3

### Screenshots

_No response_

### Additional Context

_No response_

richardshiue:
resolves #5003

Also resolves another bug where after changing the field type of a checkbox to single/multi-select, clicking on a is_some cell will not show the "Yes" or "No" option as selected.

This is the behavior on `main`:

https://github.com/AppFlowy-IO/AppFlowy/assets/71320345/1ae3604d-aae4-42bc-ae59-f78a8c9e9aea

This PR simplifies the logic of getting cell data for all kinds of purposes: filtering, sorting, convert to protobuf, handle to string, handle is empty, etc.

I achieve this by getting the `TypeOption::CellData` correctly, in one central location, and NOT passing the field type to all the handlers.

- if the field type of the `Cell` matches the field type of the field, decode it directly and return
- if not, see if the field type of the `Cell` allows for a transformation to the field type of the field. Afterwards, decode and transform the `Cell` into the field type of the field.
- otherwise, return `None`, `Err`, or `Default::default` depending on the requirements.

After getting the cell data, continue with whatever business logic that is required.

### Feature Preview

---

#### PR Checklist


frontend/rust-lib/flowy-database2/src/services/field/type_options/url_type_option/url_type_option.rs
```diff
@@ -61,28 +61,14 @@ impl TypeOptionCellDataSerde for URLTypeOption {
...
-    &self,
-    cell: &Cell,
-    decoded_field_type: &FieldType,
-    _field: &Field,
-  ) -> FlowyResult<<Self as TypeOption>::CellData> {
-    if !decoded_field_type.is_url() {
-      return Ok(Default::default());
-    }
-
+  fn decode_cell(&self, cell: &Cell) -> FlowyResult<<Self as TypeOption>::CellData> {
```
Xazin, line 64: Seems like we can remove `decode_cell` in favor of `parse_cell` in all type options with your changes.

Xazin (approved):

frontend/rust-lib/flowy-database2/src/services/field/type_options/url_type_option/url_type_option.rs
richardshiue: Yeah... I don't know why make the distinction when implementing it much early on. If it's okay, I'll do it in a follow-up PR.

frontend/rust-lib/flowy-database2/src/services/calculations/service.rs
```diff
@@ -130,50 +128,39 @@ impl CalculationsService {
...
-                .is_empty())
-            .collect::<Vec<_>>()
-            .len()
-        );
-      }
+    match TypeOptionCellExt::new(field, None).get_type_option_cell_data_handler() {
+      Some(handler) if !row_cells.is_empty() => row_cells
+        .iter()
+        .filter(|row_cell| {
+          if let Some(cell) = &row_cell.cell {
```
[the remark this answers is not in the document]
richardshiue: I tried it and I had to add an additional `as_ref` so it didn't come out as clean as I wanted

image

LucasXu0 (approved):

PR #5070, opened 2024-04-05
feat: keyboard navigation for relation cell editor

richardshiue:
### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/database/widgets/cell_editor/relation_cell_editor.dart
```diff
@@ -14,126 +16,252 @@ import '../../application/cell/bloc/relation_row_search_bloc.dart';
...
+  late final FocusNode focusNode;
+  late final bloc = RelationRowSearchBloc(
+    databaseId: widget.relatedDatabaseMeta.databaseId,
+  );
+
+  @override
+  void initState() {
+    super.initState();
+    focusNode = FocusNode(
+      onKeyEvent: (node, event) {
```
Xazin, line 62: I feel like this is overengineered..

IIRC ListView offers this type of focus traversal out of the box.. Just needs to have focus first.. (Eg. you could just build the arrow down key event from the search field, and omit the rest?)

richardshiue: I'm not sure how to do that without having the text field lose focus after arrow up or down. Can you show me an example?

Xazin: I see what you mean, but that's non-standard behavior no? Is that really what we want to do?

And if so, I'd rather see a common implementation that we can reuse if that's the case.

richardshiue: What is standard behavior, do you mean the focus traversal that you'd get with tab and shift tab? If so, it's very common to deviate from this when catering to arrow-key-focused UX.

I'd be down for an abstraction at the widget level if it's not going to make the code more complex. We would only have 2 implementations so far including this one. If that number keeps going up, the abstraction would definitely be helpful.

frontend/appflowy_flutter/lib/plugins/database/application/cell/bloc/relation_row_search_bloc.dart
```diff
@@ -24,11 +26,22 @@ class RelationRowSearchBloc
       (event, emit) {
         event.when(
           didUpdateRowList: (List<RelatedRowDataPB> rowList) {
-            allRows.clear();
-            allRows.addAll(rowList);
-            emit(state.copyWith(filteredRows: allRows));
+            allRows
```
LucasXu0, line 29: Does it make sense to move the `allRows` field to the `RelationRowSearchState` class?

frontend/appflowy_flutter/lib/plugins/database/application/cell/bloc/relation_row_search_bloc.dart
```diff
@@ -45,12 +58,56 @@ class RelationRowSearchBloc
 
   void _updateFilter(String filter, Emitter<RelationRowSearchState> emit) {
     final rows = [...allRows];
+
     if (filter.isNotEmpty) {
       rows.retainWhere(
-        (row) => row.name.toLowerCase().contains(filter.toLowerCase()),
+        (row) =>
+            row.name.toLowerCase().contains(filter.toLowerCase()) ||
```
LucasXu0, line 65: Adding brackets to wrap the conditions would make it clearer.

`( A || B) && C`

frontend/appflowy_flutter/lib/plugins/database/application/cell/bloc/relation_row_search_bloc.dart
```diff
@@ -45,12 +58,56 @@ class RelationRowSearchBloc
...
+      return;
+    }
+
+    final currentIndex = rowIds.indexWhere((id) => id == state.focusedRowId);
+
+    final newIndex = currentIndex == -1
+        ? 0
+        : (currentIndex + (previous ? -1 : 1)) % rowIds.length;
+
+    emit(state.copyWith(focusedRowId: rowIds[newIndex]));
```
LucasXu0, line 110:
```suggestion
  void _focusOption(bool previous, Emitter<RelationRowSearchState> emit) {
    if (state.filteredRows.isEmpty) {
      return;
    }

    final rowIds = state.filteredRows.map((e) => e.rowId).toList();
    final currentIndex = state.focusedRowId == null
        ? -1
        : rowIds.indexWhere((id) => id == state.focusedRowId);

    // If the current index is -1, it means that the focused row is not in the list of row ids.
    // In this case, we set the new index to the last index if previous is true, otherwise to 0.
    final newIndex = currentIndex == -1
        ? (previous ? rowIds.length - 1 : 0)
        : (currentIndex + (previous ? -1 : 1)) % rowIds.length;

    emit(state.copyWith(focusedRowId: rowIds[newIndex]));
```

frontend/appflowy_flutter/lib/plugins/database/application/cell/bloc/relation_row_search_bloc.dart
```diff
@@ -61,17 +118,23 @@ class RelationRowSearchEvent with _$RelationRowSearchEvent {
...
+      _FocusPreviousOption;
+  const factory RelationRowSearchEvent.focusNextOption() = _FocusNextOption;
 }
 
 @freezed
 class RelationRowSearchState with _$RelationRowSearchState {
   const factory RelationRowSearchState({
-    required String filter,
     required List<RelatedRowDataPB> filteredRows,
+    required String? focusedRowId,
```
LucasXu0, line 133: Optional: Use @Default()

frontend/appflowy_flutter/lib/plugins/database/application/row/related_row_detail_bloc.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  @override
+  Future<void> close() {
+    state.when(
+      loading: () {},
+      ready: (databaseController, rowController) {
+        rowController.dispose();
+        databaseController.dispose();
+      },
+    );
```
LucasXu0, line 31:
```suggestion
    state.whenOrNull(
      ready: (databaseController, rowController) {
        rowController.dispose();
        databaseController.dispose();
      },
    );
```

frontend/appflowy_flutter/lib/plugins/database/application/row/related_row_detail_bloc.dart
```diff
@@ -0,0 +1,126 @@
...
+                ),
+              );
+            },
+          );
+        },
+      );
+    });
+  }
+
+  void _init(String databaseId, String initialRowId) async {
```
LucasXu0, line 63: The initial logic looks complex. could you add comments to describe the steps?

frontend/appflowy_flutter/lib/plugins/database/application/row/related_row_detail_bloc.dart
```diff
@@ -0,0 +1,126 @@
...
+    final getDatabaseResult = await DatabaseEventGetDatabases().send();
```
LucasXu0, line 64: By the way, I have added fold support to the asynchronous function.

You can use `await DatabaseEventGetDatabases().send().fold(...)`;

frontend/appflowy_flutter/lib/plugins/database/application/row/related_row_detail_bloc.dart
```diff
@@ -0,0 +1,126 @@
...
+    );
+    add(
+      RelatedRowDetailPageEvent.didInitialize(
+        databaseController,
+        rowController,
+      ),
+    );
+  }
+
+  void _startListening(DatabaseController databaseController) {
```
LucasXu0, line 106: No need to wrap it in a function.

frontend/appflowy_flutter/lib/plugins/database/application/cell/bloc/relation_row_search_bloc.dart
LucasXu0: If so, the previous condition is wrong. Because these two operators are executed from left to right.

PR #5111, opened 2024-04-10
feat: open a row as a full page

richardshiue:
Screenshot 2024-04-16 at 11 16 24 PM

### Feature Preview

---

#### PR Checklist


annieappflowy:
should provide a way for the user to come back to the database
better to show the breadcrumbs at the top so we know where we are at

PR #5131, opened 2024-04-14
feat: add codeblock package

MayurSMahajan:
### Feature Preview

Adds the Codeblock package to AppFlowy.

---

#### PR Checklist


Xazin:
Can you upload a feature preview?

Also seems like flutter tests are broken.

MayurSMahajan:
@Xazin the tests are failing due to the package not getting compiled. How to resolve this?

Xazin:
There are some discrepancies from the code block package and the existing code block:
- Line numbers
- The UI for the actions (import, copy, etc.) does not conform to AppFlowy. Also not sure why there is an icon change when pressing an action, does not conform to AppFlowy UI
- Update to use the new indentation command handler _(Generally if the commands changed in AppFlowy, if the package does not update to the same commands, then that will cause a regression and other issues will have to be reopened)_
- How do we update localization/copy for the Code block?
- Cursor color is wrong

About the UI, I know Lucas made the same comment in a previous PR:
Screenshot 2024-04-14 at 22 31 39

---

Also I checked out the branch to look at the other changes, the language dropdown and dropdown menu needs to be the same as it currently is on `main` to conform to the rest of the UI.

See this video (Also sheds light on some other issues):

https://github.com/AppFlowy-IO/AppFlowy/assets/42929161/125d2939-e5fc-42b4-ac44-09cfaedd8821

It also looks less than great in dark mode. I haven't tested on different themes.

---

Some more images:

Screenshot 2024-04-14 at 22 59 28
Screenshot 2024-04-14 at 22 59 00

MayurSMahajan:
Thanks @Xazin, for your feedback. I will work on them asap.

Xazin:
What do you think about letting me move the code block to our plugins repository, and then we move the import/export functionality from your code to that repository?

It will make the workload on you a lot simpler, as it would be a conflict-free transition.

Let me know. :pray:

PR #5169, opened 2024-04-20
fix: search launch review

Xazin:
#### PR Checklist


annieappflowy:
I have not checked this new change. If you haven't added a hover tip on the search button, it would be better to one to inform the user of the search shortcut.

frontend/appflowy_flutter/lib/plugins/inline_actions/handlers/inline_page_reference.dart
```diff
@@ -48,145 +47,100 @@ class InlinePageReferenceService extends InlineActionsDelegate {
...
+    // Filter by viewLayout
+    views.retainWhere(
+      (i) =>
+          currentViewId != i.id &&
+          (viewLayout == null || i.layout == viewLayout),
     );
 
-    _user = userProfile.fold((s) => s as UserProfilePB, (e) => null);
+    // Map to InlineActionsMenuItem, then take 5 items
+    return _recentViews = views.map(_fromView).take(5).toList();
```
LucasXu0, line 79: shouldn't use `limitResults` here?

frontend/appflowy_flutter/lib/startup/startup.dart
```diff
@@ -136,6 +138,7 @@ class FlowyRunner {
           if (isAppFlowyCloudEnabled) InitAppFlowyCloudTask(),
           const InitAppWidgetTask(),
           const InitPlatformServiceTask(),
+          getIt<CachedRecentService>(),
```
LucasXu0, line 141: should not use it here. Create a launch task, wrap it, and dispose of the sources if needed.

LucasXu0:
The cause of the failed tests is that the time for creating or removing a page has increased, so we have to increase the block response wait time for these tests. @Xazin, any changes to the creation or deletion of pages?

PR #5210, opened 2024-04-26
fix: some database bugs

### LINKED ISSUE #5209: [Bug]  [create at] cells no more auto setting date in 0.5.5
### Bug Description

in new version 0.5.5 in database [create at] cells not auto setting date for creating new rows?

### How to Reproduce

by creating a new database using [create at] type cell

### Expected Behavior

find out the problem and fix it

### Operating System

windows10 professional

### AppFlowy Version(s)

0.5.5

### Screenshots

_No response_

### Additional Context

_No response_

richardshiue:
support

image

resolves #5209

also fix another bug where all the toasts in the app aren't working.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database/widgets/cell/editable_cell_skeleton/url.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database/widgets/cell/editable_cell_skeleton/url.dart
@@ -4,7 +4,6 @@ import 'dart:io';
 import 'package:appflowy/generated/flowy_svgs.g.dart';
 import 'package:appflowy/generated/locale_keys.g.dart';
 import 'package:appflowy/mobile/presentation/widgets/flowy_mobile_quick_action_button.dart';
-import 'package:appflowy/core/helpers/url_launcher.dart';
 import 'package:appflowy/plugins/database/application/cell/cell_controller.dart';
 import 'package:appflowy/plugins/database/application/cell/cell_controller_builder.dart';
 import 'package:appflowy/plugins/database/application/database_controller.dart';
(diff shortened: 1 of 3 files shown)


frontend/appflowy_flutter/lib/plugins/database/widgets/cell/editable_cell_skeleton/url.dart
```diff
@@ -214,29 +214,19 @@ class MobileURLEditor extends StatelessWidget {
...
+      await InternetAddress.lookup(uri.host);
+    }
   } catch (_) {
-    url = "https://www.google.com/search?q=${Uri.encodeComponent(content)}";
+    uri = Uri.parse(
+      "https://www.google.com/search?q=${Uri.encodeComponent(content)}",
+    );
   } finally {
-    await afLaunchUrlString(url);
+    await launchUrl(uri);
```
Xazin, line 230:
```suggestion
    await afLaunchUrl(uri);
```

Xazin (approved):

frontend/appflowy_flutter/lib/plugins/database/widgets/cell/editable_cell_skeleton/url.dart
richardshiue: @Xazin @LucasXu0 While I was implementing this logic, I didn't like the behavior with `afLaunchUrl`'s `addingHttpSchemeWhenFailed` in certain cases. If I don't provide a scheme and I call it after parsing it, like below

```
final uri = Uri.parse('instagram.com');
await afLaunchUrl(uri, addingHttpSchemeWhenFailed: true);
```

`afLaunchUrl` tries to immediately launch it. When that fails or returns false, it then adds the http scheme and retry, resulting in this even though on the second try the link launches successfully.

image

So here in this function, I did the handling myself and removed the need for `afLaunchUrl`. I also needed to support searching with google if the link cannot open.

How should I proceed? Should I make any changes to `afLaunchUrl?

LucasXu0: > I also needed to support searching with google if the link cannot open.

This logic should not be included in `afLaunchUrl` function. I have reviewed your changes. It's ok to keep your logic.

> afLaunchUrl tries to immediately launch it. When that fails or returns false, it then adds the http scheme and retry, resulting in this even though on the second try the link launches successfully.

I suggest skipping the first return if addingHttpSchemeWhenFailed is set to true.

Issue #4771, opened 2024-02-28
[Bug] Wrong material ink color on new group button on mobile


Xazin:
### Bug Description

When pressing on `+ New Group` the ink color does not match any other inks in the application theme-wise.

Also, the localized value should be `New group` and not `New Group`.

Lastly, when pressing on the button a small card appears where you can enter a title and add a group. The `x` (clear) button for the TextField has a splash on pressed, but it goes behind the card. Also should it even be displayed or clickable when TextField is already empty?

### How to Reproduce

1. Create a Board on Mobile
2. Press and hold `+ New Group` button and see ink color is dark instead of appflowy blue
3. Press `+ New Group` and then press and hold the `x` (clear) button to see the splash go behind the card

### Expected Behavior

1. Better ink color on `+ New group`
2. Change translated value to `New group`
3. Fade-in-out the clear button or make it opaque when TextField is empty
4. Splash radius shows above card just underneath the icon button (the clear button)

### Operating System

iOS

### AppFlowy Version(s)

0.5.0

### Screenshots

https://github.com/AppFlowy-IO/AppFlowy/assets/42929161/48c06898-4c14-491e-8c7f-fde78d336458

ChiragKr04:
Hey, please assign this to me. Thanks

PR #5214, opened 2024-04-26
Fix/new group clear splash

ChiragKr04:
fixes https://github.com/AppFlowy-IO/AppFlowy/issues/4771


richardshiue (approved):
LGTM except that it doesn't implement the "don't show clear button when text field is empty" requirement. Not a dealbreaker for me though.

Xazin:
I've resolved the rest of the ticket in this PR, I'll merge once pipelines go green.

PR #5335, opened 2024-05-14
Update German translations

suntorytimed:
Updated German translation


suntorytimed:
@LucasXu0 @Xazin might make sense to include this PR in 0.5.8 as the translation mostly targets the most recent menu changes in this release.

LucasXu0:
@suntorytimed Sure. We can include the translations in the upcoming release.

PR #5349, opened 2024-05-16
feat: timer field

### LINKED ISSUE #4083: [FR] Time field type
### Description

A field type for determining time like 20 minutes, 3 hours, etc. It's useful for task management purposes, To show how much time a task needs or how much I've spent time on the task.

### Impact

Users who use the Kanban or grid view for managing tasks or similar stuff.

### Additional Context

I would like to implement this my self if its an acceptable feature.

zoli:
Timer field type which stores time in minutes. More details about this feature on the related issue #4083.
Resolves #4083.

#### PR Checklist


zoli:
https://github.com/AppFlowy-IO/AppFlowy/assets/3286303/19c5d423-9f5b-4d12-8940-8a3f9dd62ea2

frontend/rust-lib/flowy-database2/src/services/field/type_options/timer_type_option/timer.rs
```diff
@@ -0,0 +1,171 @@
...
+    Ok(TimerCellData::from(cell))
+  }
+}
+
+impl TimerTypeOption {
+  pub fn new() -> Self {
+    Self
+  }
+
+  pub(crate) fn format_cell_data(minutes: i64) -> String {
```
Xazin, line 63: For this use-case, I'm wondering if we shouldn't do the parsing of the cell in the frontend itself.

Using Intl I believe we can do some pretty nice transformation, including support for different localizations.

See eg. https://github.com/andresaraujo/timeago.dart/blob/master/packages/timeago/lib/src/timeago.dart

Xazin:
Also something I was thinking about when I saw "Timer", is that I was expecting it to count down. Otherwise I guess calling it "Time" is more relevant.

frontend/rust-lib/flowy-database2/src/services/field/type_options/timer_type_option/timer.rs
zoli: I wasn't sure about this either but doesn't this make it prone to inconsistency between different frontends?

zoli:
> Also something I was thinking about when I saw "Timer", is that I was expecting it to count down. Otherwise I guess calling it "Time" is more relevant.

You are right about timer but I think maybe stopwatch is a better name. I also want to add time tracking feature to this later and it works like count up.

zoli:
Thinking about this, It can be a general field type (I'm not sure about the name) which the user can choose to be a timer (count down) or stopwatch (count up) in type options. What do you think @Xazin?

Xazin:
> Thinking about this, It can be a general field type (I'm not sure about the name) which the user can choose to be a timer (count down) or stopwatch (count up) in type options. What do you think @Xazin?

Yeah I that is a good idea. Kind of like selecting a number format.

zoli:
So I will change the field type name to Time and will add a stopwatch and timer option later in a separate PR.

frontend/rust-lib/flowy-database2/src/services/field/type_options/timer_type_option/timer.rs
Xazin: You're right, but I think in this case it's the better solution.

In the frontend we have locale information, as well as a better Intl library so we can much easilier parse the time into eg. `5y 100d 2h 31m`. With timestamps we also do parsing on the frontend for dates, we also are able to get timezone information on the frontend quite easily.

In Danish the equivalent would be `5år 100d 2t 31m`.

Xazin:
> So I will change the field type name to Time and will add a stopwatch and timer option later in a separate PR.

Sounds good, feel free to ping me when you have renamed the field, I tested and feature works as expected. 👍

I think the time type data can be in minutes, but for Stop timer or Count down timer, it should be a start timestamp.

zoli:
@Xazin I renamed the field type to time and moved the formatting function to the flutter frontend.
I will implement these features on other PRs:
- Having precision option (store by minutes or second)
- Timer and Stopwatch as type option
- More localization support as per your review comment.
- Tracking time button for timer and stopwatch

PR #5362, opened 2024-05-19
chore: do not use the cell of given summary field

appflowy:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/src/manager.rs
+++ b/frontend/rust-lib/flowy-database2/src/manager.rs
@@ -424,7 +424,8 @@ impl DatabaseManager {
     if let Some(row) = database.get_row(&view_id, &row_id) {
       let fields = database.get_fields(&view_id, None);
       for field in fields {
-        // when summarizing row, we don't need to summarize the cell of the field that we want to summarize.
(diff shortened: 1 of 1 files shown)


frontend/rust-lib/flowy-database2/src/manager.rs
```diff
@@ -424,8 +424,11 @@ impl DatabaseManager {
     if let Some(row) = database.get_row(&view_id, &row_id) {
       let fields = database.get_fields(&view_id, None);
       for field in fields {
-        if let Some(cell) = row.cells.get(&field.id) {
-          summary_row_content.insert(field.name.clone(), stringify_cell(cell, &field));
+        // when summarizing row, we don't need to summarize the cell of the field that we want to summarize.
```
Xazin, line 427:
```suggestion
        // when summarizing a row, we only need the cell of the specific field
```

Xazin (approved):

frontend/rust-lib/flowy-database2/src/manager.rs
appflowy: I've updated it

PR #5366, opened 2024-05-19
fix: chaning field type #5360

### LINKED ISSUE #5360: [Bug] changing field type from text to checkbox causes exception
### Bug Description

In board view changing field type from text to checkbox throws an exception when you hover the cursor on the card.
`'CellController<CheckboxCellDataPB, String>' is not a subtype of type 'CellController<String, String>' in type`

### How to Reproduce

Open a card and create new field which its type is text by default. Change the field type to checkbox and close the card. Hover on any card and you get exception.
converting any field from `CellController<String, String>` to a `CellController<SomeCellDataPB, String>` causes exception.

### Expected Behavior

Not throwing any exception.

### Operating System

Linux

### AppFlowy Version(s)

v0.5.7

### Screenshots

_No response_

### Additional Context

_No response_

zoli:
Before this changing any field from CellController<String, String> to a CellController<SomeCellDataPB, String> caused an exception. The root of the problem was that _CardContent wasn't getting rebuilt when field type got changed. In CardState it was storing a list of CellContext which contains only field id and row id so on changing field type the state wasn't changing.
I made a new type CellMeta which stores field type alongside row id and field id and now CardState stores list of CellMeta.

Fixes #5360

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database/widgets/card/card_bloc.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database/widgets/card/card_bloc.dart
@@ -1,5 +1,7 @@
 import 'dart:async';
 
+import 'package:appflowy/plugins/database/application/row/row_service.dart';
+import 'package:appflowy_backend/protobuf/flowy-database2/protobuf.dart';
 import 'package:flutter/foundation.dart';
 
 import 'package:appflowy/plugins/database/application/cell/cell_controller.dart';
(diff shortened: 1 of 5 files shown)


Xazin (approved):

frontend/appflowy_flutter/lib/plugins/database/widgets/card/card_bloc.dart
```diff
@@ -116,33 +117,54 @@ List<CellContext> _makeCells(
         !(fieldInfo.visibility?.isVisibleState() ?? false) ||
         (groupFieldId != null && cellContext.fieldId == groupFieldId);
   });
-  return cellContexts.toList();
+  return cellContexts
+      .map(
+        (cellCtx) => CellMeta(
+          fieldId: cellCtx.fieldId,
+          rowId: cellCtx.rowId,
+          fieldType: fieldController.getField(cellCtx.fieldId)!.fieldType,
```
LucasXu0, line 125: A future improvement: We are force unwrapping the value from `getField` everywhere. This doesn't make sense. If we are sure the value must not be null, we should change the return type of the `getField` function.

Issue #5434, opened 2024-05-29
[Bug] the Settings window layer in between the editing menu and the document


annieappflowy:
### Bug Description

<img width="869" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/12026239/fc63d4cc-1b33-470a-8fcf-d458a278b375">

### How to Reproduce

Highlight text to trigger the editing menu
Open up Settings

### Expected Behavior

the editing menu should be behind the Settings window

### Operating System

macOS

### AppFlowy Version(s)

v0.5.8

Issue #4970, opened 2024-03-23
[Bug] Tools popup is on behalf of the Setting window


unbreakabl3:
### Bug Description

If the size of the app relatively small, the tools popup window is overriding the Settings window

### How to Reproduce

Make the app main window small, write some text, select it, wait until the tools will appear and open the app settings.

### Expected Behavior

The text editor tools should disappear

### Operating System

MacOS

### AppFlowy Version(s)

0.5.2

### Screenshots

<img width="932" alt="image" src="https://github.com/AppFlowy-IO/AppFlowy/assets/13556249/f89c0841-cc0a-428f-b3ef-5bcaa44393f2">

PR #5443, opened 2024-05-30
fix: launch review 0.5.9

Xazin:
Relates: #4970 #5434

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/video/upload_video_menu.dart
```diff
@@ -86,6 +94,10 @@ class _EmbedUrlState extends State<_EmbedUrl> {
       return widget.onSubmit(inputText);
     }
 
+    if (inputText.startsWith('https://youtube')) {
```
LucasXu0, line 97: `^(https?:\/\/)?(www\.)?(youtube\.com|youtu\.be)\/`

use regex would more accurately. for example, here're some edge case
1. youtube
2. www.youtube
3. youtu.be

LucasXu0:
The AFFocusManager is not registered in the test environment, which causes tests to fail. If it's not required in the test environment, I think we can use `maybeOf` instead.

PR #5456, opened 2024-06-03
Update German translations

suntorytimed:
German translation for Recent, AI Chat, default questions, favorites and video feature.


suntorytimed:
Updated to include translation for d5cfd05

PR #5477, opened 2024-06-05
fix: use bitsdojo_window on windows

Xazin:
This is a workaround for an issue where `window_manager` improperly renders the window on Windows.

---

We don't have a listener on Size of Window in bitsdojo, so we still rely on `window_manager`, the only difference is that we use Bitsdojo to render the frameless window instead of window_manager


frontend/appflowy_flutter/lib/startup/tasks/windows.dart
```diff
@@ -43,23 +45,38 @@ class InitAppWindowTask extends LaunchTask with WindowListener {
...
-
-      final position = await windowsManager.getPosition();
-      if (position != null) {
-        await windowManager.setPosition(position);
-      }
-    });
+    final position = await windowSizeManager.getPosition();
+
+    if (PlatformExtension.isWindows) {
+      doWhenWindowReady(() {
```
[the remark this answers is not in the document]
Xazin: Should be by default the way it's built, it renders a frameless window.

frontend/appflowy_flutter/windows/runner/main.cpp
```diff
@@ -5,6 +5,9 @@
 #include "flutter_window.h"
 #include "utils.h"
 
+#include <bitsdojo_window_windows/bitsdojo_window_plugin.h>
+auto bdw = bitsdojo_window_configure(BDW_CUSTOM_FRAME | BDW_HIDE_ON_STARTUP);
```
Xazin, line 9: This is the deciding factor on the frame for the window @LucasXu0

LucasXu0: Nice. I tested it, and it works!

Issue #2904, opened 2023-06-27
[FR] The application remembers being maximized


richardshiue:
### Description

Currently, Appflowy is capable of remembering the window's last size upon startup. However, if it was maximized, the newly-opened application would open in the largest size, but as a floating window. It would be nice if it could remember being maximized, or as a floating window with a smaller size.

### Impact

Users who open AppFlowy as a maximized window.

annieappflowy:
reopen window position is changed and some areas are outside the screen

Issue #3287, opened 2023-08-28
[Bug] app opens by overflowing to the side 


mu-us61:
### Bug Description

i installed flatpak version , when i click desktop shortcut app opens wrongly , I use debian 12 bookworm KDE

### How to Reproduce

1- use debian 12 Bookworm KDE , default is wayland
2- install appflowy by flatpak
3- my screen is 1920 x 1080 laptop

### Expected Behavior

it opens fullscreen

### Operating System

debian 12 bookworm

### AppFlowy Version(s)

0.29

### Screenshots

![image](https://github.com/AppFlowy-IO/AppFlowy/assets/73366454/aa6de04f-63e5-49ee-bf6e-c79dd325a850)

LucasXu0:
Hi, @mu-us61. Do you mean the default size of AppFlowy is too large on Linux?

mu-us61:
hello as in the screen shot it is not in the middle of screen , after every start i have to drag the window to middle then press to fullscreen icon at top right , it would be better if it starts full screen or at least in the middle of screen without overflowing

mu-us61:
in other way of saying the app does not start at position (0,0) so thats why it overflows, it is not about size it is about starting position I guess

SaifEl-Deen:
I can confirm this bug, it happens every time I open the app.

annieappflowy:
@Hexara7777 , do you still encounter this issue?

PR #5622, opened 2024-06-25
fix: maximized window optimization

Xazin:
I've stopped saving the size of the window when it has been maximized, this means that when closing the application on Linux and MacOS from a maximized state, the application will open in it's previously regular state and size.

For Windows, I've added the previous maximized state to persistence, and maximize the window if it was previously maximized on application launch. This seems to be the standard behavior for Windows applications.

Relates: #3287 
Closes: #2904  - PS. This is for windows only, applications on MacOS when closed will turn to regular state, so it cannot open maximized.

#### PR Checklist


annieappflowy:
on macOS and Linux, wouldn't it remember the maximized window?

PR #5645, opened 2024-06-27
fix: score filter

appflowy:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-server/src/af_cloud/impls/search.rs
+++ b/frontend/rust-lib/flowy-server/src/af_cloud/impls/search.rs
@@ -32,7 +32,7 @@ where
     // Filter out irrelevant results
     let result = result
       .into_iter()
-      .filter(|r| r.score < SCORE_LIMIT)
+      .filter(|r| r.score > SCORE_LIMIT)
       .collect();
 
     Ok(result)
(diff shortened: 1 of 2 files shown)


frontend/rust-lib/flowy-server/src/af_cloud/impls/search.rs
```diff
@@ -32,7 +32,7 @@ where
     // Filter out irrelevant results
     let result = result
       .into_iter()
-      .filter(|r| r.score < SCORE_LIMIT)
+      .filter(|r| r.score > SCORE_LIMIT)
```
Horusiath, line 35: Score limit needs adjustment. Right now it's 0.8, which will be very strict after switching the sings. I propose `0.3` as a new default.

Horusiath (changes requested):
It's alright, but I'm afraid, that without changing the threshold may useful results will simply never be shown. 

Here's the link to discussion forum: https://community.openai.com/t/rule-of-thumb-cosine-similarity-thresholds/693670/5

frontend/rust-lib/flowy-server/src/af_cloud/impls/search.rs
appflowy: Understood

Xazin (approved):

Horusiath (approved):

PR #5722, opened 2024-07-11
chore: Update German translations for 0.6.4

suntorytimed:
Describe the changes you have conducted here

Preview the messages on https://inlang.com/github.com/suntorytimed/AppFlowy .


suntorytimed:
@LucasXu0 please note that I translated `web.signInAgreement` including `web.termOfUse`, `web.and` and `web.privacyPolicy` inside of the translated text, as it isn't possible to translate the Text to German with the sentence ending on "[...]AppFlowys [Terms] [and] [Privacy Policy]". So the three other strings will be duplicated and the translation in Germany will look weird, as long as the strings are handled like this in AppFlowy.

LucasXu0:
> And something should happen with this itemThree in the English translation. InLang automatically optimizes the translation and always removes it.

Because itemThree is empty, I should remove it. cc @Xazin 

> please note that I translated web.signInAgreement including web.termOfUse, web.and and web.privacyPolicy inside of the translated text

I want to confirm if you can translate it to German correctly with this format:

```
"signInAgreement": "By clicking "Continue" above, you confirm that
you have read, understood, and agreed to
AppFlowy's {} and {}",
"privacyPolicy": "Privacy Policy",
"termOfUse": "Terms",
```

the {} means the linked text and I'll fill it up with "privacyPolicy" and "termOfUse".

LucasXu0:
@suntorytimed The way I mentioned doesn't work. If using the translation like that, it's unable to set the terms and privacy link.

```dart
          TextSpan(
            text: LocaleKeys.web_signInAgreement.tr(
              args: [
                LocaleKeys.web_termOfUse.tr(),
                LocaleKeys.web_privacyPolicy.tr(),
              ],
            ),
            style: const TextStyle(color: Colors.grey, fontSize: 12),
          ),
```

suntorytimed:
@LucasXu0 that's unfortunate. I will try to formulate the text different in the German translation without moving too far away from the original meaning. Will sound weird for a native speaker, but at least it won't have trailing strings.

suntorytimed:
Not ideal, but this should work. It basically refers to a list of "documents" that need to be read, understood and accepted and "AppFlowy's Privacy Policy and Terms" would be the one and only entry of the "list"

suntorytimed:
Does `Commit messages lint / commitlint (pull_request)` complain about [Fix wrongfully removed itemThree](https://github.com/AppFlowy-IO/AppFlowy/pull/5722/commits/a503be102e051bf5b97a5e419e2ede0389fb23f3) not having a tag like `chore:`?

Xazin:
> > And something should happen with this itemThree in the English translation. InLang automatically optimizes the translation and always removes it.
> 
> Because itemThree is empty, I should remove it. cc @Xazin
>

It's removed in our plan+billing changes, but the PR won't be opened before Wednesday morning earliest.

suntorytimed:
rebased and hopefully fixed the commit message for the CI

PR #5738, opened 2024-07-15
feat: Hebrew translation

yarons:
Hebrew translation.
The UI should be RTL just like Farsi and Arabic.
Thanks.


yarons (commented):
Thank you! I'm not sure why inlang did that.

LucasXu0:
Never mind. I have fixed it.

PR #5844, opened 2024-07-31
fix: try to reopen the first workspace if the workspace deletion failed

LucasXu0:
### Feature Preview

In some cases, the delete workspace API may return an error even though the deletion actually succeeded. Verify by checking the workspace list again to cover this case.

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-core/src/deps_resolve/user_deps.rs
+++ b/frontend/rust-lib/flowy-core/src/deps_resolve/user_deps.rs
@@ -11,6 +11,7 @@ use flowy_user_pub::workspace_service::UserWorkspaceService;
 use lib_infra::async_trait::async_trait;
 use std::collections::HashMap;
 use std::sync::Arc;
+use tracing::info;
 
 pub struct UserDepsResolver();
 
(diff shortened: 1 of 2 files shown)


frontend/rust-lib/flowy-core/src/deps_resolve/user_deps.rs
```diff
@@ -61,9 +62,14 @@ impl UserWorkspaceService for UserWorkspaceServiceImpl {
...
-    self
+    // The remove_indices_for_workspace should not block the deletion of the workspace
+    // Log the error and continue
+    if let Err(err) = self
       .folder_manager
-      .remove_indices_for_workspace(workspace_id)?;
+      .remove_indices_for_workspace(workspace_id)
+    {
+      info!("Error removing indices for workspace: {}", err);
+    }
```
LucasXu0, line 72: @Xazin FYI, the `remove_indices_for_workspace` should not block the deletion of the workspace if it has actually been deleted successfully.

However, the `remove_indices_for_workspace` might sometimes throw an error, possibly with a 1 in 5 chance.

```
{"msg":"[Flutter]: delete workspace error: code: FolderIndexManagerUnavailable\nmsg: FolderIndexManager or its dependencies are unavailable\n","time":"07-30 21:33:30","target":"dart_ffi"}
```

PR #5859, opened 2024-08-01
fix: select option chip size

Xazin:
### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/packages/flowy_infra_ui/lib/src/flowy_overlay/appflowy_popover.dart
```diff
@@ -152,7 +153,7 @@ extension on BuildContext {
         side: BorderSide(
           width: 1,
           strokeAlign: BorderSide.strokeAlignOutside,
-          color: borderColor,
+          color: color != Colors.transparent ? borderColor : color!,
```
Xazin, line 156: @LucasXu0 This is for another issue, I noticed when setting the background of the Popover to transparent, it will still have a border (borderColor). I chose this workaround since I don't see a need to expand on it right now.

PR #6009, opened 2024-08-19
chore: attach rocksdb to newly created databases and database rows

Horusiath:


appflowy:
I revamp the DatabaseCollabService. it return a Persistence service that implement rocksdb action. As result, collab-database doesn't depends on collab-plugin anymore.  I will close this PR.

PR #6015, opened 2024-08-20
chore: non work stealing thread pool

Horusiath:
This PR comes with dedicated non work-stealing thread pool, that is used by appflowy dispatcher. It enables distributing work over multiple threads (by default equal to the number of logical cores on the machine) without having `Send` restriction over executed futures that the tokio default thread pool requires.


appflowy:
@Horusiath I will close this PR and I will reopen it when we start to implment it

PR #6123, opened 2024-08-30
chore: ios build

Xazin:
I've fixed warnings and optimized build space for iOS build to succeed more often.


.github/workflows/ios_ci.yaml
```diff
@@ -45,14 +45,21 @@ jobs:
...
+          model: "iPhone 15"
           shutdown_after_job: false
 
-  build-macos:
+  build-iOS:
     if: github.event.pull_request.head.repo.full_name != github.repository
     runs-on: macos-latest
 
     steps:
+      - name: Maximize build space
```
LucasXu0, line 56: It appears that this cleanup step doesn't work most of the time.

Normally, I execute this command: `gh cache delete --all --repo AppFlowy-IO/AppFlowy`. It has a better effect than this step when there's a 'no space left' error.

.github/workflows/ios_ci.yaml
```diff
@@ -45,14 +45,27 @@ jobs:
...
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf "/usr/local/share/boost"
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+
+      - name: Delete GH Cache
```
LucasXu0, line 63: I didn't explain it clearly before. This action can only be executed locally. It would clear all the existing cache, which may cause the CI build time to be longer.

LucasXu0:
I've used `du` to check the space. The machine has enough space to save the intermediate product. I assume that it's an issue with the latest macOS version or cargo-lipo.

```
💻 💻 💻  Disk space before build
+ echo '💻 💻 💻  Disk space before build'
+ df -h
Filesystem       Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk1s5s1  300Gi  8.7Gi  150Gi     6%  356840 15679125[60](https://github.com/AppFlowy-IO/AppFlowy/actions/runs/10677109542/job/29591704599#step:8:61)    0%   /
devfs           183Ki  183Ki    0Bi   100%     634          0  100%   /dev
/dev/disk1s2    300Gi  1.9Gi  150Gi     2%     903 1567912560    0%   /System/Volumes/Preboot
/dev/disk1s4    300Gi  1.0Mi  150Gi     1%       1 1567912560    0%   /System/Volumes/VM
/dev/disk1s6    300Gi  4[64](https://github.com/AppFlowy-IO/AppFlowy/actions/runs/10677109542/job/29591704599#step:8:65)Ki  150Gi     1%      20 1567912560    0%   /System/Volumes/Update
```

https://github.com/AppFlowy-IO/AppFlowy/pull/6168

PR #6130, opened 2024-08-30
feat: fetch url meta data on creating row

### LINKED ISSUE #5963: [FR] Fetch url meta tags into database fields
### Description

When a URL is pasted to the description field in the database (kanban/grid view) fetch URL metadata to the field for example the HTML title of the URL will get into the description field which is the primary field, if there is a URL field the URL will go there and the meta description will go in the document.
The use case for me is when I want to create kanban cards from URL. I want to use the URL title as the card primary field and also keep the URL as a reference in URL field.

### Impact

Those who want to fill database fields from a URL automatically.

### Additional Context

It can get hooked to paste shortcut so when a URL is pasted it will fetch the URL get meta tags and fill them. To enable or disable it we can make it a setting for database.
I think it's better to implement it on frontend side but I'm not sure.

zoli:
On submitting a new row in board view, if the setting is enabled and the text is a URL fetch url metadata and fill the row with it.

closes #5963

### Feature Preview

https://github.com/user-attachments/assets/26ffc1c1-e009-4079-be28-a1c92c3a1861

#### PR Checklist


zoli:
At first, I wanted to also enable this feature in grid view on filling URL field, but filling other fields inside `EditableURLCell` isn't that straightforward. So I put it aside for maybe in the future.

The next step can be adding the description and image of URL metadata in the row page.

LucasXu0:
After inputting the URL, there's a noticeable delay before the parsed URL is inserted (I understand this is due to fetching metadata). This lag affects the user experience. We should consider a better way to handle this. For example:

1. insert the raw URL text immediately.
2. create a new item with the raw URL text.
3. update the cell content with parsed metadata once it's available.

@zoli What do you think?

Screenshots: 
Screenshot 2024-09-09 at 09 34 59
Screenshot 2024-09-09 at 09 35 04
Screenshot 2024-09-09 at 09 35 10

zoli:
@LucasXu0 By creating a new item do you mean sending `CreateRow` event to the backend? I think creating a pseudo card while fetching the URL is better, Maybe with less opacity so the user knows it's not final yet.

richardshiue:
What should the behavior be when there are more than one URL fields?

zoli:
> What should the behavior be when there are more than one URL fields?

I can change the settings to choose a specific field, for now, I thought selecting the first URL field type would be good enough.

zoli:
@richardshiue will choosing specific field in settings and pseudo card while fetching the URL be enough for this?

appflowy:
Hi @zoli , please add bloc tests or integration test.

zoli:
@appflowy, Added both bloc and integration tests.
@richardshiue, Enabled choosing the field (URL type) to fill the URL.
@LucasXu0,  Added loading state.

LucasXu0:
@richardshiue What do you think about it? I'm thinking that the toggle function should be an option for the URL field type, not in the global settings of the database.

PR #6156, opened 2024-09-01
fix: edit group title + rebuild improvement

Xazin:
- Fix an issue where the end editing event returned before changing the state to not-editing
- Fix a rebuild issue with groups generally, where renaming a group caused the whole board to rebuild.
- Fix a rebuild issue with FlowyHover causing eg. Board Cards to completely rebuild (new lifecycle) meaning the BLoCs would also reinstantiate causing a lot of backend requests on hover.

#### PR Checklist


LucasXu0:
I tried to build this pull request and found that adding a new group refreshes the entire board.

Regarding the issue where the name of a newly created field doesn't refresh in the board, I found something. The group header get the name from the option value before the option value is updated. This is why it sometimes works and sometimes doesn't.

LucasXu0:
This PR broke some text button styles. 

Screenshot 2024-09-06 at 11 28 53

PR #6172, opened 2024-09-03
chore: observe file upload state

appflowy:
### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/startup/tasks/file_storage_task.dart
```diff
@@ -0,0 +1,145 @@
...
+    if (json.containsKey('file_url') && json.containsKey('progress')) {
+      return FileProgress(
+        fileUrl: json['file_url'] as String,
+        progress: (json['progress'] as num).toDouble(),
+        error: json['error'] as String?,
+      );
+    } else {
+      return null;
+    }
+  }
```
LucasXu0, line 113:
```suggestion

  static FileProgress? fromJson(Map<String, dynamic>? json) {
    if (json == null) {
      return null;
    }

    try {
      if (json.containsKey('file_url') && json.containsKey('progress')) {
        return FileProgress(
          fileUrl: json['file_url'] as String,
          progress: (json['progress'] as num).toDouble(),
          error: json['error'] as String?,
        );
      }
    } catch (e) {
      Log.error('unable to parse file progress: $e');
    }
    return null;
  }
```

LucasXu0: Just in case the `file_url` is not a string type or `progress` is not a numeric type.

PR #6182, opened 2024-09-04
fix(flutter): various grid ui issues

richardshiue:
1. fix this monstrosity

    Screenshot 2024-09-04 at 11 48 39 AM

    Screenshot 2024-09-04 at 5 18 26 PM

3. On desktop, make enter on checklist items end editing, only allow shift + enter to go to a new line
4. don't animate the wrap text toggle when opening it
5. add integration test for when changing field type during creation
6. allow opening a related database from the popover

    https://github.com/user-attachments/assets/31e94760-de5d-40a4-a2c0-0db0838e9c40

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database/application/field/field_editor_bloc.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database/application/field/field_editor_bloc.dart
@@ -189,7 +189,7 @@ class FieldEditorEvent with _$FieldEditorEvent {
 @freezed
 class FieldEditorState with _$FieldEditorState {
   const factory FieldEditorState({
-    required final FieldInfo? field,
+    @Default(null) FieldInfo? field,
     @Default(false) bool wasRenameManually,
   }) = _FieldEditorState;
 }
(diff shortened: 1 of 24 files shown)


LucasXu0:
@richardshiue How about adding a 'before' and 'after' comparison?

We have a confirmation dialog style. You may replace the existing one in the database with it.

`showConfirmDeletionDialog`, `showConfirmDialog` and `showCancelAndConfirmDialog`.

frontend/appflowy_flutter/lib/plugins/database/application/field/field_editor_bloc.dart
```diff
@@ -175,7 +184,7 @@ class FieldEditorEvent with _$FieldEditorEvent {
 @freezed
 class FieldEditorState with _$FieldEditorState {
   const factory FieldEditorState({
-    required final FieldInfo field,
+    required final FieldInfo? field,
```
Xazin, line 187:
```suggestion
    @Default(null) FieldInfo? field,
```

Xazin (approved):

annieappflowy:
Number 2: shift + enter -> new line
This is inconsistent with other blocks across mobile and desktop.
Since Mobile is difficult to have the equivalence of shift + enter, we have changed the logic of adding a new line for both desktop and mobile apps.
- Enter -> New line within the same block (quote, callout, etc)
- Shift + Enter / Tap on mobile -> Exit the current block and go to the next line

frontend/appflowy_flutter/lib/plugins/database/application/field/field_editor_bloc.dart
richardshiue: Decided to not feed null into constructor, should I still do this regardless?

Xazin: It's nullable, so we might as well default it I guess.

richardshiue: Sure thing

PR #6212, opened 2024-09-06
fix: newly-created group rows appear out of order

richardshiue:
https://github.com/user-attachments/assets/d7c9a1ba-c7f3-4f4d-984a-90f86f848465

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/src/services/group/controller.rs
+++ b/frontend/rust-lib/flowy-database2/src/services/group/controller.rs
@@ -154,6 +154,61 @@ where
     changeset.deleted_rows.extend(deleted_row_ids);
     Some(changeset)
   }
+
+  fn insert_row_into_group(
+    &mut self,
+    new_row: &Row,
+    index_in_row_order_array: usize,
+    group_id: &str,
+    all_rows: &[Arc<Row>],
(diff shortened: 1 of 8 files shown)


frontend/rust-lib/flowy-database2/src/services/group/controller.rs
```diff
@@ -229,54 +284,71 @@ where
...
-            group.id.clone(),
+    for group_id in group_ids {
+      if self.can_group(&group_id, &cell_data) {
+        // make sure that the row isn't already in the group data
+        if let Some((_, group_data)) = self.get_group(&group_id) {
+          if group_data.contains_row(&row.id) {
+            did_insert_into_status_group = true;
+            continue;
+          }
+        }
```
richardshiue, line 321: @appflowy Something I found. Moving an object in an `ArrayRef` from one position to another also triggers a `Change::Added` delta, meaning that `v_did_create_row` is then called unnecessarily. I've added checks here in the group controllers, but not sure if we should do some handling at the `view_observer` level.

appflowy: yes. I am working on it

PR #6229, opened 2024-09-08
fix(flutter_desktop): more checklist multiline issues

richardshiue:
from 0.6.8 and most recent review, due to introduction of multi-line checklist tasks

- cannot create multi-line tasks from the "add new task" text field
- cannot quickly create checklist tasks in the row detail page

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/database/widgets/cell_editor/checklist_cell_editor.dart
+++ b/frontend/appflowy_flutter/lib/plugins/database/widgets/cell_editor/checklist_cell_editor.dart
@@ -138,6 +138,10 @@ class _EndEditingTaskIntent extends Intent {
   const _EndEditingTaskIntent();
 }
 
+class _UpdateTaskDescriptionIntent extends Intent {
+  const _UpdateTaskDescriptionIntent();
+}
+
 /// Represents an existing task
 @visibleForTesting
 class ChecklistItem extends StatefulWidget {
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/lib/plugins/database/widgets/cell_editor/checklist_cell_editor.dart
```diff
@@ -339,29 +351,28 @@ class _NewTaskItemState extends State<NewTaskItem> {
...
+                  isCollapsed: true,
+                  contentPadding: const EdgeInsets.symmetric(
+                    vertical: 6.0,
+                    horizontal: 2.0,
+                  ),
+                  hintText: LocaleKeys.grid_checklist_addNew.tr(),
                 ),
-                hintText: LocaleKeys.grid_checklist_addNew.tr(),
+                onSubmitted: (_) => _createNewTask(context),
+                onChanged: (value) => setState(() {}),
```
LucasXu0, line 374: Can you add a comment here explaining why we need to `setState` when the text is changing?

richardshiue: Of course

LucasXu0: Actually, I think creating a new field to keep the enable status would be better than using an empty callback.

For example: `setState(() => isCreateButtonEnabled = !textEditingController.text.isEmpty)`

LucasXu0 (approved):

Issue #6221, opened 2024-09-06
[Bug] sort conditions make filters not working anymore


annieappflowy:
### Bug Description

<img width="477" alt="image" src="https://github.com/user-attachments/assets/537316cf-eba5-48c1-8bee-195e243a65a9">
<img width="658" alt="image" src="https://github.com/user-attachments/assets/6ae11cbf-ff2a-49e5-89e2-b3b4ef311d15">

### How to Reproduce

Follow my screenshots

### Expected Behavior

Should filter out data that doesn't meet filter conditions

### Operating System

macOS

### AppFlowy Version(s)

v0.6.9

PR #6241, opened 2024-09-09
chore: fix database rows filter and sort

appflowy:
Workaround to fix filter and sort don't work when open database. Currently rendering all rows  when database rows are loaded. In the future, all rows should be lazy load.


appflowy:
fix #6221

PR #6275, opened 2024-09-11
chore: update zh-CN translations

Chikage0o0:
After in Chinese is translated as `以后`, before in Chinese is translated as `之前`, and there is a mistake in the translation here.

Preview the messages on https://inlang.com/github.com/Chikage0o0/AppFlowy .


LucasXu0:
@Chikage0o0 Is the ja-JP translation auto formatted by inlang?

Chikage0o0:
> @Chikage0o0 Is the ja-JP translation auto formatted by inlang?
Yes, should I revert the ja-JP changes？

LucasXu0:
> @LucasXu0 Yes, should I revert the ja-JP changes？

Yes. Please revert the ja-JP changes. :)

PR #6288, opened 2024-09-12
feat: date filter

zoli:
### Feature Preview

---

#### PR Checklist


richardshiue (commented):
Hi @zoli, thanks for your contribution as always!

1. `DateFilterCondition::Is`, the default filter condition, refers to today. So it will immediately hide all the rows that are not on today. That logic is fine, but the filter popup seems to say that the cells must be at that exact time. Maybe we should delete the time field? This also applies to `before`, `after`, `on or before` and `on or after`

    Screenshot 2024-09-17 at 2 46 23 PM

2. `is between` needs a start and end datetime in order to function. Here, users can't do that

   Screenshot 2024-09-17 at 2 50 39 PM

zoli:
@richardshiue about `is`, `before`, `after`, ... you are right I will change it so it applies without a specific time. about the `is between` user can set start and end because the date field in this case enables choosing range. So on other conditions its only single date and time but on `is between` the datepicker allows selecting a range of date and time.

richardshiue:
Oh that's my bad. I only saw one text field and so I got super confused. Splitting the start and end datetime values into two text fields while still being edited by a single date picker could be a future improvement.

zoli:
@richardshiue done.

PR #6324, opened 2024-09-16
fix: dismiss settings dialog after signing in

LucasXu0:
### Feature Preview

Cancel `_deepLinkSubscription` before disposing auth service to ensure new `completer` triggers.

---

#### PR Checklist


Xazin (approved):
Good catch - Something is wrong with tests though

LucasXu0:
@Xazin I replaced the stream with a broadcast stream, so we don't need to use static values to keep the subscription.

Issue #6341, opened 2024-09-18
[FR] When create a new page the focus must be on page edit input


henri9813:
### Description

When I create a page, I want to have my focus automatically to change the page name.

### Impact

I will be more efficient in creating new pages and avoid to forget changing page name.

### Additional Context

Before: 
<img width="380" alt="image" src="https://github.com/user-attachments/assets/8429c197-a488-493f-8bdd-6bb0536c1140">

After ( wanted )

<img width="551" alt="image" src="https://github.com/user-attachments/assets/18e5d249-689e-4917-977e-9849152d6baf">

emmggi:
There used to be a preference to prompt a dialog for the title. I don't know why it was removed.

Xazin:
Once we have completed #3191 we will most likely auto-focus the title for easy editing on creation, whether that will solve this, I'm not entirely sure.

LucasXu0:
I'm currently working on the document title feature. Once completed, we won't need to show a dialog or popup for editing the title anymore. Users will be able to edit the title directly in the document.

preview 
<img width="1491" alt="Screenshot 2024-09-18 at 20 12 08" src="https://github.com/user-attachments/assets/11eac65b-fda4-4e94-b0b5-b75720f4598e">

PR #6353, opened 2024-09-19
feat: support cover title

LucasXu0:
### Feature Preview

closes https://github.com/AppFlowy-IO/AppFlowy/issues/6341

- [x] Remove the default name for view
- [x] Display the default name if the title is empty.
- [x] Add cover title
- [x] Support arrow keys to navigate between title and content
  - [x] arrow down and arrow right in title
  - [x] arrow up and arrow left in doc
- [x] Press enter key in title to jump to the first line of content (split the title text if the cursor is not at the end)
- [x] Replace the cover title selection color
- [x] Sync the name between cover title and view name
- [x] Press backspace key at the first line to go to the end of the title
- [ ] ~~Select all~~ (not supported yet)
- [x] Breaking change: The name of newly created pages is null or empty in this version if the user doesn't rename it. Users of the old version will see an empty name.
- [x] The backspace key doesn't work after pressing Enter in the title
- [ ] ~~Bring the paragraph to title when pressing backspace.~~ (blocking by date/reminder/linked page)
- [x] Only change the name for document, not including database

##### preview
https://github.com/user-attachments/assets/ebbe64bc-f778-470e-8796-7204c072414f

---

#### PR Checklist


frontend/appflowy_flutter/integration_test/desktop/document/document_title_test.dart
```diff
@@ -0,0 +1,43 @@
...
+import 'package:integration_test/integration_test.dart';
+
+import '../../shared/util.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('document title', () {
+    testWidgets(
+        '''create a new document, the title will be displayed at the first line''',
```
Xazin, line 12:
```suggestion
        'create a new document, the title will be displayed at the first line',
```

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/base/cover_title_command.dart
```diff
@@ -0,0 +1,102 @@
...
+final CommandShortcutEvent backspaceToTitle = CommandShortcutEvent(
+  key: 'backspace to title',
+  command: 'backspace',
+  getDescription: () => 'backspace to title',
+  handler: (editorState) => _backspaceToTitle(
+    editorState: editorState,
+  ),
+);
+
+KeyEventResult _backspaceToTitle({
```
Xazin, line 20: Shouldn't it also bring the paragraph up to the title?

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/header/cover_title.dart
```diff
@@ -0,0 +1,225 @@
...
+import 'package:appflowy/workspace/application/appearance_defaults.dart';
+import 'package:appflowy/workspace/application/view/view_bloc.dart';
+import 'package:appflowy_backend/protobuf/flowy-folder/view.pb.dart';
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:easy_localization/easy_localization.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+FocusNode? coverTitleFocusNode;
```
Xazin, line 12: I'm not a huge fan of this dangling global variable. Can we use an inherited widget or a Provider?

LucasXu0: Let me see if there is a way to share this value. Using a provider or inherited widget doesn't work here unless we register this value at the top of the widget tree.

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/base/cover_title_command.dart
LucasXu0: Yes. But I'm struggling with the reminder/date/linked page issue. These texts were formatted with $, it makes no sense to convert them back to $ when bringing the paragraph to title. :(

PR #6356, opened 2024-09-19
refactor: filter blocs

richardshiue:
Also fixes a bug where changing a field's properties such as its name and the available options for select field types doesn't change those properties on the filter chips.

### Feature Preview

---

#### PR Checklist


appflowy:
It would be better if we have filter bloc test. After filter bloc, we can have other bloc tests as well.

frontend/appflowy_flutter/lib/plugins/database/grid/application/filter/select_option_filter_bloc.dart
```diff
@@ -1,146 +0,0 @@
...
-            emit(state.copyWith(filterDesc: desc));
-          },
-        );
-      },
-    );
-  }
-
-  void _startListening() {
-    _listener.start(
-      onUpdated: (filter) {
```
appflowy, line 85: without this listener, how to refresh the UI

richardshiue: Filter editor bloc listens to entire set of filter changes. For any database view, I think the number of filters will never be too high and I use BlocSelectors to reduce widget rebuilds

PR #6365, opened 2024-09-20
fix: 0.7.0 launch review issues

LucasXu0:
### Feature Preview

- [x] cannot use 'move to' from sidebar
- [x] show a forbiend cusor when moving a page to database
- [x] check if the move to path is same as current path
- [x] document width doesn't refresh after resetting
- [x] disable tooltip when dragging blocks
- [x] reminder not render when dragging
- [x] after changing the language in the settings page, some of the text doesn't change
- [x] support moving a block to another block's child
- [x] fix assertion when dragging table
- [ ] tests
  - [x] cannot use 'move to' from sidebar
  - [x] show a forbiend cusor when moving a page to database
  - [x] check if the move to path is same as current path
  - [x] document width doesn't refresh after resetting
  - [x] disable tooltip when dragging blocks
  - [x] reminder not render when dragging
  - [ ] after changing the language in the settings page, some of the text doesn't change
  - [x] support moving a block to another block's child
  - [x] fix assertion when dragging table

---

#### PR Checklist


appflowy:
> ### Feature Preview
> * [x]  cannot use 'move to' from sidebar
> * [x]  show a forbiend cusor when moving a page to database
> * [x]  check if the move to path is same as current path
> * [x]  document width doesn't refresh after resetting
> * [x]  disable tooltip when dragging blocks
> * [x]  reminder not render when dragging
> * [x]  after changing the language in the settings page, some of the text doesn't change
> * [x]  support moving a block to another block's child
> * [x]  fix assertion when dragging table
> * [ ]  tests
> 

Since you had fixed lots of bugs, can you make some tests for them?

appflowy:
LGTM

PR #6376, opened 2024-09-21
fix: semi-transparent colors being forced to full opacity in FlowySvg

richardshiue:
As an example, hintColor on mobile is not a fully-opaque gray color, but black with its opacity toned down. When passed to FlowySvg, this color's opacity is cranked back up to 1.0. In the sort panel, both the "no active sorts" text and the sort icon specify this hint color but look completely different. In the second example, notice the plus sign next to the words "new row".

Before:
Screenshot 2024-09-21 at 10 07 41 PM

Screenshot 2024-09-21 at 10 08 20 PM

After:
Screenshot 2024-09-21 at 10 07 01 PM

Screenshot 2024-09-21 at 10 08 53 PM

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
+++ b/frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
@@ -79,8 +79,8 @@ class FlowySvg extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     Color? iconColor = color ?? Theme.of(context).iconTheme.color;
-    if (opacity != null && iconColor != null) {
-      iconColor = iconColor.withOpacity(opacity!);
+    if (opacity != null) {
+      iconColor = iconColor?.withOpacity(opacity!);
     }
 
     final textScaleFactor = MediaQuery.textScalerOf(context).scale(1);
--- a/frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
+++ b/frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
@@ -73,7 +73,7 @@ class FlowySvg extends StatelessWidget {
 
   /// The opacity of the svg
   ///
-  /// The default value is 1.0
+  /// if null then use the opacity of the iconColor
   final double? opacity;
 
   @override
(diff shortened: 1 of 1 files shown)


frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
```diff
@@ -23,7 +23,7 @@ class FlowySvg extends StatelessWidget {
     this.size,
     this.color,
     this.blendMode = BlendMode.srcIn,
-    this.opacity = 1.0,
```
appflowy, line 26: I wonder there might some svg widgets rely on `opacity=1.0.` default value. So it might cause UI diplay issue

richardshiue: Hmm... AFAIK, color values in our themes are copied out of the figma file without any modification, so purposefully changing the opacity should be situational like applying a visual effect according to an user interaction.

We should remove defaulting to 1.0 it because semi-transparent color copied from our mockups doesn't look the same in app. The opposite holds true; if we notice that there is an icon that looks off in app, then it should also look off in the mockups and we should improve the design.

frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
```diff
@@ -74,12 +74,15 @@ class FlowySvg extends StatelessWidget {
...
   /// The default value is 1.0
-  final double opacity;
+  final double? opacity;
 
   @override
   Widget build(BuildContext context) {
-    final iconColor =
-        (color ?? Theme.of(context).iconTheme.color)?.withOpacity(opacity);
+    Color? iconColor = color ?? Theme.of(context).iconTheme.color;
+    if (opacity != null && iconColor != null) {
```
Xazin, line 82: Why not just `if (opacity != null) {`?

Xazin (approved):

frontend/appflowy_flutter/packages/flowy_svg/lib/src/flowy_svg.dart
richardshiue: iconColor is still nullable here so we still have to check for null. I guess it's a bit cleaner?

```
  Color? iconColor = color ?? Theme.of(context).iconTheme.color;
  if (opacity != null) {
    iconColor = iconColor?.withOpacity(opacity!);
  }
```

LucasXu0 (approved):

PR #6381, opened 2024-09-23
fix: row cover improvements

Xazin:
Requires: https://github.com/AppFlowy-IO/AppFlowy-Collab/pull/294

- [x] Adjust size of the file upload menu in media cell
- [x] Fix issue where sometimes the upload menu for media wouldn't open
- [x] Use same style for Cover in Document on Row Detail Page
- [x] Show the row cover when opened in full page
- [x] Add ability to change cover after it is added, and adjust behavior so it's the same as the document cover. (Adding a cover will make it into the default asset cover)
- [x] Adjusted tests to the new changes
- [x] Set image from media cell as cover
- [x] Duplicate row meta when duplicating a row (eg. also bring cover, and other meta data in new row)

### Feature Preview

TBD

#### PR Checklist


LucasXu0:
@Xazin When can I merge this PR?

Xazin:
> @Xazin When can I merge this PR?

We need to merge https://github.com/AppFlowy-IO/AppFlowy-Collab/pull/294 and update collab revision. Then we can run the tests and double check everything is working as it should.

PR #6396, opened 2024-09-24
chore: remove supabase dependency

Xazin:
Supabase has been sunsetted and is not being used in the client.

I've removed the code related to supabase (mostly from the Flutter side and in language files). - _If we plan to add it back, it remains mostly untouched in the backend as of yet, although there is a lot of unused strategies and edge-cases that litter the authentication logic._

For backwards compatibility, I didn't remove the `EncryptSecretScreen` and surrounding logic.

#### PR Checklist


LucasXu0:
There are some pub deps that can't be updated because of constraints set by the `supabase_flutter` version. Now, we can remove those dependency overrides in `dependency_overrides`.

Xazin:
> There are some pub deps that can't be updated because of constraints set by the `supabase_flutter` version. Now, we can remove those dependency overrides in `dependency_overrides`.
> 
>

I wasn't aware 👍 I will take a look.

Issue #6405, opened 2024-09-26
[Bug] 'edit grid cell: date time' test failed randomly


LucasXu0:
### Bug Description

https://github.com/AppFlowy-IO/AppFlowy/actions/runs/11035472076/job/30652197480

### Operating System

Linux (Github Action)

### AppFlowy Version(s)

0.7.0

PR #6406, opened 2024-09-26
fix: grid cell date time test

LucasXu0:
### Feature Preview

closes https://github.com/AppFlowy-IO/AppFlowy/issues/6405

The hours and minutes are from the previously set up time, not from the current time. So there may be a potential issue when using `datetime.now` to check the result.

Added a 50-time test to reproduce this issue. Delete the test if all of them are successful.

---

#### PR Checklist


richardshiue (approved):
Makes sense, LGTM

PR #6432, opened 2024-09-30
feat(flutter_desktop): use regular rather than medium font weight by default

richardshiue:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/workspace/application/settings/appearance/base_appearance.dart
+++ b/frontend/appflowy_flutter/lib/workspace/application/settings/appearance/base_appearance.dart
@@ -2,7 +2,6 @@ import 'package:appflowy/shared/google_fonts_extension.dart';
 import 'package:flowy_infra/size.dart';
 import 'package:flowy_infra/theme.dart';
 import 'package:flutter/material.dart';
-import 'package:universal_platform/universal_platform.dart';
 
 // the default font family is empty, so we can use the default font family of the platform
 // the system will choose the default font family of the platform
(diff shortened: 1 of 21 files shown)


frontend/appflowy_flutter/lib/workspace/application/settings/appearance/base_appearance.dart
```diff
@@ -31,8 +30,7 @@ abstract class BaseAppearance {
     double? lineHeight,
   }) {
     fontSize = fontSize ?? FontSizes.s14;
-    fontWeight = fontWeight ??
```
LucasXu0, line 34: Is it a request from the designer?

richardshiue: Not explicitly, no. But ever since moving from Poppins, the mockups have been mostly regular, and use medium to symbolize importance

Xazin: Makes sense right, the normal body fontweight is regular.

But I'm wondering how this affects text that already go by the original case, do we need to double check everywhere?

richardshiue: Yeah... that's the thing. I'm checking as much as I can

LucasXu0: There are some texts in buttons that may be affected. I remember that most text font weights in buttons should be 500.

richardshiue: Done! It defaults to w500 nicely now. See the latest commit

PR #6483, opened 2024-10-05
fix: import data

appflowy:
1. Fix the issue with importing the AppFlowy data folder. After importing the data folder, ensure that all imported views are placed in the current workspace.
2. Fix the migration process for anonymous users. All documents and databases from anonymous users should be moved to a newly created workspace.
3. fix import appflowy data folder which contains spaces


appflowy:
https://github.com/user-attachments/assets/39c61434-b64f-4c7b-801d-239be0c20906

PR #6506, opened 2024-10-08
fix: don't process on every view while only handling one view

richardshiue:
The database row observe should only edit one view when handling one databaseviewchange. Otherwise, we have a bug like the following

https://github.com/user-attachments/assets/4185313c-182f-49d2-9d4b-c1c5a18723d7

### Feature Preview

---

#### PR Checklist


LucasXu0:
Is it introduced in v0.7.1 or the previous version? I suggest we have an integration test (not only unit tests or bloc tests) to cover it.

richardshiue:
> Is it introduced in v0.7.1 or the previous version? I suggest we have an integration test (not only unit tests or bloc tests) to cover it.

This was 0.6.9

PR #6595, opened 2024-10-21
fix: enable sub page block

Xazin:
### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/packages/appflowy_popover/lib/src/layout.dart
```diff
@@ -11,16 +11,12 @@ class PopoverLayoutDelegate extends SingleChildLayoutDelegate {
     required this.direction,
     required this.offset,
     required this.windowPadding,
-    this.showAtCursor = false,
```
LucasXu0, line 14: Is this a wrong commit? This deletion will break the right-click context menu functionality.

Xazin: It's not s wrong commit, I made the ability to choose to show an existing popover programmatically at a specific offset.

I am just testing this approach to make sure all tests work properly. This is to avoid having multiple of the same AppFlowyPopover when what we want is to show the same popover just at another offset.

For the case where only showAtCursor for the popover is needed, I will add it back in this PR as well.

Issue #6608, opened 2024-10-23
[FR] Delete a property to adopt the same design for the delete button


annieappflowy:
### Description

Current: 
<img width="421" alt="image" src="https://github.com/user-attachments/assets/f83f5700-d118-4e17-b032-ac87f354f405">

Consistent design:
<img width="336" alt="image" src="https://github.com/user-attachments/assets/2595fefc-bbad-4f9d-93e2-024f52aa1af6">

### Impact

AppFlowy users

PR #6620, opened 2024-10-24
feat: add same delete design in database

Ahad-patel:
closes : #6608

### Feature Preview

I have updated FlowyButton with FlowyIconTextButton. I have also added resetHoverOnRebuild in FlowyIconTextButton. Since leftIconBuilder and rightIconBuilder require non-nullable widget, I have updated action.leading and action.trailing to return Sizedbox.shrink() instead of null and made return type Widget in pace of Widget?. 

One question though, is any where else this delete design needed to be added? and does it require test cases? @richardshiue  @annieappflowy 

---

https://github.com/user-attachments/assets/05e74709-52a3-4fce-be49-774a56bd5699

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/database/widgets/field/field_editor.dart
```diff
@@ -229,7 +237,7 @@ enum FieldAction {
   delete,
   wrap;
 
-  Widget? leading(FieldInfo fieldInfo, Color? color) {
+  Widget leading(FieldInfo fieldInfo, Color? color) {
```
richardshiue, line 240: Please revert to returning `Widget?`. This adds unexpected, extra padding to the end of the row in the button. Otherwise, looks good to me.

richardshiue (changes requested):

richardshiue:
No test case needed

frontend/appflowy_flutter/lib/plugins/database/widgets/field/field_editor.dart
Ahad-patel: @richardshiue I don't think it will add padding, I have attached video and screenshot of the same, are you referring to this file https://github.com/Ahad-patel/AppFlowy/blob/e66835297a2dc84febf8d4f1517aa64252fc4624/frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/view_more_action_button.dart#L257, as explicit padding is added in it.

Or I am referring at wrong place, I don't know 🤷 ?.

https://github.com/user-attachments/assets/b13ac2df-9261-4f1a-aa8f-33c80b76a1dd


Screenshot 2024-10-25 at 2 53 44 PM
Screenshot 2024-10-25 at 2 54 09 PM

richardshiue: Yes that's it. Do you see that whereas edit property there is no extra padding at the end, the others have it? For languages where the string becomes too long, or we have another action that has a longer name, the right end will not line up

Ahad-patel: I got your point. But since in FlowyIconTextButton , the leftIconBuilder and rightIconBuilder have a return type of Widget, so I have updated to Widget? and added a conditon that will check if resutant widget is nullable or not   `final Widget? leftIcon = leftIconBuilder?.call(onHover);` similar for the rightIconBuilder. So the extra padding will be avoided. Please let me know if any more changes needed for this PR. 

Screenshot 2024-10-25 at 6 59 56 PM

LucasXu0:
@Ahad-patel Can you merge the latest main branch to re-trigger the CI actions?

richardshiue (approved):

PR #6661, opened 2024-10-29
fix(flutter_desktop): glitchy kanban card editing

richardshiue:
taking the most stripped-down steps to reproduce:

1. user types in 'a'
2. onChanged callback starts a 300ms debounce timer to update the cell in the backend with 'a'
3. immediately after 300ms and before the backend notifies the frontend of the cell change, user types in 'b'. The text controller's value is now 'ab'
4. cell controller gets the cell-did-update notification from backend, data is 'a'
5. cell controller updates the text controller's value with 'a'
6. The user observes that b disppears

I had previously checked whether the cell has focus in step 5, which fixed this issue. The problem is that if another user edits the same cell, those changes don't get reflected.

Solution now is don't listen to onChanged callbacks, only when the user finishes ediiting it.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/integration_test/desktop/board/board_row_test.dart
+++ b/frontend/appflowy_flutter/integration_test/desktop/board/board_row_test.dart
@@ -6,6 +6,7 @@ import 'package:easy_localization/easy_localization.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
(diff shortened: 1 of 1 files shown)


LucasXu0:
I haven't tested it yet, but I have a guess about the changes: if I click the empty area to cancel text editing, will the onFinish callback be called?

richardshiue:
Yes, there is a onfocuschanged listener doing this!

richardshiue:
Or so I thought

PR #6747, opened 2024-11-08
feat: include member count in get_workspace API

speed2exe:


frontend/rust-lib/flowy-sqlite/src/schema.rs
```diff
@@ -101,6 +101,7 @@ diesel::table! {
         created_at -> BigInt,
         database_storage_id -> Text,
         icon -> Text,
+        member_count -> BigInt,
```
LucasXu0, line 104: Do we need to handle up.sql and down.sql manually?

speed2exe: yeah, I guess you are right. I missed that

speed2exe: added the `up.sql` and `down.sql` to `frontend/rust-lib/flowy-sqlite/migrations`

Issue #4153, opened 2023-12-16
[FR] i18n for duplicated field names


richardshiue:
### Description

When duplicating a field, the new field will have the old name + the suffix "(copy)". For example, duplicating a field named `Due Date` will create a new field called `Due Date (copy)`. It would be better if copy were translated according to the user's current language.

Requires both Flutter and Rust

### Impact

Non-english AppFlowy users

Akila-Kavisinghe:
Hey Richard I can pick this up as a good first task

annieappflowy:
Thanks, @Akila-Kavisinghe . Assigned.

PR #6769, opened 2024-11-12
feat: i18n for duplicated field names

Ahad-patel:
closes: #4153 

### Feature Preview

We can pass the copy from suffix param, this will not require to add i18n string for copy in from the rust side. I have added a new key, as per the convention used. Should I use an existing key? as there will be no translations for this. @richardshiue 

---

#### PR Checklist


frontend/resources/translations/en.json
```diff
@@ -253,7 +253,8 @@
     "moreButtonToolTip": "Remove, rename, and more...",
     "addPageTooltip": "Quickly add a page inside",
     "defaultNewPageName": "Untitled",
-    "renameDialog": "Rename"
+    "renameDialog": "Rename",
+    "copy": "Copy"
```
richardshiue, line 257: Please give a more descriptive name to this translation token. Simply "copy" is misleading

richardshiue (commented):
Can you add a test for this change?

LucasXu0 (approved):

PR #6783, opened 2024-11-14
feat: use new api to get the workspace member role

speed2exe:


LucasXu0:
@speed2exe The Rust CI failed. After checking the integration test error, I found that it was unable to create a new page in local mode.

frontend/rust-lib/flowy-ai/src/local_ai/watch.rs
```diff
@@ -5,13 +5,15 @@ use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
 use tracing::{error, trace};
 
 #[cfg(any(target_os = "windows", target_os = "macos", target_os = "linux"))]
+#[allow(dead_code)]
 pub struct WatchContext {
   #[allow(dead_code)]
```
LucasXu0, line 10: Do we still need to keep `#[allow(dead_code)]` for the `watcher`?

speed2exe: we dont need this anymore, I'll remove it

LucasXu0:
Fetch the member role from the workspace list instead of sending a request for each workspace to reduce API calls and UI loading time.

PR #6793, opened 2024-11-14
chore(mobile): remove autoscrolling sample questions on mobile

richardshiue:
When I call `scrollController.animateTo`, the first user drag or click will always interrupt that in-progress scroll, and the hit test doesn't get to the actual button that's inside the custom scroll view. When a second tap occurs, since the scrolling has stopped, the button inside can now receive the pointer event.

Since this may cause bad UX, we are completely removing the autoscrolling behavior on mobile.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/chat_welcome_page.dart
+++ b/frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/chat_welcome_page.dart
@@ -1,14 +1,11 @@
 import 'package:appflowy/generated/flowy_svgs.g.dart';
 import 'package:appflowy/generated/locale_keys.g.dart';
-import 'package:appflowy/util/debounce.dart';
 import 'package:appflowy/util/theme_extension.dart';
 import 'package:appflowy_backend/protobuf/flowy-user/user_profile.pb.dart';
 import 'package:easy_localization/easy_localization.dart';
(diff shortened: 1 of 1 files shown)


Xazin:
I'm on the plane so I can't read documentation (internet slower than _the_ snail) or play with it right now.

I'm unsure of the ideal implementation, but here are my thoughts:

If you want to stop the user from breaking the scroll, you can set the physics to NeverScrollablePhysics() whilst processing a message from AI. You mention that pressing anywhere might cause the animateTo to stop as well, I can't recall the behavior of this on mobile on top of my head, but you might be able to use a Tween and trade animateTo for jumpTo. That should completely stop user from interacting with the chat area whilst AI is writing its response.

I haven't used AI on phone a lot, but AI spews more lines than I can read, and my natural response is to try to stay at the top of the message to actually read the message, rather than wait for it to finish. So I would rather break the scroll and scroll myself, there are some cases where I prefer the auto scroll, but mostly end up breaking it before it's done, 90% of the time. Just my two cents, as I'm unsure what the ideal solution was in the first place.

<details>

Snail to be added.

</details>

Xazin:
> Oh I'm talking about this
> 

That was something entirely different than I expected 😂 I will take a look when I land. I have some ideas that might be able to work.

richardshiue:
Merging for now, we can revisit this later.

Issue #6816, opened 2024-11-17
[Bug] Broke AppFlowy Boards by changing Status to Date


Oshirowanen:
### Bug Description

I somehow was able to chose "date" in a board, and now I can no longer get into the board.  

### How to Reproduce

Open any board
click Settings > Properties > Status
Select Date
Click out of the board and then try to get back in.

### Expected Behaviour

If this is going to break the board, I shouldn't be able to make such a change.
I should at least be able to check the status back to a default value to bring the board contents back.

### Operating System

Ubuntu 24.04
Windows 11

### AppFlowy Version(s)

0.7.3

### Video recording

https://github.com/user-attachments/assets/440ffa98-3aa3-4834-a7f7-aa8ad32725d5

### Additional Context

1. appflowy snap package from the snap repository
2. How do I fix this without reinstalling?

Oshirowanen:
Hi @richardshiue 

Sorry about the badly worded post, it was done in a rush.  I have now updated the post and have included a video to show the steps more clearly.

I hope the updated post clarifies the issue.  I can confirm that I get the same issue on Windows 11 and Ubuntu 24.04.

richardshiue:
Hi @Oshirowanen, thanks for the update. Yes I can reproduce it now, I will see to fixing this issue as soon as possible.

PR #6860, opened 2024-11-26
fix: RwLock race condition during group controller instantiation

richardshiue:
resolves https://github.com/AppFlowy-IO/AppFlowy/issues/6816

~~This fix introduces a regression to the calculation feature. Therefore, another solution needs to be found~~

Sometimes a bug fix is dirty. But sometimes a dirty bug fix works.

The test doesn't pass on main

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/rust-lib/flowy-database2/src/services/database_view/views.rs
+++ b/frontend/rust-lib/flowy-database2/src/services/database_view/views.rs
@@ -71,9 +71,11 @@ impl DatabaseViews {
       )
       .await?,
     );
-    let mut editor_map = self.view_editors.write().await;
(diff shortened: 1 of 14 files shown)


frontend/rust-lib/flowy-database2/src/services/database_view/views.rs
```diff
@@ -72,7 +71,10 @@ impl DatabaseViews {
       )
       .await?,
     );
+    let mut editor_map = self.view_editors.write().await;
     editor_map.insert(view_id.to_owned(), editor.clone());
```
appflowy, line 75: can use `self.view_editors.write().await`.insert then you don't need to drop

PR #6917, opened 2024-12-03
feat: open ai response url source in browser

richardshiue:
Screenshot 2024-12-03 at 10 34 07 PM

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/ai_chat/chat_page.dart
+++ b/frontend/appflowy_flutter/lib/plugins/ai_chat/chat_page.dart
@@ -1,9 +1,12 @@
+import 'dart:io';
+
 import 'package:appflowy/generated/locale_keys.g.dart';
 import 'package:appflowy/plugins/ai_chat/application/chat_bloc.dart';
 import 'package:appflowy/plugins/ai_chat/application/chat_entity.dart';
 import 'package:appflowy/plugins/ai_chat/application/ai_prompt_input_bloc.dart';
 import 'package:appflowy/plugins/ai_chat/application/chat_message_stream.dart';
 import 'package:appflowy/plugins/ai_chat/presentation/chat_related_question.dart';
+import 'package:appflowy_backend/log.dart';
 import 'package:appflowy_backend/protobuf/flowy-folder/view.pb.dart';
 import 'package:appflowy_backend/protobuf/flowy-user/protobuf.dart';
 import 'package:desktop_drop/desktop_drop.dart';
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/lib/plugins/ai_chat/chat_page.dart
```diff
@@ -273,10 +278,26 @@ class _ChatContentPage extends StatelessWidget {
...
+          onSelectedMetadata: (metadata) async {
+            if (isURL(metadata.name)) {
+              late Uri uri;
+
+              try {
+                uri = Uri.parse(metadata.name);
+                // `Uri` identifies `localhost` as a scheme
+                if (!uri.hasScheme || uri.scheme == 'localhost') {
+                  uri = Uri.parse("http://${metadata.name}");
+                  await InternetAddress.lookup(uri.host);
```
LucasXu0, line 290: If the address is not found, will this function throw an error?

By the way, I saw this openlink logic in other files. We may move it to an extension and utility file to share the same logic.

frontend/appflowy_flutter/lib/plugins/ai_chat/chat_page.dart
```diff
@@ -273,10 +278,26 @@ class _ChatContentPage extends StatelessWidget {
           chatId: view.id,
           refSourceJsonString: refSourceJsonString,
           isLastMessage: isLastMessage,
-          onSelectedMetadata: (metadata) {
-            context
-                .read<ChatSidePanelBloc>()
-                .add(ChatSidePanelEvent.selectedMetadata(metadata));
+          onSelectedMetadata: (metadata) async {
+            if (isURL(metadata.name)) {
```
LucasXu0, line 282: What do you think about putting the open link logic in the bloc?

frontend/appflowy_flutter/lib/plugins/ai_chat/chat_page.dart
richardshiue: > If the address is not found, will this function throw an error?

No, the exception will be caught and do nothing.

> By the way, I saw this open link logic in other files. We may move it to an extension and utility file to share the same logic.

This one is copied from the database version, but without searching google if it cannot be launched. I don't use `afLaunchUrl` doesn't work for all the cases I need it to. We can unify later if you wish.

frontend/appflowy_flutter/lib/plugins/ai_chat/chat_page.dart
richardshiue: This code will be changed in a following PR because the chat side panel bloc will be gone.

PR #6976, opened 2024-12-12
fix(flutter_desktop): cannot copy ai response with ctrl c

richardshiue:
### Feature Preview

---

#### PR Checklist


richardshiue (commented):
@LucasXu0 if I enable keyboard service on desktop, will there be regressions? This is supposed to be uneditable.

LucasXu0:
@richardshiue I think it would be better to just enable the custom copy command. Otherwise, the arrow keys will still work in this case.

Issue #6334, opened 2024-09-17
[FR] Drag and Drop linking to pages


lynrayy:
### Description

It would be nice to implement page linking via Drag and Drop. At the moment, the implementation is only via "+"

### Impact

Creating page links and table of contents will become intuitive. For example, I didn't know this feature even existed (via the "+" symbol) until I started searching for the feature on GitHub.

### Additional Context

Related
https://github.com/AppFlowy-IO/AppFlowy/issues/6301

Screenshot
![image](https://github.com/user-attachments/assets/81e4cd23-df6a-4326-9ac8-4a31b16cd5d8)

annieappflowy:
Supported

PR #6991, opened 2024-12-15
feat: drag view into document to link to view

Xazin:
Relates: #6334 

### Feature Preview

https://github.com/user-attachments/assets/9808d244-7c20-41f9-b10b-bdeba9c395ee

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_drop_handler.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_drop_handler.dart
@@ -1,11 +1,11 @@
+import 'package:appflowy/plugins/document/presentation/editor_plugins/mention/mention_page_block.dart';
+import 'package:appflowy/plugins/document/presentation/editor_plugins/plugins.dart';
+import 'package:appflowy_backend/protobuf/flowy-folder/view.pb.dart';
 import 'package:flutter/material.dart';
 
 import 'package:appflowy/plugins/document/presentation/editor_drop_manager.dart';
 import 'package:appflowy/plugins/document/presentation/editor_plugins/copy_and_paste/paste_from_file.dart';
 import 'package:appflowy/plugins/document/presentation/editor_plugins/copy_and_paste/paste_from_image.dart';
-import 'package:appflowy/plugins/document/presentation/editor_plugins/file/file_block_component.dart';
-import 'package:appflowy/plugins/document/presentation/editor_plugins/image/custom_image_block_component/custom_image_block_component.dart';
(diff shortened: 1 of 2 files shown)


frontend/appflowy_flutter/lib/plugins/document/presentation/editor_drop_handler.dart
```diff
@@ -37,12 +37,34 @@ class EditorDropHandler extends StatelessWidget {
...
-      builder: (context, dropState, _) => DropTarget(
-        enable: dropState.isDropEnabled,
-        onDragExited: (_) => editorState.selectionService.removeDropTarget(),
-        onDragUpdated: _onDragUpdated,
-        onDragDone: _onDragDone,
-        child: child,
+      builder: (context, dropState, _) => DragTarget<ViewPB>(
+        onLeave: (_) => editorState.selectionService.removeDropTarget(),
+        onMove: (details) {
+          if (details.data.id == viewId) {
```
LucasXu0, line 43: Do we still need to skip the move action here when returning false in the `onWillAcceptWithDetails` function?

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/draggable_view_item.dart
```diff
@@ -65,6 +65,11 @@ class _DraggableViewItemState extends State<DraggableViewItem> {
       onMove: (data) {
         final renderBox = context.findRenderObject() as RenderBox;
         final offset = renderBox.globalToLocal(data.offset);
+
+        if (offset.dx > renderBox.size.width) {
```
LucasXu0, line 69: Is it to cancel the highlight status of the view item?

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_drop_handler.dart
Xazin: Yes, the `onMove` is called no matter the result of `onWillAcceptWithDetails`, I also assumed that returning false there would be enough to cancel auxiliary callbacks, but it's not.

But we can just return here, no need to remove drop target, let me fix.

frontend/appflowy_flutter/lib/workspace/presentation/home/menu/view/draggable_view_item.dart
Xazin: It's because it will try to compute the drop of the view item in the sidebar, despite it not being relevant (outside of the boundary).

Without this part, when you drag and drop a view item, it can disappear from sidebar or be moved somewhere unexpected.

PR #7009, opened 2024-12-18
fix: the index error when deleting text and then using the up or down arrow key

asjqkkkk:
---

- Using `ClampingScrollPhysics` to standardize the scroll animation when selecting language with arrow keys
- Fix the index error when deleting text and then using the up or down arrow key

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/integration_test/desktop/uncategorized/code_block_language_selector_test.dart
+++ b/frontend/appflowy_flutter/integration_test/desktop/uncategorized/code_block_language_selector_test.dart
@@ -1,6 +1,5 @@
 import 'package:appflowy/generated/locale_keys.g.dart';
 import 'package:appflowy/plugins/document/presentation/editor_plugins/code_block/code_block_language_selector.dart';
-
 import 'package:appflowy_editor/appflowy_editor.dart';
 import 'package:appflowy_editor_plugins/appflowy_editor_plugins.dart';
 import 'package:easy_localization/easy_localization.dart';
@@ -45,8 +44,10 @@ void main() {
     await onKey(tester, LogicalKeyboardKey.enter);
 
     final editorState = tester.editor.getCurrentEditorState();
-    final language =
-        editorState.getNodeAtPath([0])!.attributes['language'].toString();
+    String language = editorState
+        .getNodeAtPath([0])!
+        .attributes[CodeBlockKeys.language]
+        .toString();
     expect(
       language.toLowerCase(),
       defaultCodeBlockSupportedLanguages.first.toLowerCase(),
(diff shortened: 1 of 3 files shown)


frontend/appflowy_flutter/integration_test/desktop/uncategorized/code_block_language_selector_test.dart
```diff
@@ -45,7 +44,7 @@ void main() {
     await onKey(tester, LogicalKeyboardKey.enter);
 
     final editorState = tester.editor.getCurrentEditorState();
-    final language =
+    var language =
```
LucasXu0, line 47: You can replace the 'language' with `CodeBlockKeys.language`.

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/code_block/code_block_language_selector.dart
```diff
@@ -222,6 +222,9 @@ class _LanguageSelectionPopoverState extends State<_LanguageSelectionPopover> {
   void onArrowKey(AxisDirection direction) {
     if (filteredLanguages.isEmpty) return;
     final isUp = direction == AxisDirection.up;
+    if (selectedIndex < 0) {
```
LucasXu0, line 225: In which case, the selectedIndex will be less than 0? 

Alternatively, you can use clamp to ensure the selectedIndex is within the correct range.

```dart
    setState(() {
      if (isUp) { // 
        selectedIndex = selectedIndex == 0 ? length - 1 : selectedIndex - 1;
      } else {
        selectedIndex = selectedIndex == length - 1 ? 0 : selectedIndex + 1;
      }
     // <- clamp the selectedIndex here.
    });
```

frontend/appflowy_flutter/integration_test/desktop/uncategorized/code_block_language_selector_test.dart
asjqkkkk: `CodeBlockKeys.language` is a global const value, and the `language` variable here is a local variable, I'm using 'var' here for convenience in case I need to reuse it later, so I don't have to recreate the object.

Xazin: We don't use `var` as it is ambiguous, declare the type properly.

asjqkkkk: This's a reasonable explanation, I'll declare it as `String`

LucasXu0: @asjqkkkk I was referring to the attribute key, not the variable name.

asjqkkkk: I've changed `attributes['language']` to `attributes[CodeBlockKeys.language]`

Issue #6248, opened 2024-09-09
[Bug] Icon search missing result


Timisorean:
### Bug Description

For example, when I search the calendar icon: 📅 

As you can see, it exists:
![image](https://github.com/user-attachments/assets/c8199c7d-7ce8-4c21-9eee-09586f9341c2)

But search not finding it:
![image](https://github.com/user-attachments/assets/0353fcc1-fc65-4deb-a2ff-5dc9c2cfaa1e)
![image](https://github.com/user-attachments/assets/8ee9b25f-6a42-4ec4-81af-77b4625330f8)

### How to Reproduce

1. On a document, "Add icon":
![image](https://github.com/user-attachments/assets/cae4a2c7-a852-4d79-b3fc-c7cccd7ab798)
2. Search

### Expected Behavior

Find the corresponding item.

### Operating System

macOS Sonoma Version 14.6.1

### AppFlowy Version(s)

Version 0.6.8

annieappflowy:
Fixed in the upcoming release (v0.7.9)

PR #7015, opened 2024-12-19
feat: support recent section in icon picker

asjqkkkk:
### Feature Preview

Add the recent icons and emojis to the selector

Recent Emoji | Recent Icon
---|---
image|  image

### Bug fixes
- Make the Tooltip corresponding to Emojis appear above in the selector.
- Dismiss Tooltip while scrolling
- Improve the logic to filter emojis or icons by keyword (#6248)
- The focus disappears after the Emoji Picker is opened 

[Filter Emoji] Before:

https://github.com/user-attachments/assets/449dcce0-b5d4-41e7-8cb4-7a6240effcc2

[Filter Emoji] After:

https://github.com/user-attachments/assets/637d344c-776d-43e9-a69f-c98757eb9397

---

#### PR Checklist


frontend/appflowy_flutter/lib/shared/icon_emoji_picker/recent_icons.dart
```diff
@@ -0,0 +1,89 @@
...
+import 'flowy_icon_emoji_picker.dart';
+
+class RecentIcons {
+  static final Map<String, List<String>> _dataMap = {};
+  static bool _loaded = false;
+  static const maxLength = 20;
+  @visibleForTesting
+  static bool enable = true;
+
+  static Future _save() async {
```
Xazin, line 19: Please declare return types always:
```suggestion
  static Future<void> _save() async {
```
Same for all methods in this class.

frontend/appflowy_flutter/integration_test/desktop/database/database_row_page_test.dart
```diff
@@ -22,7 +23,7 @@ import '../../shared/util.dart';
 
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
-
+  RecentIcons.enable = false;
```
LucasXu0, line 26: Please add a comment to explain why `RecentIcons.enable` is set to false. I suggest disabling it in the `setupAll` function and enabling it in the `tearDownAll` function to avoid side effects.

```suggestion

  setUpAll(() {
    RecentIcons.enable = false;
  });

  tearDownAll(() {
    RecentIcons.enable = true;
  });

```

frontend/appflowy_flutter/lib/plugins/base/emoji/emoji_picker.dart
```diff
@@ -24,28 +27,49 @@ class FlowyEmojiPicker extends StatefulWidget {
 }
 
 class _FlowyEmojiPickerState extends State<FlowyEmojiPicker> {
-  EmojiData? emojiData;
+  late EmojiData emojiData;
+  bool loaded = false;
+
+  void loadEmojis(EmojiData data) {
```
LucasXu0, line 33: Move this function below the build function. We prefer to place private functions after public functions.

frontend/appflowy_flutter/lib/plugins/base/emoji/emoji_picker.dart
```diff
@@ -24,28 +27,49 @@ class FlowyEmojiPicker extends StatefulWidget {
...
+      if (v.isEmpty) {
+        emojiData = data;
+        setState(() => loaded = true);
+        return;
+      }
+      final categories = List.of(data.categories);
+      categories.insert(
+        0,
+        Category(
+          id: 'Recent',
```
LucasXu0, line 44: Define the category id (Recent) as a const value.

```dart
  Future<void> loadEmojis(EmojiData data) async {
    final recentEmojiIds = await RecentIcons.getEmojiIds();

    if (recentEmojiIds.isEmpty) {
      emojiData = data;
    } else {
      final categories = [
        Category(
          id: _kRecentEmojiCategoryId,
          emojiIds: recentEmojiIds.take(widget.emojiPerLine).toList(),
        ),
        ...data.categories,
      ];
      emojiData = EmojiData(categories: categories, emojis: data.emojis);
    }

    setState(() => loaded = true);
  }
```

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/icon_picker.dart
```diff
@@ -78,31 +79,57 @@ class FlowyIconPicker extends StatefulWidget {
...
 }
 
 class _FlowyIconPickerState extends State<FlowyIconPicker> {
-  late final Future<List<IconGroup>> iconGroups;
+  final List<IconGroup> iconGroups = [];
+  bool loaded = false;
   final ValueNotifier<String> keyword = ValueNotifier('');
   final debounce = Debounce(duration: const Duration(milliseconds: 150));
 
+  Future loadIcons() async {
```
LucasXu0, line 99:
```suggestion
  Future<void> loadIcons() async {
```

Move this function below the build function

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/icon_picker.dart
```diff
@@ -250,43 +272,49 @@ class _IconPickerState extends State<IconPicker> {
...
-                          },
-                        );
-                },
-              ).toList(),
+            GridView.builder(
+              gridDelegate: SliverGridDelegateWithFixedCrossAxisCount(
+                crossAxisCount: widget.iconPerLine,
+              ),
+              itemCount: iconGroup.icons.length,
+              shrinkWrap: true,
```
LucasXu0, line 280: Would using `shrinkWrap` here cause the first-time loading to be slower?

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/recent_icons.dart
```diff
@@ -0,0 +1,89 @@
...
+import 'flowy_icon_emoji_picker.dart';
+
+class RecentIcons {
+  static final Map<String, List<String>> _dataMap = {};
+  static bool _loaded = false;
+  static const maxLength = 20;
+  @visibleForTesting
+  static bool enable = true;
+
+  static Future<void> _save() async {
```
LucasXu0, line 19: put the private functions after the public functions.

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/recent_icons.dart
```diff
@@ -0,0 +1,89 @@
...
+          (k, v) => MapEntry<String, List<String>>(k, List<String>.from(v)),
+        ),
+      );
+    } on FormatException catch (e) {
+      Log.error('RecentIcons load failed with :$v', e);
+    } on TypeError catch (e) {
+      Log.error('RecentIcons load failed with :$v', e);
+    }
+    _loaded = true;
+  }
```
LucasXu0, line 43:
```suggestion
  static Future<void> _load() async {
    if (_loaded || !enable) {
      return;
    }

    final storage = getIt<KeyValueStorage>();
    final value = await storage.get(KVKeys.kRecentIcons);

    if (value == null || value.isEmpty) {
      _loaded = true;
      return;
    }

    try {
      final data = jsonDecode(value) as Map;
      _dataMap
        ..clear()
        ..addAll(
          Map<String, List<String>>.from(
            data.map((k, v) => MapEntry(k, List<String>.from(v))),
          ),
        );
    } catch (e) {
      Log.error('RecentIcons load failed with: $value', e);
    }

    _loaded = true;
  }
```

frontend/appflowy_flutter/test/unit_test/util/recent_icons.dart
```diff
@@ -0,0 +1,93 @@
+import 'package:appflowy/core/config/kv.dart';
```
LucasXu0, line 1: rename the file name to `recent_icons_test.dart`

frontend/appflowy_flutter/test/unit_test/util/recent_icons.dart
```diff
@@ -0,0 +1,93 @@
...
+import 'package:collection/collection.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+void main() {
+  group('Testing for RecentIcons', () {
+    TestWidgetsFlutterBinding.ensureInitialized();
+    SharedPreferences.setMockInitialValues({});
+    getIt.registerFactory<KeyValueStorage>(() => DartKeyValue());
+    Log.shared.disableLog = true;
```
LucasXu0, line 16: Put these code into setUpAll function.

frontend/appflowy_flutter/integration_test/desktop/database/database_row_page_test.dart
asjqkkkk: Good suggestion，I'll make the changes based on this suggestion

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/icon_picker.dart
asjqkkkk: > Would using `shrinkWrap` here cause the first-time loading to be slower?

Using `shrinkWrap` is to allow the `GridView` to be properly nested within a ListView, and using `GridView` is to control the number of items per row. And the children of the GridView here won't be too many, so there is no need to worry about performance issues

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/recent_icons.dart
```diff
@@ -0,0 +1,97 @@
...
+    return _dataMap[FlowyIconType.emoji.name] ?? [];
+  }
+
+  static Future<List<Icon>> getIcons() async {
+    await _load();
+    final iconList = _dataMap[FlowyIconType.icon.name] ?? [];
+    try {
+      return iconList
+          .map((e) => Icon.fromJson(jsonDecode(e) as Map<String, dynamic>))
+          .toList();
```
Xazin, line 46: Do we need to deserialize from JSON every time we call getIcons? We could consider doing this once instead.

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/recent_icons.dart
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  static Future<void> _save() async {
+    await getIt<KeyValueStorage>().set(
+      KVKeys.kRecentIcons,
+      jsonEncode(_dataMap),
+    );
+  }
+
+  static Future _load() async {
```
Xazin, line 62: Still missing return type on methods in this class.

```suggestion
  static Future<void> _load() async {
```

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/icon_picker.dart
```diff
@@ -78,31 +80,57 @@ class FlowyIconPicker extends StatefulWidget {
...
   final ValueNotifier<String> keyword = ValueNotifier('');
   final debounce = Debounce(duration: const Duration(milliseconds: 150));
 
+  Future<void> loadIcons() async {
+    final localIcons = await loadIconGroups();
+    final recentIcons = await RecentIcons.getIcons();
+    if (recentIcons.isNotEmpty) {
+      iconGroups.add(
+        IconGroup(
+          name: _kRecentIconGroupName,
```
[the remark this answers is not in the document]
asjqkkkk: Localization is not needed at this stage

frontend/appflowy_flutter/lib/shared/icon_emoji_picker/recent_icons.dart
asjqkkkk: There is a variable `_loaded` here, which is used to control the operation of deserialization only once. Moreover, the amount of data related to recent emojis is quite small, so there is no need to worry about its performance overhead for now.

frontend/appflowy_flutter/lib/plugins/base/emoji/emoji_picker.dart
```diff
@@ -105,4 +113,24 @@ class _FlowyEmojiPickerState extends State<FlowyEmojiPicker> {
       },
     );
   }
+
+  void loadEmojis(EmojiData data) {
```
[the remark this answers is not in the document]
asjqkkkk: I think FutureBuilder might be more suitable for fetching singular data. The emoji data and recent emoji data are obtained from different sources, and placing them within the same asynchronous method could reduce the readability of the code

frontend/appflowy_flutter/lib/core/config/kv_keys.dart
```diff
@@ -109,4 +109,9 @@ class KVKeys {
   ///
   /// The value is a boolean string
   static const String hasUpgradedSpace = 'hasUpgradedSpace060';
+
+  /// The key for saving the rencent icons
+  ///
+  /// The value is a json string of [RecentIcons]
+  static const String kRecentIcons = 'kRecentIcons';
```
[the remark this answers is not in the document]
asjqkkkk: I also think this prefix is unnecessary, but I borrowed some coding style from `KVKeys`, which is why I wrote it that way. I have corrected it now.

Issue #6788, opened 2024-11-14
[FR] Add TextFormatter to the time text field in date picker


Xazin:
### Description

We should limit the users option for inputting wrong times in the Date Pickers time text field, we can do this by using a TextFormatter and based on the time format (12-hour vs. 24-hour) can allow them to only input certain characters and in a certain format, eg. `16:14` or `12:30 AM` versus the non-managed options right now like `ABCDEF` that can be input but is wrong (and will be disregarded on submit).

### Impact

Better UX for users, easier to correctly input time.

PR #7024, opened 2024-12-20
chore: add loose restrictions to date time text field

Xazin:
Closes: #6788 

This just adds some loose restrictions to the Date and Time text fields in the Date Picker.

This is done to help the user write the correct format. It's not completely enforced according to the date format, but does limit the user a bit more.

24-hours - Can only enter 5 characters (12:34) and only enter numbers and `:`
12-hours - Can enter `AM` and `PM` as well (12:34 AM) 8 characters

Date field - Can only enter 12 characters (Unsure if should limit it to 10) and allows `.`, `-`, and `/` characters other than numbers.

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/workspace/presentation/widgets/date_picker/widgets/date_time_text_field.dart
+++ b/frontend/appflowy_flutter/lib/workspace/presentation/widgets/date_picker/widgets/date_time_text_field.dart
@@ -278,13 +278,6 @@ class _DateTimeTextFieldState extends State<DateTimeTextField> {
                         key: const ValueKey('date_time_text_field_date'),
                         focusNode: dateFocusNode,
                         controller: dateTextController,
-                        maxLength: 12,
-                        inputFormatters: [
-                          // We can be more restrictive according to the DateFormatPB
-                          FilteringTextInputFormatter.allow(
-                            RegExp('[0-9.-/ ]'),
-                          ),
-                        ],
                         style: Theme.of(context).textTheme.bodyMedium,
                         decoration: getInputDecoration(
                           const EdgeInsetsDirectional.fromSTEB(12, 6, 6, 6),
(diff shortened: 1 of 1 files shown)


richardshiue:
Hmm... I'm not sure I actually like this change. For some users, this helps to give an input that will be correctly parsed. On the other hand, for others (or it's me only) this could be a hindrance. If I put an extra space somewhere in the input, I could hit the length limit and have to go back and correct that before entering my last few characters. 

Also, this prevents users from entering dates in the way that is most comfortable to them. The unit tests are failing since the tests are using human-readable date inputs (Nov 30, 2024) to take advantage of the date parsing library we're using

Xazin:
> Hmm... I'm not sure I actually like this change. For some users, this helps to give an input that will be correctly parsed. On the other hand, for others (or it's me only) this could be a hindrance. If I put an extra space somewhere in the input, I could hit the length limit and have to go back and correct that before entering my last few characters. 
> 
> 
> 
> Also, this prevents users from entering dates in the way that is most comfortable to them. The unit tests are failing since the tests are using human-readable date inputs (`Nov 30, 2024`) to take advantage of the date parsing library we're using

I see what you mean, but how would users know they can type `Nov 30, 2024`? We don't have a way for them to discover this right?

The hint in the textfield show the "correct" format, so users would assume that to be followed.

We can remove the formatter for the Date field if you prefer?

richardshiue:
Yes I think that would be better

Xazin:
> > We don't have a way for them to discover this right?
> 
> You're right. We don't need to though, since if we let users enter dates however they want and we parse them correctly every time, it'll give them flexibility and convenience.
> 
> > We can remove the formatter for the Date field if you prefer?
> 
> Yes I think that would be better

Done, sorry about the delay :)

richardshiue (approved):

Xazin:
@richardshiue Can you merge for me? I no longer have write rights.

PR #7089, opened 2024-12-30
chore: remove tarui

appflowy:
### Feature Preview

---

#### PR Checklist


eloquence:
@appflowy Is there a tl;dr somewhere on why y'all ended up removing Tauri? We're considering it for a new project and am curious if (other than maintaining it alongside Flutter) there were any major downsides for you.

appflowy:
@eloquence The main drawback is that we don’t have a web developer who’s also proficient in Rust. In my opinion, using Tauri is perfectly acceptable.

khorshuheng:
There is one downside with Tauri - there is no standard webview implementation for Linux, so there might be some inconsistent behaviour from one Linux distro from another.

PR #7121, opened 2025-01-02
feat(flutter_mobile): adjust bottom sheet quick action button style

richardshiue:
This is in accordance with the design, which previously hadn't specified the visual tapping effect

Screenshot 2025-01-02 at 10 13 23 AM

Before

https://github.com/user-attachments/assets/315c2ea0-486a-4776-a1e2-8cc3e1796158

After

https://github.com/user-attachments/assets/205e7422-5e40-4c89-ade2-c97579753582

### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/mobile/presentation/database/board/widgets/group_card_header.dart
```diff
@@ -146,6 +142,7 @@ class _GroupCardHeaderState extends State<GroupCardHeader> {
                             context.pop();
                           },
                         ),
+                        const MobileQuickActionDivider(),
```
LucasXu0, line 145: Do we need to add the divider at the end?

richardshiue: That's a really good question. 

I looked at our designs today, and we seem to have many styles of this quick action bottom sheet. So I'm actually not very sure.

For most database quick action bottom sheets, the actions have dividers, and the last item will have a divider under it as well. That's why I added one in this PR, to make it similar to the others in the database.

Screenshot 2025-01-02 at 6 05 33 PM

In the workspace, I see that the actions don't have dividers at all like the screenshot in the original post

And then in the AI chat, there is extra space between the drag handle and the first item.
Screenshot 2025-01-02 at 6 07 03 PM

richardshiue: Do you have a preference? Then I can just make them all the same style.

LucasXu0: In most cases, I think the divider will be used to separate the items. Therefore, I suggest we avoid adding a divider at the beginning of the first item and the end of the last item. However, we can confirm this with the designer.

richardshiue: Alrighty then, sounds good. I'll remove the final dividers and merge. We can revise the design later.

Issue #7111, opened 2024-12-31
[Bug] Navigation Bar Issue on Linked Pages


renjithsamuel:
### Bug Description

When a page linked from another page is clicked, the linked page opens, but the left side navigator bar, highlights the page only if its already spread out, if the particular page is inside lot of branched pages the left nav bar doesn't move to the particular page.

### How to Reproduce

- Have a page inside another page, and keep the root level page collapsed
- Now in the parent page (or from any page), link the child page with the `+` key, and go to the linked page 
- The Page is highlighted internally, but one would expect the navbar to roll open to the redirected page 

### Expected Behavior

The navbar should roll open to the redirected page.

### Operating System

Windows 11

### AppFlowy Version(s)

0.7.7

### Screenshots

Before clicking on link:
![image](https://github.com/user-attachments/assets/c4163d52-90a2-4621-b84a-ddfffaa95848)

After clicking on link:
![image](https://github.com/user-attachments/assets/d517f682-61d6-46b3-84b1-df6c0a0506f0)

Expected:
![image](https://github.com/user-attachments/assets/f999a8c2-1ca0-466e-b0a1-2099595e2c59)

PR #7142, opened 2025-01-04
fix: navigation bar issue on linked pages(#7111)

asjqkkkk:
Fix the issue #7111

https://github.com/user-attachments/assets/15a60405-b0e0-4cfe-95be-b20125ed6e85

### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/util/expand_views.dart
```diff
@@ -0,0 +1,27 @@
...
+bool isViewExpanded(String id) => getViewBloc(id)?.state.isExpanded ?? false;
+
+void recordBloc(String id, ViewBloc bloc) {
+  final set = _recordBlocs[id] ?? {};
+  set.add(bloc);
+  _recordBlocs[id] = set;
+}
+
+void eraseBloc(String id, ViewBloc bloc) {
+  final set = _recordBlocs[id] ?? {};
```
Xazin, line 15: I think `set` is a poor name, maybe just `record` or similar.

asjqkkkk: You are right, I'll rename them

Xazin (approved):

frontend/appflowy_flutter/lib/util/expand_views.dart
```diff
@@ -0,0 +1,27 @@
...
+    _recordBlocs.remove(id);
+  } else {
+    _recordBlocs[id] = blocSet;
+  }
+}
+
+ViewBloc? getViewBloc(String id) {
+  final blocSet = _recordBlocs[id] ?? {};
+  return blocSet.isEmpty ? null : blocSet.first;
+}
```
LucasXu0, line 27: Using a global map to store the view bloc is not a suitable solution in this case. In my opinion, there are two better options:

1. Create a new bloc at the top level of the document and sidebar. Trigger the expand event in this new bloc when the view in the sidebar needs to expand. The sidebar can listen for updates from the bloc using `BlocListener`.
2. If using a global map, create a class to wrap these functions and provide a "namespace." Register this class in GetIt to allow for proper cleanup. Additionally, the names `recordBloc` and `eraseBloc` are too generic and difficult to understand in context.

frontend/appflowy_flutter/lib/workspace/application/tabs/tabs_bloc.dart
```diff
@@ -212,6 +221,27 @@ class TabsBloc extends Bloc<TabsEvent, TabsState> {
     }
   }
 
+  Future<void> _expandAncestors(ViewPB view) async {
+    if (isViewExpanded(view.parentViewId)) return;
+    final value = await getIt<KeyValueStorage>().get(KVKeys.expandedViews);
+    final Map expandedViews = value == null ? {} : jsonDecode(value);
```
LucasXu0, line 227: When using `jsonDecode`, it is better to wrap the function in a `try-catch` function to handle potential parsing errors caused by invalid values.

frontend/appflowy_flutter/lib/util/expand_views.dart
asjqkkkk: After trying out using Bloc, I found that Bloc does not seem to provide a way to listen to events outside of the Bloc.
It can only listen for state changes, although using state could work,  I believe it is not suitable for handling this issue. 
The best solution would be to use a top-level Bloc to manage expansion and collapse data, but that would require too many changes. 
So, I went with the second option and made the suggested changes.

Issue #7148, opened 2025-01-05
Bug: New user was shown a page that is not supposed to be editable


georgem7154:
i wrote my college notes in a page and i clicked new page and i dont seem to find my old one at all anywhere can anybody help, the notes were very much needed for me. i checked in the file storage folder and i have a db file and in colab_db i have a log file full of unicode. but some lines have content of what i typed, could someone help.

khorshuheng:
@georgem7154 Is this a self hosted instance, or is this the official server? Are you using offline/anonymous mode, or log in into an account?

georgem7154:
log in to account

georgem7154:
i have logged in with gmail opened a page and wrote

khorshuheng:
@georgem7154 Can you join our discord channel, and post this in the general section? I can help you with this over there.

georgem7154:
ok  link?

khorshuheng:
https://discord.com/invite/appflowy-903549834160635914

georgem7154:
i posted

khorshuheng:
Root cause of the issue: when the user first create an account, the app led the user to an empty page associated with the `General` space. This is unexpected - AppFlowy currently assumes that the special page associated with a `space` does not have any content and should not be rendered, and the user shouldn't be shown the page in the first place.

The old content was written on this page, and once the user navigate away from this page (creating a new document in this case), there's no way to revisit this page.

The data has been recovered, but we still need to fix the bug, which a new user who log in for the first time landed on the wrong page.

PR #7152, opened 2025-01-06
fix: filter out the space when opening new tab

LucasXu0:
### Feature Preview

closes https://github.com/AppFlowy-IO/AppFlowy/issues/7148

- [x] filter out the space when opeing new tab
- [x] test

---

#### PR Checklist


LucasXu0:
A workaround is to filter out the space when opening tabs. For future improvement, we need to set the correct view id when creating a workspace and avoid sharing the latest view ID across members. @khorshuheng

khorshuheng:
With https://github.com/AppFlowy-IO/AppFlowy-Cloud/pull/1134 , the correct view id should be set after creating a new workspace.

PR #7185, opened 2025-01-10
fix: launch review 0.8.1

richardshiue:
Screenshot 2025-01-10 at 3 14 24 PM

Screenshot 2025-01-10 at 3 15 12 PM

### Feature Preview

---

#### PR Checklist


frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/chat_input/mobile_ai_prompt_input.dart
```diff
@@ -265,6 +269,11 @@ class _MobileAIPromptInputState extends State<MobileAIPromptInput> {
               AIType.appflowyAI => LocaleKeys.chat_inputMessageHint.tr(),
               AIType.localAI => LocaleKeys.chat_inputLocalAIMessageHint.tr()
             },
+            hintStyle: Theme.of(context).textTheme.bodyMedium?.copyWith(
```
LucasXu0, line 272: You can move these colors to an extension file to make it easier to refactor in the future if we have the shared color palette.

richardshiue: Do you think I should move it just to the scope of the chat component or even wider? I don't know how stable the colored palette is at this moment

LucasXu0: Moving to the chat component is enough for now because the colors used here are hardcoded and same, just to ensure that these widgets are using the same color.

richardshiue: OK!

PR #7262, opened 2025-01-21
chore: support pasting image links that have no file extension

richardshiue:
Note to reviewers:

It will be easier to review this PR going commit by commit.

The git diff for the test file looks messed up, but the only change in the most recent commit is to move some ungrouped tests into the group.

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/message/ai_message_action_bar.dart
+++ b/frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/message/ai_message_action_bar.dart
@@ -188,9 +188,6 @@ class CopyButton extends StatelessWidget {
 
   String _getTrimmedPlainText(String plainText) {
     // match and capture inner url as group
-    final singleLineMarkdownImageRegex =
-        RegExp("^!\\[.*\\]\\((${hrefRegex.pattern})\\)\$");
-
     final matches = singleLineMarkdownImageRegex.allMatches(plainText);
     if (matches.length != 1) {
       return plainText;
--- a/frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/message/ai_message_action_bar.dart
+++ b/frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/message/ai_message_action_bar.dart
@@ -13,6 +13,7 @@ import 'package:appflowy/plugins/ai_chat/application/chat_select_sources_cubit.d
 import 'package:appflowy/plugins/document/application/prelude.dart';
 import 'package:appflowy/plugins/document/presentation/editor_plugins/copy_and_paste/clipboard_service.dart';
 import 'package:appflowy/shared/markdown_to_document.dart';
+import 'package:appflowy/shared/patterns/common_patterns.dart';
 import 'package:appflowy/startup/startup.dart';
 import 'package:appflowy/util/theme_extension.dart';
 import 'package:appflowy/workspace/application/sidebar/space/space_bloc.dart';
@@ -170,7 +171,7 @@ class CopyButton extends StatelessWidget {
           );
           await getIt<ClipboardService>().setData(
             ClipboardServiceData(
-              plainText: textMessage.text,
+              plainText: _getTrimmedPlainText(textMessage.text),
               inAppJson: jsonEncode(document.toJson()),
             ),
           );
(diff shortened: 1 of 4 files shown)


frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/message/ai_message_action_bar.dart
```diff
@@ -184,6 +185,19 @@ class CopyButton extends StatelessWidget {
       ),
     );
   }
+
+  String _getTrimmedPlainText(String plainText) {
+    // match and capture inner url as group
+    final singleLineMarkdownImageRegex =
```
LucasXu0, line 191: You can move this regex to `common_patterns.dart` file.

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/custom_paste_command.dart
```diff
@@ -173,44 +170,52 @@ Future<bool> _pasteAsLinkPreview(
...
     textTransaction,
     skipHistoryDebounce: true,
   );
 
-  final linkPreviewTransaction = editorState.transaction;
-  final insertedNodes = [
-    linkPreviewNode(url: text),
+  // convert it to image or link preview node
+  final replacementInsertedNodes = [
+    if (isImageUrl) imageNode(url: text) else linkPreviewNode(url: text),
```
LucasXu0, line 203: Using `? :` may be more readable.

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/custom_paste_command.dart
```diff
@@ -173,44 +170,52 @@ Future<bool> _pasteAsLinkPreview(
...
-    0,
-    text,
-    attributes: {AppFlowyRichTextKeys.href: text},
-  );
+  final bool isImageUrl;
+  try {
+    isImageUrl = await _isImageUrl(text);
+  } catch (e) {
+    Log.info('unable to get content header');
+    return false;
```
LucasXu0, line 185: I think we should fall back to the following code even if the status code of the HTTP request is not 20

richardshiue: If the link returns anything apart from a 200 and we continue with the rest of the logic, then it will have to be a link preview because I can't assume that an image exists or not.

Additionally, we need to rewrite this test

```dart
testWidgets('paste the image url', (tester) async {
  const plainText = 'https://appflowy.io/1.jpg';
  final image = await rootBundle.load('assets/test/images/sample.jpeg');
  final bytes = image.buffer.asUint8List();
  await tester.pasteContent(plainText: plainText, image: ('jpeg', bytes),
      (editorState) {
    final node = editorState.getNodeAtPath([0])!;
    expect(node.type, ImageBlockKeys.type);
    expect(node.attributes[ImageBlockKeys.url], isNotEmpty);
  });
});
```

Original flow:

1. begin try check for link preview
2. url is valid, has protocol, but is image (ends with jpg)
3. return false to cancel adding link preview
4. try paste image from clipboard buffer instead
5. success

New flow:

1. begin try check for link preview
2. url is valid, has protocol
3. check content header, get 404
4. add link preview block, ignore image in clipboard buffer
5. success

frontend/appflowy_flutter/lib/plugins/ai_chat/presentation/message/ai_message_action_bar.dart
richardshiue: Done!

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/custom_paste_command.dart
richardshiue: Done

frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/copy_and_paste/custom_paste_command.dart
LucasXu0: Ok. I misunderstood the PR flow. I thought it would return without handling the rest of the logic. So, LGTM for the PR workflow.

appflowy:
LGTM!

PR #7490, opened 2025-03-10
fix: don't include icon while exporting callout to md

richardshiue:
There isn't a call out block in markdown, so we convert to quote instead. We also include the icon (emoji) in the first line to preserve the information. Icons, however, can't be converted.

Screenshot 2025-03-10 at 11 10 09 AM

### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/parsers/callout_node_parser.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/parsers/callout_node_parser.dart
@@ -1,4 +1,5 @@
 import 'package:appflowy/plugins/document/presentation/editor_plugins/plugins.dart';
+import 'package:appflowy/shared/icon_emoji_picker/flowy_icon_emoji_picker.dart';
 import 'package:appflowy_editor/appflowy_editor.dart';
 
 class CalloutNodeParser extends NodeParser {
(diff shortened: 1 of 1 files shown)


frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/parsers/callout_node_parser.dart
```diff
@@ -10,7 +10,6 @@ class CalloutNodeParser extends NodeParser {
   @override
   String transform(Node node, DocumentMarkdownEncoder? encoder) {
     assert(node.children.isEmpty);
-    final icon = node.attributes[CalloutBlockKeys.icon];
```
LucasXu0, line 13: We need to check whether the icon value is an icon type or an emoji type. If it's an emoji type, we should include it in the Markdown.

richardshiue: Done!

LucasXu0: The CI failed due to merge conflicts.

Merge the latest main branch after this PR is merged: https://github.com/AppFlowy-IO/AppFlowy/pull/7491

PR #7499, opened 2025-03-11
fix: ai writer gestures

richardshiue:
1. When generating, disable scrolling along with any mouse interaction. Such interactions will pop a confirmation dialog to cancel and remove
2. When not generating, enable scrolling, but don't allow any other sort of mouse interaction, including but not limited to:
    - tapping
    - tap and hold
    - tap and drag
    - three finger drag

### Feature Preview

---

#### PR Checklist


LucasXu0:
Should we also block the long press gesture?

richardshiue:
> Should we also block the long press gesture?

I believe it's handled! The dialog opens immediately on pointer down.

LucasXu0 (approved):

PR #7509, opened 2025-03-12
fix: prevent segfault due to infinite recursion in trash view

khorshuheng:
Currently, if the trash's parent and child hierarchy forms a cycle, the app will fail to launch due to segmentation fault, caused by infinite recursion. There is already a method in AppFlowy Collab which address this, so we just need to reuse the same method.

### Feature Preview

---

#### PR Checklist


richardshiue (approved):
LGTM

PR #7590, opened 2025-03-23
chore: support switch ai model in chat/ AI writer

appflowy:
### Feature Preview

---

#### PR Checklist


appflowy:
image

image


Select local AI directly

image

frontend/appflowy_flutter/lib/ai/service/ai_prompt_input_bloc.dart
```diff
@@ -18,49 +12,31 @@ part 'ai_prompt_input_bloc.freezed.dart';
 
 class AIPromptInputBloc extends Bloc<AIPromptInputEvent, AIPromptInputState> {
   AIPromptInputBloc({
+    required String objectId,
     required PredefinedFormat? predefinedFormat,
-  })  : _listener = LocalAIStateListener(),
-        super(AIPromptInputState.initial(predefinedFormat)) {
+  })  : _aiModelStateNotifier = AIModelStateNotifier(objectId: objectId),
+        super(AIPromptInputState.initial(objectId, predefinedFormat)) {
```
richardshiue, line 18: Can we use the same `AIModelStateNotifier` for the `AIPromptInputBloc` and the `SelectModelBloc`? That way we can speed up loading the models when opening the popup.

frontend/rust-lib/flowy-ai/src/ai_manager.rs
```diff
@@ -238,30 +244,161 @@ impl AIManager {
...
+    let source_key = ai_available_models_key(&source);
+    self
+      .store_preferences
+      .set_object::<AIModel>(&source_key, &model.clone().into())?;
+
+    chat_notification_builder(&source, ChatNotification::DidUpdateSelectedModel)
+      .payload(model)
+      .send();
+    Ok(())
+  }
```
richardshiue, line 308: For AI chats, should we save the selected model into the chat settings?

frontend/appflowy_flutter/lib/ai/service/ai_prompt_input_bloc.dart
appflowy: We can do that, but it adds complexity in managing init, startListening, and stop. If speed is a concern, we can optimize further.

frontend/rust-lib/flowy-ai/src/ai_manager.rs
appflowy: I thought about it, but since not every device has local AI set up, I decided to simply store it locally.

PR #7617, opened 2025-03-25
feat: regenerate message with different model

richardshiue:
#### PR Checklist


frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
```diff
@@ -0,0 +1,174 @@
...
+
+    final editable = localAiState.state == RunningStatePB.Running;
+    final hintText = editable
+        ? LocaleKeys.chat_inputLocalAIMessageHint.tr()
+        : LocaleKeys.settings_aiPage_keys_localAIInitializing.tr();
+
+    return (AiType.local, hintText, editable);
+  }
+
+  (List<AiModel>, AiModel?) getAvailableModels() {
```
appflowy, line 136: why don't we just use AIModelPB directly?

frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
```diff
@@ -0,0 +1,174 @@
...
+  }) {
+    if (onStateChanged != null) {
+      _stateChangedCallbacks.remove(onStateChanged);
+    }
+    if (onAvailableModelsChanged != null) {
+      _availableModelsChangedCallbacks.remove(onAvailableModelsChanged);
+    }
+  }
+
+  Future<void> dispose() async {
```
appflowy, line 100: When multiple parts of our code call addListener, who is responsible for removing them? Because it might be hard to remember I didn't remove others listeners

frontend/appflowy_flutter/lib/plugins/ai_chat/application/chat_bloc.dart
```diff
@@ -505,6 +506,9 @@ class ChatBloc extends Bloc<ChatEvent, ChatState> {
     if (format != null) {
       payload.format = format.toPB();
```
appflowy, line 507: we should use PB version for format and model

frontend/appflowy_flutter/lib/ai/service/select_model_bloc.dart
```diff
@@ -1,82 +1,113 @@
...
-    _aiModelStateNotifier.startListening(
-      onAvailableModelsChanged: (models) {
-        if (!isClosed) {
-          add(SelectModelEvent.didLoadModels(models));
-        }
-      },
-    );
-
+    required AIModelStateNotifier aiModelStateNotifier,
+  })  : _aiModelStateNotifier = aiModelStateNotifier,
```
appflowy, line 38: TBH, the notification stuff is very light weight. We don't need to worry about its performance.

frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
```diff
@@ -0,0 +1,174 @@
...
+import 'package:protobuf/protobuf.dart';
+import 'package:universal_platform/universal_platform.dart';
+
+typedef OnModelStateChangedCallback = void Function(AiType, bool, String);
+typedef OnAvailableModelsChangedCallback = void Function(
+  List<AiModel>,
+  AiModel?,
+);
+
+class AIModelStateNotifier {
```
appflowy, line 19: I don't know why we have lib/ai directly. why not just put in the chat directory. we can move service to application folder under chat.

frontend/appflowy_flutter/lib/ai/service/select_model_bloc.dart
```diff
@@ -1,82 +1,113 @@
...
-    AvailableModelsPB models,
+    List<AiModel> models,
+    AiModel? selectedModel,
   ) = _DidLoadModels;
 }
 
 @freezed
 class SelectModelState with _$SelectModelState {
   const factory SelectModelState({
-    AvailableModelsPB? availableModels,
```
appflowy, line 80: I prefer use PB directly. We use PB as single source of true for data struct that used in flutter side

frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
richardshiue: This allows me to extend the AiModel type without using extension methods. For example, for i18n, icon, description.

frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
richardshiue: I designed it to be like Flutter's other controllers or notifiers. If a widget or a bloc instantiates the notifier, then it is responsible for disposing it. If a widget or bloc adds a listener, it is responsible for removing it.

frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
richardshiue: I put it like this because the AI writer is an editor plugin in the document plugin, while AI chat is it's own chat plugin. And then, they use shared components in lib/ai. Should I change this?

frontend/appflowy_flutter/lib/ai/service/select_model_bloc.dart
richardshiue: https://github.com/user-attachments/assets/2a5caee2-ea55-4617-8560-04bf0d6fc367

appflowy: Ok. It likes a UI refresh issues. Because it's async action. It doesn't releate to performance. And yes, by using sharing instance can fix this UI flashing

frontend/appflowy_flutter/lib/ai/service/ai_model_state_notifier.dart
appflowy: I can't answer you how to do it now. I will take a look later.

Issue #6422, opened 2024-09-27
[Bug] AppFlowy is not visible on Launch (first-time installation)


Joshua-E-Weinstein:
### Bug Description

Appflowy's GUI is not launching after first-time installation on my system. First-time installation was with version 0.6.9.

I have tried a large variety of troubleshooting:
- Uninstall/Reinstall with versions 0.5.3, as well as all versions between 0.5.8-0.7.0
- Clearing all AppFlowy data in %AppData% and all remnant files in install location when uninstalling.
    -note: %AppData%/io.appflowy/AppFlowy/data never had any files in it after installation.
- Debugging, throwing logs from the exe. No output to be seen.

### How to Reproduce

1. Install AppFlowy for the first time on a Windows 10 PC.
2. Launch AppFlowy

### Expected Behavior

AppFlowy GUI app opens to the login screen.

### Operating System

Windows 10

### AppFlowy Version(s)

0.5.3, and all versions between 0.5.8-0.7.0

### Screenshots

![image](https://github.com/user-attachments/assets/076f4c9a-82fc-474b-82da-bda1c366ba1e)

### Additional Context

-You can also see a blank window for `AppFlowy` when switching tabs, but you cannot open it nor see it anywhere. It has not loaded yet, but counts as an app.
-You can see location in use.

birnam:
Same problem here. I've tried with the installer version of 0.7.1, and portable (.zip) versions of 0.7.1, 0.7.0, 0.6.9, and 0.5.9.

When I run AppFlowy, I can see a "Background processes" listing in Task Manager, but no other indication of an app running. I've tried a fresh restart. No windows, no icons in the task bar, no icons in the tray.

Xazin:
I took out my old Windows 10 laptop (x86_64) which hasn't been used in 5 years and had a nearly fresh install of windows, and I couldn't reproduce.

I tried walking through some possible fixes with Josion on Discord, but we didn't manage to find anything that worked.

We will keep exploring what can be causing this.

birnam:
Thank you @Xazin -- I'm on Windows 11 23H2 (22631.4317). Can you tell me, from a task manager perspective, is there supposed to be an AppFlowy in both Apps and Background Processes, or just the one in Background Processes like I'm seeing?

I'm a heavy virtual desktop user. I tried scanning through the other desktops to see if it had launched somewhere else but did not find the app. I wonder if it's related though.

Xazin:
[...]

Issue #6979, opened 2024-12-12
[Bug] Appflowy 0.7.7 is not opening / failing to launch / does not apear on Windows 10


candideu:
### Bug Description

I uninstalled my old Appflowy install (probably 0.7.2 or older) from my Windows 10 PC, and the app doesn't open anymore, despite appearing in the task manager...

### How to Reproduce

Install Appflowy 0.7.7 exe from Github

### Expected Behavior

App opens normally

### Operating System

Win 10

### AppFlowy Version(s)

0.7,7

WinXaito:
I don't know if it's related, but I had a similar problem when I tried to install on a virtual machine.

I had the same problem yesterday when I tried to opened it on a VirtualBox VM (Windows 11). The app is shown in task manager, but nothing rendered on the screen, no error no log.

I tried v0.7.7 (installer and portable) and 0.7.6 (portable). Nothing work.

I finally tried the v0.7.7 on my computer (not a VM) (Windows 10) and it worked.

candideu:
@LucasXu0 no erro; it just doesn't open at all when I click on the .exe

I'm not using a virtual machine

LucasXu0:
@candideu Could you check if version 0.7.6 works for you?

candideu:
@LucasXu0 Unfortunately it doesn't: both the standalone and installed app don't open (I always do a fresh install).

shunuy:
Windows 11
both the standalone and installed app （0.7.7） don't open

shunuy:
![image](https://github.com/user-attachments/assets/150d39e1-32bd-4f09-bd44-ad05768e9a95)
local_notifier_plugin.dll load OK
but，app don't open

asjqkkkk:
![image](https://github.com/user-attachments/assets/35dd1ee1-72d5-4804-bdb7-68a7cb16d48d)

There is no issue to open it on windows 11

Have you tried to delete the data folder : [C:\Users\[Username]\AppData\Local\Temp\appflowy_flutter...]

candideu:
> Have you tried to delete the data folder , then uninstall it and install it again : [C:\Users[Username]\AppData\Local\Temp\appflowy_flutter...]

@asjqkkkk There is no such folder in that directory (I'm on Windows 10)...

I could only find a folder called "io.appflowy" in AppData/Roaming

candideu:
[...]

Issue #6677, opened 2024-10-30
[Bug] Desktop app does not open after fresh install


ReisMiner:
### Bug Description

I downloaded and installed the latest version of AppFlowy for Desktop. I never had it installed on my computer so there were no left over traces from a previous installation. 

After it was installed, i tried to open it but the program never showed up. 

In the task manager it was staying at around 70MB RAM usage with no change, which seemed kinda odd to me.

Then i tried to install a less recent version, same issue...

### How to Reproduce

To be honest, not really sure, maybe some program interferes with AppFlowy, i don't know.

But as said, i never installed it before, so no traces and no version conflicts etc.

### Expected Behavior

I expected for the program to open and show something.

### Operating System

Windows 10 Pro Version 22H2 Build 19045.5011

### AppFlowy Version(s)

0.7.2

RainSharky:
I can confirm this. I was confused at first and thought maybe it didn't install right or something.

**Specs**
Edition: Windows 10 Enterprise
Version: 22H2
system build number: 19045.4894

hug0b:
When starting AppFlowy.exe, the Appflowy process is visible in the task manager but no window opens.

Things I have tried:
- uninstalling the latest version and installing a previous version -> Same issue
- starting the executable from CMD -> no additional terminal output
- Checking out logs under "\AppData\Roaming\io.appflowy" -> there doesn't seem to be any new log generated

Here are some of the last error lines in the latest log file:

`{"msg":"flowy_user::event_handler","time":"10-31 09:30:19","target":"flowy_user::event_handler","error":"NotSupportYet: Not support yet"}`

`{"msg":"[GET_CHAT_CONFIG - EVENT] flowy_ai::local_ai::local_llm_resource","time":"10-31 09:30:19","target":"flowy_ai::local_ai::local_llm_resource","error":"LocalAIError: Local AI resources are not ready"}`

`{"msg":"[INIT - EVENT] [LLM Resource] Failed to watch offline app path: FlowyError { code: Internal, msg: \"Unsupported platform for offline app watching\", payload: [] }","time":"10-31 09:30:19","target":"flowy_ai::local_ai::local_llm_resource"}`


This is kind of annoying because I have been using Appflowy daily to take notes.

LucasXu0:
> starting the executable from CMD -> no additional terminal output

Hi, @hug0b. Is there any output when launching AppFlowy from CMD?

[...]

Issue #5874, opened 2024-08-04
[Bug] No visible window on Windows


eugenesvk:
### Bug Description

AppFlowy-0.6.6-windows-x86_64 fails to show its GUI/Window even though the app is running (it's visible in the Task Manager, can also see the app in the app switcher)

### How to Reproduce

Install AppFlowy
Launch it

### Expected Behavior

The app launches and shows its GUI

### Operating System

Windows 10

### AppFlowy Version(s)

AppFlowy-0.6.6-windows-x86_64

### Screenshots

Taskbar is empty (there is also no visible window)
<img width="522" alt="Taskbar" src="https://github.com/user-attachments/assets/d263133d-1b8d-42ce-9698-940e7beba779">

But the alt-tab app switcher shows the app is running

<img width="164" alt="AppSwitcher" src="https://github.com/user-attachments/assets/4917b938-5007-4763-a486-9dd529c47d32">

### Additional Context

AppFlowy-0.4.9-windows-x86_64 shows a Window, but it's just a blank empty  one, removing appdata as suggested https://github.com/AppFlowy-IO/AppFlowy/issues/5602 didn't help

PhilipElonroad:
Hi, 

I had the same problem and found a workaround for me. AppFlowy seem to get off screen for some reason ( I use three monitors). I tried uninstalling and deleting all previous application files and cache and reinstalled but still the AppFlowy window didn't display on screen.

I use the third party software Alt-Tab Terminator for window management and it has a feature to maximize an application. Maximizing the AppFlowy brought it back to my primary monitor. You can likely do the same with a PowerShell script if you look up how.

![appflowy_offscreen_bug_solution_241107](https://github.com/user-attachments/assets/1137b267-0487-41f5-a6cc-729458a65bea)

Windows 10
AppFlowy 0.4.3
Alt-Tab Terminator 5.7
https://www.ntwind.com/software/alttabter.html

candideu:
Were you able to solve this? I'm having the exact same issue...

candideu:
I had the same issue, and solved it this way: https://github.com/AppFlowy-IO/AppFlowy/issues/6979#issuecomment-2571743828

PR #7782, opened 2025-04-18
chore: bumped device_info_plus to 11.2.2

birnam:
### Bug Resolution

fixes https://github.com/AppFlowy-IO/AppFlowy/issues/6422
and the related issues:
https://github.com/AppFlowy-IO/AppFlowy/issues/6677
https://github.com/AppFlowy-IO/AppFlowy/issues/6979
https://github.com/AppFlowy-IO/AppFlowy/issues/5874

Bug: When running AppFlowy.exe on some Windows machines, an uncaught error stopped all app processes resulting in the window never appearing and the application processes to be stuck in 'Background processes' in Task Manager. Unlike some similar reports, this case was not caused by the window appearing offscreen or corrupted AppData files.

Cause: A system call to `GetPhysicallyInstalledSystemMemory` returns an unexpected `0`, causing device_info_plus to throw an error when used to retrieve a deviceId. See device_info_plus issue https://github.com/fluttercommunity/plus_plugins/issues/2957 for their discussion. After doing more research into this on my own, I found the root cause. 

On Windows, device_info_plus calls `GetPhysicallyInstalledSystemMemory` to get the amount of system memory. This function is known to fail in Windows VMs, but I found that it also fails when Microsoft Hyper-V is active on the computer. I believe Hyper-V to be the missing commonality between the multiple issues linked above. Hyper-V is enabled to meet various requirements, such as WSL, Docker Desktop, or running other hypervisors in order to pass CPU based virtualization to other VMs.

The device_info_plus fix was to return 0 when `GetPhysicallyInstalledSystemMemory` failed, which prevents the crash although does not actually provide the memory information. This is not an issue for AppFlowy since we only need the deviceId, although I will recommend they include a fallback call to `GlobalMemoryStatusEx` to actually retrieve the system memory information when the existing call fails. 

For reference, this is their comment on this solution from the linked issue. The other two properties mentioned involve calls to GetComputerNameEx and GetUserName. In my tests, neither of these have shown vulnerable to the VM/Hyper-V condition, and regardless neither of these omissions affect AppFlowy.

> There are three properties that will throw a WindowsException when the result is 0, which has been like that since 2022 when it was originally implemented, so I am kind of surprised no one proposed a change.
> 
> I will do a PR returning default values instead of throwing, as I don't see the benefit of throwing an error, specially if 3rd party packages are not taking care of caching possible errors.

Fix description: the device_info_plus dependency override was updated to 11.2.2, which was the first to integrate the 2957 fix. (There are newer 11.3.x releases but I found dependency clashes when trying the latest version in AppFlowy)

---

#### PR Checklist

## Summary by Sourcery

Update device_info_plus dependency to version 11.2.2 to resolve Windows-specific memory retrieval issues

Bug Fixes:
- Fixed an issue causing application crashes on Windows machines with Hyper-V enabled by updating the device_info_plus package to handle memory information retrieval errors

Chores:
- Updated pubspec.yaml to override device_info_plus dependency to version 11.2.2

### DIFF
--- a/frontend/appflowy_flutter/pubspec.lock
+++ b/frontend/appflowy_flutter/pubspec.lock
@@ -533,10 +533,10 @@ packages:
     dependency: "direct main"
     description:
       name: device_info_plus
-      sha256: a7fd703482b391a87d60b6061d04dfdeab07826b96f9abd8f5ed98068acc0074
+      sha256: "72d146c6d7098689ff5c5f66bcf593ac11efc530095385356e131070333e64da"
       url: "https://pub.dev"
     source: hosted
-    version: "10.1.2"
+    version: "11.3.0"
   device_info_plus_platform_interface:
     dependency: transitive
     description:
--- a/frontend/appflowy_flutter/pubspec.yaml
+++ b/frontend/appflowy_flutter/pubspec.yaml
@@ -177,7 +177,7 @@ dev_dependencies:
 
 dependency_overrides:
   http: ^1.0.0
-  device_info_plus: ^10.1.0
+  device_info_plus: ^11.2.2
 
   url_protocol:
     git:


birnam:
ugh trying to get the commit message showing up better but oops'ed. sorry there are no tangible changes in the new commits

CamiloMM:
just curious and sorry if it's a dumb question but will the release after v0.8.9 already have this?

PR #7845, opened 2025-04-28
fix: some mobile search UI issues

asjqkkkk:
### Feature Preview

---

#### PR Checklist

## Summary by Sourcery

Refine the user interface for mobile search results and the selection menu.

Bug Fixes:
- Display a default name for search results when the original name is empty.
- Correct the alignment and increase the tappable area of the search cancel button.
- Ensure the "No results" text in the selection menu uses the correct theme color instead of a hardcoded value.
- Center the icon within its allocated space in the search result cell.
- Add vertical spacing before the summary text in search result cells for better readability.
- Adjust the height of the mobile search text field.


frontend/appflowy_flutter/lib/user/application/reminder/reminder_bloc.dart
```diff
@@ -512,19 +496,11 @@ class ReminderBloc extends Bloc<ReminderEvent, ReminderState> {
   ) async {
     final List<ReminderPB> availableReminders = [];
     final reminderIds = reminders.map((e) => e.id).toSet();
-    final removedIds = <String>{};
```
LucasXu0, line 515: If we delete the `removedIds`, how do we perform the unreferenced reminder now?

asjqkkkk: putting removeIds here is not a correct way, however, I haven't thought of the most appropriate way to handle it

we'll ignore the unreferenced reminder now

Issue #7850, opened 2025-04-28
[FR] Please use GitHub CI to build AppFlowy packages for Linux Arm64 architecture


nissimk:
### Description

The GitHub workflows create Linux packages for x64 architecture but not for Arm64. Having release packages available for Arm64 will unlock users on Raspberry pi, snapdragon X laptops and Chromebooks with ARM CPU architectures. I have successfully built from source on Arm64 and tested on Raspberry pi and chromeOS and I will send a pull request. I changed a .yml file in the .github folder and also a shell script in frontend/scripts.  I can test the script but I don't know how to test the changes to the GitHub workflow. Should I submit the PR without testing the GitHub workflows change or is there some way that I can test it myself in my fork?

### Impact

This will create release artifacts that can be installed on Raspberry pi, Snapdragon X laptops and Chromebooks with ARM cpu's.

LucasXu0:
@nissimk I'll do that.

PR #7858, opened 2025-04-29
ci: get github ci to build arm64 linux deb and rpm release packages

nissimk:
This is the first attempt to get GitHub ci to build release packages for Arm64 Linux. I tested the script on my Raspberry pi but I have not tested the .yml file changes. Please let me know if there are errors when trying to run the workflow in GitHub and I can fix them here.

addresses https://github.com/AppFlowy-IO/AppFlowy/issues/7850

A couple of things to note:

1) I skip the AppImage build on Arm64 because the appImage repo seems to only have builds for amd64 architecture. I left in the alien call to convert the .Deb to an rpm, but I have not tested that.

2) when developing on arm architecture you get errors when trying to commit because the dev environment script downloads the amd64 binary of go-gitlint and puts it in the .githooks folder. I was able to solve this by building go-gitlint from source, but in case the ci does any git commits it's going to need to do this also when running on an Arm64 runner.


LucasXu0:
Hi @nissimk, could you add ARM64 support to the `flutter_ci.yaml`, except for the `release.yml`?

nissimk:
So this still is not building...I figured out how to run the actions in my fork, so I've been testing there: https://github.com/nissimk/AppFlowy/pull/2

The first issue was that the flutter-action was not working for aarch64. There's an issue here: https://github.com/subosito/flutter-action/issues/345 with a workaround to use master instead of stable for the channel. I was trying to do that for just aarch64 but I couldn't figure that out so I just changed the channel to master for all architectures. Not sure what impact this has.

The next issue that I haven't been able to get around is the install-action for cargo-make fails to find the installer for cargo-make on aarch64. I'm not sure how to work around this one without changing the action to manually installing the appropriate version of cargo-make or contributing to the install-action upstream repo.

Do you have any suggestions of how to proceed?

LucasXu0:
I have tried to set up Flutter on arm64 in GitHub Actions, but neither of my attempts worked.

PR #7863, opened 2025-04-30
fix: edit link will break the text styles

asjqkkkk:
### Feature Preview

---

#### PR Checklist

### DIFF
--- a/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/desktop_toolbar/link/link_extension.dart
+++ b/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/desktop_toolbar/link/link_extension.dart
@@ -8,18 +8,13 @@ extension LinkExtension on EditorState {
     if (node == null) {
       return;
     }
+    final attributes = _getAttribute(node, selection)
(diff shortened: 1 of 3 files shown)


frontend/appflowy_flutter/lib/plugins/document/presentation/editor_plugins/desktop_toolbar/link/link_extension.dart
```diff
@@ -8,32 +8,65 @@ extension LinkExtension on EditorState {
     if (node == null) {
       return;
     }
+    final attributes = _getAttribute(node, selection);
```
LucasXu0, line 11: I think we're not getting all the attributes from the selection. For example:

Before:

_http_://**appflowy**.io <- (add one more o here)

After:

_http_://**appflowy**.ioo The result should be ioo without extra formatting.

We have a slice attributes function that gets the attributes based on the index. `appflowyEditorSliceAttributes`

asjqkkkk: yes, the logic about getting Attributes is not correct, I need to adjust it

PR #7905, opened 2025-05-09
fix: allow ':keyword' make mistakes

asjqkkkk:
When inserting an emoji by using “:”, allow the continuation of input with a string of up to 5 characters even if no results are found

### Feature Preview

---

#### PR Checklist

## Summary by Sourcery

Improve emoji search input handling by allowing users to continue typing and exploring emoji suggestions even when initial results are empty

Bug Fixes:
- Prevent premature dismissal of emoji picker when no initial results are found

Enhancements:
- Extend emoji search input tolerance to allow continued typing up to 5 additional characters
- Add a 'no results' message when no emojis match the current search query


asjqkkkk:
> The change is good, but it seems like it's become too lenient now:
> 
>  Screen.Recording.2025-05-12.at.2.01.33.PM.mov

This issue has existed since the previous version, and I haven't found a good solution to address it yet

richardshiue (approved):

PR #7916, opened 2025-05-12
fix: using the shortcut to show the search panel will hide the 'Ask AI anything' button

asjqkkkk:
- Fix: using the shortcut to show the search panel will hide the 'Ask AI anything' button
- Fix: layout error with search result lists
- Fix: improve the behavior for displaying "document path" in search results preview

### Feature Preview

---

#### PR Checklist

## Summary by Sourcery

Fix UI issue where toggling the search panel via shortcut would hide the 'Ask AI anything' button and improve state synchronization between sidebar and command palette.

Bug Fixes:
- Ensure the 'Ask AI anything' button remains visible when toggling the search panel via shortcut.

Enhancements:
- Synchronize workspace and space bloc state with the command palette after sidebar initialization.


richardshiue:
This solution means that if the sidebar is collapsed at startup, then the CommandPaletteValue will still be null

asjqkkkk:
> This solution means that if the sidebar is collapsed at startup, then the CommandPaletteValue will still be null

yes, thanks for reminding me, I've updated the logic

Issue #8668, opened 2026-04-15
[Bug] Settings page shows error code when opened without internet connection (App Store version)


danteboe:
### Bug Description

When opening the Settings page without an internet connection, an error code is visibly displayed in the UI. This feels unintended and breaks the user experience, especially since accessing settings should not require connectivity.

### How to Reproduce

1. Open AppFlowy (App Store version).
2. Disconnect from the internet.
3. Navigate to Settings.
4.Observe the error message/code displayed

### Expected Behavior

App: AppFlowy (App Store version)
Platform: iOS
Internet: Disconnected

### Operating System

iOS 26.3.1

### AppFlowy Version(s)

0.11.7

### Additional Context

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/216c35e9-5774-46af-92c1-ab97c7cff507" />

PR #8790, opened 2026-06-05
Hide raw cloud config errors in settings

miachillgood:
## Summary
- replace raw cloud-config error strings in settings with user-friendly sync error copy
- keep the failure path non-technical when cloud config cannot be loaded

## Why
When settings loads while offline or cloud config fetch fails, the UI should not surface raw backend/internal error text directly to the user.

Related to #8668.

## Validation
- git diff --check

## Notes
- I could not run Dart/Flutter validation locally because this machine does not have `dart` or `flutter` installed.

## Summary by Sourcery

Replace raw cloud configuration error messages in settings with localized, user-friendly sync error copy.

Bug Fixes:
- Prevent internal cloud configuration error strings from being shown directly to users in settings by using generic sync error messaging.

Enhancements:
- Use localized sync error title and hint text on cloud settings error pages instead of raw exception strings.


miachillgood:
Follow-up pushed on the same branch. I kept the user-facing copy generic, added explicit logging for the underlying cloud-config error, and now show a more specific no-network message when the error looks connectivity-related.

PR #8866, opened 2026-07-17
feat: show and adjust app zoom level in workspace settings

### LINKED ISSUE #5396: [FR] Show current UI zoom state on settings page ("Workspace" section)
### Description

Just polishing this feature, that users were waiting for so long.

### Impact

In some cases user could be need to see current zoom state, in order to set better scaling level.

### Additional Context

_No response_

hkarmoush:
### What this does

Fixes #5396.

Right now the app zoom can only be changed with the keyboard (`Cmd/Ctrl` + `+` / `-` / `0`), and the current zoom level isn't shown anywhere. If you've nudged the zoom and lost track of where you are, there's no way to tell without resetting.

This adds a small **Zoom** control to **Settings → Workspace**, right under Appearance. It shows the current zoom as a percentage and lets you step it up/down with `−` / `+` buttons or reset it — using the exact same 0.5×–2.0× range and 0.1 step as the existing shortcuts.

### A small cleanup along the way

The logic that actually applies a scale factor (update the binding, or the test-mode value, then persist it) was copy-pasted in two places — the zoom hotkeys and the mobile display-size setting. Rather than add a third copy for this new control, I pulled it into a single `applyAppScaleFactor()` helper that all three now call. Behaviour is unchanged; it's just one source of truth now.

### How it works

- New `applyAppScaleFactor()` in `hotkeys.dart`, next to the existing `appflowyScaleFactor` state it already owns.
- The hotkey handler and the mobile setting now delegate to it.
- New `_AppZoomSetting` widget in `settings_workspace_view.dart`, built from the same `SettingListTile` / `SettingsResetButton` / `FlowyIconButton` pieces used by the neighbouring settings.
- One new English string pair under `settings.workspacePage.appZoom` in `resources/translations/en-US.json`.

### Feature preview

Settings → Workspace now shows: `App zoom level   [ − ]  100%  [ + ]   ⟲`

### Testing

Added an integration test (`integration_test/desktop/settings/workspace_zoom_test.dart`, registered in `settings_runner.dart`) that opens Workspace settings, taps `+` twice → 120%, `−` once → 110%, reset → 100%, asserting both the live scale factor and the persisted value each step — mirroring the existing `zoom_in_out_test.dart`.

Note: I wasn't able to run the analyzer / test suite locally (my machine's Flutter is ahead of the pinned 3.27.4), so I'm relying on CI for the full check and will fix anything it flags.

### PR checklist

- [x] I've added a test to validate the changes in this PR.
- [ ] All existing tests are passing. *(deferring to CI — see the testing note above)*

## Summary by Sourcery

Add a configurable app zoom control to workspace settings and centralize zoom scale application logic.

New Features:
- Expose app zoom level in Workspace settings with controls to increase, decrease, and reset the zoom within the existing supported range.

Enhancements:
- Extract shared applyAppScaleFactor helper to unify how zoom scale changes are applied and persisted across hotkeys, desktop settings, and mobile display size settings.

Tests:
- Add an integration test verifying zoom in/out and reset behavior from workspace settings, including both runtime and persisted scale factors.

### DIFF
--- a/frontend/appflowy_flutter/integration_test/desktop/settings/settings_runner.dart
+++ b/frontend/appflowy_flutter/integration_test/desktop/settings/settings_runner.dart
@@ -4,6 +4,7 @@ import 'notifications_settings_test.dart' as notifications_settings_test;
 import 'settings_billing_test.dart' as settings_billing_test;
 import 'shortcuts_settings_test.dart' as shortcuts_settings_test;
 import 'sign_in_page_settings_test.dart' as sign_in_page_settings_test;
+import 'workspace_zoom_test.dart' as workspace_zoom_test;
 
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
@@ -12,4 +13,5 @@ void main() {
   settings_billing_test.main();
   shortcuts_settings_test.main();
   sign_in_page_settings_test.main();
+  workspace_zoom_test.main();
 }
--- a/frontend/appflowy_flutter/integration_test/desktop/settings/workspace_zoom_test.dart
+++ b/frontend/appflowy_flutter/integration_test/desktop/settings/workspace_zoom_test.dart
@@ -1,0 +1,53 @@
+import 'package:appflowy/startup/tasks/app_window_size_manager.dart';
+import 'package:appflowy/workspace/application/settings/settings_dialog_bloc.dart';
+import 'package:appflowy/workspace/presentation/home/hotkeys.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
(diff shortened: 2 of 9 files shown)


hkarmoush:
Thanks for the review! Pushed a follow-up:

- Restored the `setState`-before-persist order in `text_scale_setting.dart` — good catch, that ordering flipped unintentionally when I extracted `applyAppScaleFactor`.
- While in there, noticed the new settings zoom control didn't update live if you zoomed via the keyboard shortcut while the settings page was open, so I made `appflowyScaleFactor` a `ValueNotifier` and had the widget listen to it. Added a test for that case too.

PR #8928, opened 2026-08-06
fix: handle new card creation from board footer

### LINKED ISSUE #8922: [Bug] Add a new card at the bottom
### Bug Description

In a board, when I try to add a card, I write the card title and click outside it, it crates it but hidden. I hit create once again and I can see it, with the desired name fully selected. If I press enter it creates a card with empty name, deleting the name I gave to it.
Only if I hit enter it creates it.

### How to Reproduce

1. Create or enter to an existing board;
2. Under a column click "+ New" ("Add a new card at the bottom" when hovered);
3. write the name of the card you want to create, eg: "abc";
4. click outside the card -> the card disappears;
5.  Click "+ New" ("Add a new card at the bottom" when hovered) again;
6. A new card is created, with the name "abc", from the eg, all selected. If you click outside, step 4 will happen. If you hit enter, a the acrd will be created, but with name "Untitled";
If, at step 6, deselct the name and press enter the card will be created with the desired name, "abc".

### Expected Behavior

1. When I click outside the card, the card is created, with the desired name, "abc" from the above eg.
2. If I click outside the card and the card is not created, for whatever reason, when I click "+ New" again, the new card should not have any name, not the old one, "abc" or if it has it, it should not be selected at all, for when I press enter it creates it with the desired name, not "Untitled".

### Operating System

Windows

### AppFlowy Version(s)

Version 0.13.0

### Screenshots

_No response_

### Additional Context

- I checked for another references to this bug, I did not found any.
- please, make it so when I click outside the card, the card is created OR when I press enter the card is created.
- when I press ESCAPE after writing a name, a newly generated one should not have the formerly given name, "abc" (might be suitable for another bug).

tollenceld:
Fixes #8922

## Problem

On desktop, in a board's column footer, clicking `+ New`, typing a title, and then clicking outside the input closed the input without creating a card but kept the typed draft. Reopening `+ New` showed the stale draft, and pressing Enter could then create a card with an empty title (reported as "Untitled"). Escape also left the old draft in place for the next attempt.

## Root cause

In `_BoardColumnFooterState` (`frontend/appflowy_flutter/lib/plugins/database/board/presentation/board_page.dart`):

- Losing focus only set `_isCreating = false`; the `TextEditingController` was never cleared, so the draft survived.
- Escape only called `_focusNode.unfocus()`, which went through the same blur path and left the draft intact.
- `onSubmitted` had no empty-title guard, so pressing Enter with an empty title dispatched `BoardEvent.createRow` with an empty name, producing a card with an empty title. Because the input stayed focused after submit and the controller was cleared immediately after dispatch, a second Enter submitted the now-empty title.

## Behavior

Consistent with the board's other trailing create input (`_BoardTrailingState`, "Add a new group", which cancels and clears on Escape and on blur), the footer now follows:

- Enter with a non-empty title → create the card (existing behavior, kept).
- Empty title + Enter → no-op (no empty/"Untitled" card).
- Blur (click outside) → cancel creation and clear the draft.
- Escape → cancel creation and clear the draft.
- Reopening `+ New` always starts with a clean, empty input.

## Changes

- `_BoardColumnFooterState`:
  - Added `_cancelCreating()`, which clears the controller and exits the creating state.
  - The focus-loss listener now calls `_cancelCreating()`, so blur and Escape share one cancellation path (Escape just unfocuses, and the listener performs the single cancellation).
  - `onSubmitted` ignores empty/whitespace-only titles.

## Tests

Extended `frontend/appflowy_flutter/integration_test/desktop/board/board_add_row_test.dart` with five footer scenarios: empty-title submit, click-outside cancel + draft clear, Escape cancel + draft clear, double Enter, and submit-then-click-outside.

Local results:

- `flutter analyze` (full project): **No issues found**.
- `dart format --output=none --set-exit-if-changed` on changed files: **clean**.
- Board integration tests (`board_add_row_test.dart`, macOS local): **4 of 7 passed** (`from header`, `from footer`, `empty title submit`, `clicking outside`). The `escape` test could not complete: `sendKeyEvent(escape)` hangs in the macOS live binding (key-event channel round trip). The same `sendKeyEvent(escape)` pattern is already used by existing board tests (`board_row_test.dart`, `board_hide_groups_test.dart`) on the repo's Linux CI, so this appears to be a local macOS tooling quirk rather than a product regression. The remaining two tests were not run locally.

Not run locally / pending: full integration-test pass on Linux (repo CI runs board tests via `integration_test/desktop_runner_3.dart`) and Windows behavior.

## Manual verification

The contributor manually verified the current fix on a real device and did not observe any issues. The specific device and platform are not claimed here, and this does not replace automated cross-platform CI coverage.

## CI and limitations

- `Ninja i18n` check passes. `license/cla` is pending (the contributor must sign the CLA via the CLA assistant link; this cannot be done from the CLI).
- The Flutter-CI workflow is not triggered while the PR is a draft, and external-contributor workflows may require a maintainer to approve them; `action_required` on the workflow means "awaiting maintainer approval", not a test failure.
- Windows and other platforms have not been tested locally; the change is platform-neutral Flutter code.

## Draft status

Still a draft. Waiting on:

- Issue assignment or maintainer confirmation (claim comment posted on #8922);
- CI workflow approval and execution;
- Maintainer feedback on the chosen interaction rules.

## Summary by Sourcery

Ensure board column footer new-card input cancels cleanly and does not create unintended cards when blurred, escaped, or submitted with empty text.

Bug Fixes:
- Prevent creating cards with empty or whitespace-only titles from the board column footer input.
- Clear any draft text and exit creation mode when the footer input loses focus or Escape is pressed, avoiding stale drafts and accidental submissions.

Tests:
- Extend desktop board integration tests to cover footer behaviors for empty submit, blur cancel, Escape cancel, double Enter, and submit-then-blur without creating duplicates.


tollenceld:
Addressed the Sourcery feedback by replacing the index-based footer lookup with a stable finder and extracting shared board/footer test setup helpers. No product behavior was changed.

